PR #2, opened 2024-03-25
DEVPROD-5136: Configure workspace

sophstad:
### Description
- Configure Yarn Workspace: run `yarn install` at root to install all project dependencies
- Resolve some dependencies that clashed between apps
- Pin `@tanstack/react-table` to version currently used in Spruce
- Fix type errors (not sure why they weren't thrown in the existing apps 🤨)

### Testing
- Evergreen files will be configured in the next PR
- Verified locally that dev processes (`yarn dev`, `yarn prod`) work as expected from within each app directory


.gitignore
```diff
@@ -0,0 +1 @@
+node_modules
```
khelif96, line 1: There are a bunch of other files in our current `.gitignore` should we migrate those over as well?

apps/spruce/src/constants/task.ts
```diff
@@ -221,11 +221,6 @@ export const mapUmbrellaStatusColors: Pick<
     border: purple.dark3,
     text: purple.light3,
   },
-  [TaskStatus.UndispatchedUmbrella]: {
```
khelif96, line 224: This change doesn't seem related, was it intentional?

apps/parsley/src/components/DetailsMenu/DetailsMenuCard/styles.tsx
```diff
@@ -27,7 +27,6 @@ interface DetailsLabelProps {
 const DetailsLabel: React.FC<DetailsLabelProps> = ({ children, label }) => (
   <LabelWrapper>
     <DetailName>{children}</DetailName>
-    {/* @ts-expect-error https://jira.mongodb.org/browse/LG-3797 */}
```
khelif96, line 30: It makes me sad that typescript didn't catch this before :(

I wonder if there was something wrong with how we declared `@ts-expect-error`?

khelif96 (approved):
Great job looking forward to this!

apps/spruce/src/constants/task.ts
sophstad: This key actually appears in the object twice which threw a type error! check line 209

.gitignore
sophstad: The other `.gitignore` files still apply and I don't anticipate many of those files will appear at the root. We can unify at some point but no need right now

PR #3, opened 2024-03-25
DEVPROD-5136: Add testing

sophstad:
### Description
- Merge Evergreen config files. I resolved most inconsistencies between Spruce and Parsley tasks/functions, but the e2e tasks had some differences that I thought would be better addressed later in the project, so they are separate tasks for now
- Use separate build variants for Spruce and Parsley
- Use new AWS role and associated bucket for test artifacts
- Parsley unit tests are running slowly right now due to ts-jest performance. I will address this asap in DEVPROD-5138
- Update snapshots and e2e test failures caused by package changes in #2 
- Use `shallow: true` when running `git.get_project`

### Testing
- Tests are running in the new `evergreen-ui` project!
- ❌ Failing
  - `test` tasks are failing right now because the repo has no git tags. It will pass once we deploy
  - `check_codegen` tasks are failing right now because they're out of date


apps/parsley/scripts/check_codegen.sh
sophstad, line 1: This is the result of running `chmod +x` on this file, removing the need to do it in the Evergreen task

apps/parsley/jest.config.js
```diff
@@ -17,7 +17,6 @@ module.exports = {
   snapshotSerializers: ["@emotion/jest/serializer"],
   testEnvironment: "jsdom",
   testMatch: ["<rootDir>/{src,scripts}/**/*.{spec,test}.{js,jsx,ts,tsx}"],
-  testRunner: "<rootDir>/node_modules/jest-circus/runner.js",
```
sophstad, line 20: This reference to `<rootDir>/node_modules` was no longer valid now that node_modules are installed in the root, but jest-circus [is the default runner](https://jestjs.io/docs/configuration#testrunner-string) so this line is superfluous anyway

.evergreen/attach.yml
```diff
@@ -0,0 +1,170 @@
...
+
+  attach-cypress-results:
+    - command: s3.put
+      type: system
+      params:
+        aws_key: ${AWS_ACCESS_KEY_ID}
+        aws_secret: ${AWS_SECRET_ACCESS_KEY}
+        aws_session_token: ${AWS_SESSION_TOKEN}
+        local_files_include_filter: ["ui/${app_dir}/cypress/screenshots/*"]
+        remote_file: ${build_variant}/${task_id}/
```
khelif96, line 23: I think this needs an execution since task files can be fetched by execution and this would end up overwriting previous executions

.evergreen/attach.yml
```diff
@@ -0,0 +1,170 @@
+functions:
+  attach-codegen-diff:
+    command: s3.put
+    type: system
+    params:
+      aws_key: ${AWS_ACCESS_KEY_ID}
+      aws_secret: ${AWS_SECRET_ACCESS_KEY}
+      aws_session_token: ${AWS_SESSION_TOKEN}
+      local_files_include_filter: ["ui/${app_dir}/bin/codegen.diff"]
+      remote_file: ${build_variant}/${task_id}/codegen/
```
khelif96, line 10: same comment about execution

.evergreen/attach.yml
khelif96, line 1: I think we should include execution as part of the s3 path for files we attach here.

.evergreen/shared.yml
```diff
@@ -0,0 +1,408 @@
...
+  setup-credentials-parsley:
+    command: shell.exec
+    params:
+      working_dir: ui/apps/parsley
+      env:
+        PARSLEY_SENTRY_AUTH_TOKEN: ${PARSLEY_SENTRY_AUTH_TOKEN}
+        REACT_APP_PARSLEY_SENTRY_DSN: ${REACT_APP_PARSLEY_SENTRY_DSN}
+        NEW_RELIC_ACCOUNT_ID: ${NEW_RELIC_ACCOUNT_ID}
+        NEW_RELIC_LICENSE_KEY: ${NEW_RELIC_LICENSE_KEY}
+        NEW_RELIC_TRUST_KEY: ${NEW_RELIC_TRUST_KEY}
```
khelif96, line 97: Are these the same across both apps?

.evergreen/shared.yml
```diff
@@ -0,0 +1,408 @@
...
+
+  yarn-serve:
+    command: shell.exec
+    type: setup
+    params:
+      working_dir: ui/${app_dir}
+      background: true
+      script: |
+        ${PREPARE_SHELL}
+        yarn serve
```
khelif96, line 300: I would love if we could eventually consolidate these I always get confused when jumping between the repos on which one i should run

khelif96 (approved):
Great work Just had a few comments

.evergreen/shared.yml
sophstad: Yeah!

.evergreen/attach.yml
sophstad: This is a good point, I wonder if we have always been overwriting the files lol https://github.com/evergreen-ci/spruce/blob/2f0db0df74f25d0ca665b2e18dbda82670b7566d/.evergreen.yml#L276

.evergreen/shared.yml
sophstad: DEVPROD-5869

PR #4, opened 2024-03-26
DEVPROD-5136: Add dependabot and PR template

sophstad:
DEVPROD-5136

### Description
- Add Dependabot, scheduled to open 5 PRs/week
- Add PR template


.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 5
```
khelif96, line 7: Do you think it would make sense to ensure dependabot opens an even amount of prs per app?

khelif96 (approved):
LGTM

.github/dependabot.yml
sophstad: In a lot of cases Dependabot will be opening 1 PR that updates a dependency in both Spruce and Parsley ([example](https://github.com/evergreen-ci/ui-sandbox/pull/2/files)) so I don't know that this will be necessary. Definitely down to revisit the configuration options if things start to seem uneven though.

PR #5, opened 2024-03-26
DEVPROD-5149: Combine TSConfig

sophstad:
DEVPROD-5149

### Description
- Introduce shared `tsconfig.base.json` at root so that workspaces can extend it
- Upgrade `@types/node` since we were still using a version associated with Node 16


apps/spruce/tsconfig.json
```diff
@@ -1,38 +1,8 @@
...
-    "jsxImportSource": "@emotion/react",
-    "types": [
-      "node",
-      "jest",
-      "@testing-library/jest-dom",
-      "new-relic-browser"
-    ],
     "baseUrl": "src",
-    "noFallthroughCasesInSwitch": true,
+    "strict": false
```
minnakt, line 5: I wonder if we should turn this on! unless it causes a bunch of type errors 🫣 then maybe we should make a ticket

minnakt (approved):
yippeeeeeeee

apps/spruce/tsconfig.json
sophstad: Ya there are mad errors 😳 opened DEVPROD-5887!

tsconfig.base.json
```diff
@@ -0,0 +1,24 @@
...
+    "esModuleInterop": true,
+    "allowSyntheticDefaultImports": true,
+    "forceConsistentCasingInFileNames": true,
+    "module": "CommonJS",
+    "moduleResolution": "Node",
+    "resolveJsonModule": true,
+    "isolatedModules": true,
+    "noEmit": true,
+    "jsx": "react-jsx",
+    "jsxImportSource": "@emotion/react",
```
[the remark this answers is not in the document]
sophstad: JSON does not support comments! ESLint was complaining so I removed them :)

PR #11, opened 2024-03-26
DEVPROD-5136: Only log commits in current directory for deploy

sophstad:
DEVPROD-5136

### Description

Append `-- .` to `git log` commands used to populate deploy script and email. This should result in only the commits with changes in the current directory (i.e. Spruce or Parsley respectively) being logged.

### Testing

Tested locally


minnakt (approved):

khelif96:
Would commits to the shared directories not be logged anymore? for ex if someone changes some shared util?

PR #12, opened 2024-03-27
DEVPROD-5136: Deploy process fixes

sophstad:
DEVPROD-5136

### Description

- Provide correct URL to project commits
- Remove unused `yarn.lock` files from each app
- Update codegen


minnakt (approved):
deploy changes lgtm!

side note: i think the precommit hook still doesn't work (it let me commit with a wrong gql types file to spruce) but maybe i have it configured wrong 😭

sophstad:
> i think the precommit hook still doesn't work
Will be sure to address in DEVPROD-5140!

PR #15, opened 2024-03-27
DEVPROD-4509: Exclude "secret fields" in Sentry Breadcrumbs 

SupaJoon:
DEVPROD-4509

Porting https://github.com/evergreen-ci/spruce/pull/2284


khelif96 (approved):
Assuming this is the same as https://github.com/evergreen-ci/spruce/pull/2284 LGTM! 🚀

PR #16, opened 2024-03-27
DEVPROD-5599: Update Honeycomb Trace link end time

SupaJoon:
DEVPROD-5599

### Description
Add 1 second to the trace link end time since the unix time calculation truncates the result.

### DIFF
--- a/apps/spruce/src/constants/externalResources.test.ts
+++ b/apps/spruce/src/constants/externalResources.test.ts
@@ -19,7 +19,7 @@ describe("getParsleyBuildLogURL", () => {
   });
 });
 
-describe("getTaskTraceUrl", () => {
+describe("getHoneycombTraceUrl", () => {
   it("generates the correct url", () => {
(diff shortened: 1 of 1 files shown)


apps/spruce/src/constants/externalResources.ts
```diff
@@ -79,7 +79,7 @@ export const getHoneycombTraceUrl = (
 ) =>
   `${getHoneycombBaseURL()}/datasets/evergreen-agent/trace?trace_id=${traceId}&trace_start_ts=${getUnixTime(
     new Date(startTs),
-  )}&trace_end_ts=${getUnixTime(new Date(endTs))}`;
+  )}&trace_end_ts=${getUnixTime(new Date(endTs)) + 1}`;
```
sophstad, line 82: May be nice to add a comment here explaining the magic math

sophstad (approved):

PR #23, opened 2024-03-29
DEVPROD-5460: Update GraphQL queries to use specific arguments

minnakt:
DEVPROD-5460

### Description
Uses the new GraphQL arguments.

I'm going to temporarily support both ID and repoID for the `RepoSettings` and `RepoEvents` queries, because I noticed a weird bug where the directive didn't behave as expected (even after making updates in the backend) and I was able to bypass restrictions as a regular user. I'll have to update the directive in the backend anyways, so I'll do that after this PR is merged.

### Testing
- Since I changed the definition of `FormToGql` for project settings, I updated transformers tests


khelif96 (approved):
🚀

minnakt:
tested on beta, should be non-breaking I think

PR #24, opened 2024-03-29
DEVPROD-5138: Merge Jest config

sophstad:
DEVPROD-5138

### Description

- Use Jest's [`projects` option](https://jestjs.io/docs/configuration#projects-arraystring--projectconfig) to allow running multiple projects with one command
- For Parsley, enable `isolatedModules` with ts-jest to fix the performance issues we were seeing (down to 30s from around 2-3+ minutes). This disables type-checking when running the tests, but that's fine since we check typing separately.
- Due to the aforementioned performance problems, as well as a low level of support for the package, I opted not to move Spruce onto ts-jest.
- I suggest eventually moving our tests onto [Vitest](https://vitest.dev), Vite's testing framework. Jest has languished in terms of support, is difficult to configure, and uses a different build system than our actual application. Based on some quick testing Vitest seems a lot easier to configure and run. I added this to our roundtable agenda!

### Testing

- Lots


apps/parsley/jest.config.ts
```diff
@@ -1,25 +1,27 @@
...
-  setupFiles: ["./jest.setup.ts"],
-  preset: "ts-jest",
+  setupFiles: ["./config/jest/jest.setup.ts"],
+  preset: "ts-jest/presets/js-with-ts",
   resetMocks: true,
   setupFilesAfterEnv: ["<rootDir>/config/jest/setupTests.ts"],
   snapshotSerializers: ["@emotion/jest/serializer"],
   testEnvironment: "jsdom",
-  testMatch: ["<rootDir>/{src,scripts}/**/*.{spec,test}.{js,jsx,ts,tsx}"],
+  testMatch: ["<rootDir>/{src,scripts}/**/*.{spec,test}.{ts,tsx}"],
```
minnakt, line 21: can we change this on spruce too or do we have some jsx!

minnakt (approved):
its a YIPPEE from me

PR #25, opened 2024-04-01
DEVPROD-5130: Case on renderingType during log ingestion, introduce LOCAL_UPLOAD logType and fix bug related to exposing renderingType from hook

SupaJoon:
DEVPROD-5130

Porting https://github.com/evergreen-ci/parsley/pull/509

### Description 
These code changes:

- Introduce LogTypes.LOCAL_UPLOAD
- Fix useResolveLogURLAndRenderingType to correctly return the `resmoke` rendering type when applicable
- Update some components to read LogType from context instead of props
- Update the breadcrumb component to show the test name for EVG test logs. 
- Makes some changes to how logs from S3 are fetched

### Testing 
I unit tested the `useResolveLogURLAndRenderingType` hook and then asserted the correct rendering logic is utilized in E2E tests. It's probably useful to ensure ingestLogs is called with the correct rendering type in context which means we would have to unit test LoadingPage. We should actually do this eventually because LoadingPage makes guarantees about the state of context during the log loading phase of the app (DEVPROD-5891).


apps/parsley/scripts/bootstrap-logkeeper.sh
```diff
@@ -14,7 +14,7 @@ if [ ! -d "bin/_bucketdata" ]; then
     # Use aws cli to download the bucket data
     echo "Downloading bucket data..."
     # Try to download the bucket data
-    aws s3 sync --content-encoding gzip  s3://parsley-test/ ./bin
```
khelif96, line 17: Nice Theres a ton of unnecessary data in this bucket.

apps/parsley/scripts/bootstrap-logkeeper.sh
```diff
@@ -26,12 +26,12 @@ if [ ! -d "bin/_bucketdata" ]; then
     fi
     # Uncompress the files in the _bucketdata directory
     echo "Uncompressing bucket data..."
-    tar -xzf ./bin/_bucketdata.tar.gz -C ./bin/
+    tar -xzf bin/_bucketdata.tar.gz -C bin/
```
khelif96, line 29: Shouldn't this still be a relative path?

apps/parsley/src/components/SubHeader/EvergreenTaskSubHeader.tsx
```diff
@@ -27,14 +33,30 @@ export const EvergreenTaskSubHeader: React.FC<Props> = ({
...
 
-  if (loading || !task) {
+  const { data: testData, loading: isLoadingTest } = useQuery<
+    TestLogUrlAndRenderingTypeQuery,
+    TestLogUrlAndRenderingTypeQueryVariables
+  >(GET_TEST_LOG_URL_AND_RENDERING_TYPE, {
+    skip: !(
+      logType === LogTypes.EVERGREEN_TEST_LOGS &&
+      taskID &&
+      Number.isInteger(execution) &&
```
khelif96, line 50: What does this do? execution should already be an integer.

apps/parsley/src/components/SubHeader/EvergreenTaskSubHeader.tsx
```diff
@@ -27,14 +33,30 @@ export const EvergreenTaskSubHeader: React.FC<Props> = ({
...
     taskID,
   });
 
-  if (loading || !task) {
+  const { data: testData, loading: isLoadingTest } = useQuery<
+    TestLogUrlAndRenderingTypeQuery,
+    TestLogUrlAndRenderingTypeQueryVariables
+  >(GET_TEST_LOG_URL_AND_RENDERING_TYPE, {
+    skip: !(
+      logType === LogTypes.EVERGREEN_TEST_LOGS &&
```
khelif96, line 48: I think this might be the only thing we really need to specify everything else should be defined afaict.

khelif96 (changes requested):

khelif96 (approved):
LGTM

PR #26, opened 2024-04-01
DEVPROD-6012 Consolidate graphqlrc and move to root

khelif96:
DEVPROD-6012

### Description
I noticed the graphql integration in vscode stopped working and I think its due to the fact that the `.graphqlrc` file is no longer in the root directory. I moved the files and consolidated them.

### Testing
Intellisense now works 
image


.graphqlrc
```diff
@@ -0,0 +1,2 @@
+schema: "**/sdlschema/**/*.graphql"
+documents: "**/**/src/gql/**/*.graphql"
```
minnakt, line 2: nit. I just prefer the specificity haha
```suggestion
schema: "apps/*/sdlschema/**/*.graphql"
documents: "apps/*/src/gql/**/*.graphql"
```

minnakt (approved):
would you mind just confirming that the intellisense still works if you open a specific app folder in VSCode? for example, I always open `ui/apps/spruce` in VSCode so I never have the parent folder open

khelif96:
> would you mind just confirming that the intellisense still works if you open a specific app folder in VSCode? for example, I always open `ui/apps/spruce` in VSCode so I never have the parent folder open

image
Yup looks like it still works!

PR #27, opened 2024-04-01
DEVPROD-1008 Add new oldest allowed merge base field

hadjri:
DEVPROD-1008

### Description
This adds a new field, "Oldest allowed merge base" that was introduced as part of this ticket, to the PR section of Spruce project settings.

Unsure why `src/gql/generated/types.ts` has seemingly unrelated changes or if that's to be expected but I pulled evergreen main and ran `yarn codegen` locally.
### Screenshots
image

### Testing
Updated a unit test and tested it locally.

### Evergreen PR
https://github.com/evergreen-ci/evergreen/pull/7692


apps/spruce/src/gql/generated/types.ts
sophstad, line 1: > Unsure why src/gql/generated/types.ts has seemingly unrelated changes or if that's to be expected

It is expected! These files are generated from the GraphQL API as defined in Evergreen when codegen is run, so sometimes there are changes other people have made

sophstad (approved):

PR #33, opened 2024-04-01
DEVPROD-5277: Add user preference toggle for jump to failing line

minnakt:
DEVPROD-5277

### Description
The work for this ticket was getting kind of long so I split it up. This PR adds the toggle for the jump to failing line setting (no functionality). It is hidden behind a flag so that it will not show on production.

**Note:** Even though this ticket is part of the epic, we will be releasing this feature early as part of the main Parsley application, which is why it's being merged into the `main` branch.

### Screenshots

Screenshot 2024-04-01 at 4 05 13 PM

### Testing
- Unit tests


apps/parsley/src/components/DetailsMenu/DetailsMenuCard/index.tsx
```diff
@@ -23,6 +34,31 @@ interface DetailsMenuProps {
...
+
+    const updateSettings = (newSettings: ParsleySettingsInput) => {
+      updateParsleySettings({
+        variables: {
+          opts: {
+            parsleySettings: newSettings,
+          },
+        },
+      });
+    };
```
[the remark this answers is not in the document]
minnakt: sure! i think i can make it happen

apps/parsley/package.json
```diff
@@ -32,7 +32,7 @@
     "verify-backend": "./scripts/verify-backend.sh",
     "storybook": "storybook dev -p 6006",
     "storybook:build": "env-cmd -e local -r .env-cmdrc.local.json storybook build",
-    "test": "yarn --cwd='../..' test --selectProjects parsley",
+    "test": "yarn --cwd='../..' test --selectProjects parsley --watchAll=false",
```
minnakt, line 35: Changed this because I was unable to run specific test like `yarn test useParsleySetting`, it would just run all the tests. Adding this allowed me to run specific tests

sophstad: Could you add this to the root `test` script instead? I imagine it'd be helpful for every workspace.

sophstad: And good catch! 😁

PR #34, opened 2024-04-02
DEVPROD-5140: Fix precommit config

sophstad:
DEVPROD-5140

### Description

- Configure pre-commit hook for monorepo according to [husky](https://typicode.github.io/husky/how-to.html#project-not-in-git-root-directory) and [lint-staged](https://github.com/lint-staged/lint-staged?tab=readme-ov-file#how-to-use-lint-staged-in-a-multi-package-monorepo) recommendations
  - Move husky config to root and use lint-staged config files across workspaces to "make sure linters do not "leak" into other packages" as per the lint-staged docs.
- Upgrade Husky to v9

### Testing

- Verified commands run on each app when committing across source directories


khelif96 (approved):
🚀  Works great

PR #55, opened 2024-04-15
DEVPROD-5466: Cypress test isolation for Parsley

SupaJoon:
DEVPROD-5466

### Description
These code changes apply the e2e_test dump and reseed code to Parsley. When a mutation is detected during an e2e_test, EVG data is restored with a dump file that was created during Cypress startup.


minnakt (approved):
lgtm!

PR #59, opened 2024-04-17
DEVPROD-833 Support regex search for files table

khelif96:
DEVPROD-833

### Description
Adds regular expression search support on files table

### Screenshots
image


apps/spruce/src/pages/task/taskTabs/FileTable/index.tsx
```diff
@@ -43,10 +47,11 @@ const FileTable: React.FC<FileTableProps> = ({ execution, taskId }) => {
     <>
       <StyledSearchInput
         aria-label="Search file names"
-        placeholder="Search file names"
-        onChange={(e) => setSearch(e.target.value)}
-        value={search}
+        placeholder="Search file names (regex)"
```
minnakt, line 50: I think you can also make the placeholder `"File name regex"` if you like that option. That will be kind of consistent with the placeholder on the tasks table / My Patches page

minnakt (approved):

PR #76, opened 2024-04-23
DEVPROD-573 Hookify table page and limit functions

khelif96:
DEVPROD-573

### Description
Refactors the `getPageFromSearch` and `getLimitFromSearch` functions we use for tables into one consolidated hook `usePagination`.
Replaces usage of `usePageSizeSelector` with `usePagination`

### Testing
Unit tests and existing e2e tests should cover most of the functionality


apps/spruce/src/components/Pagination/index.tsx
```diff
@@ -28,10 +28,8 @@ const Pagination: React.FC<Props> = ({
   pageSize,
   totalResults,
 }) => {
-  const updateQueryParams = useUpdateURLQueryParams();
-  const handleChange =
-    onChange ||
-    ((page: number) => updateQueryParams({ page: page.toString() }));
+  const { setPage } = useTablePagination();
+  const handleChange = onChange || ((p: number) => setPage(p));
```
sophstad, line 32: [nit] I think you can do this
```suggestion
  const handleChange = onChange || setPage;
```

apps/spruce/src/hooks/useTablePagination/index.ts
```diff
@@ -0,0 +1,38 @@
...
+      setQueryParams({
+        ...queryParams,
+        [PaginationQueryParams.Limit]: pageSize,
+        [PaginationQueryParams.Page]: 0,
+      });
+    },
+    [queryParams, setQueryParams],
+  );
+
+  return { page, limit, setPageLimit, setPage };
```
sophstad, line 35: [nit] Slight preference to call this `setLimit` to be consistent with `page`/`setPage`
```suggestion
  return { page, limit, setLimit, setPage };
```

apps/spruce/src/hooks/useTablePagination/index.ts
```diff
@@ -0,0 +1,38 @@
...
+const useTablePagination = () => {
+  const [queryParams, setQueryParams] = useQueryParams();
+  const [page, setPage] = useQueryParam(PaginationQueryParams.Page, 0);
+  const [limit] = useQueryParam(
+    PaginationQueryParams.Limit,
+    getDefaultPageSize(),
+  );
+
+  /**
+   * `setPageLimit` updates the page size query param and saves the page size to local storage
```
sophstad, line 20: Would note that this function also resets the page to 0

apps/spruce/src/hooks/useTablePagination/useTablePagination.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+    act(() => {
+      result.current.setPageLimit(50);
+    });
+
+    expect(result.current.limit).toBe(50);
+    expect(localStorage.setItem).toHaveBeenCalledWith(
+      RECENT_PAGE_SIZE_KEY,
+      "50",
+    );
```
sophstad, line 46: Could also check that page is reset to 0

apps/spruce/src/constants/queryParams.ts
```diff
@@ -0,0 +1,4 @@
+export enum PaginationQueryParams {
+  Page = "page",
+  Limit = "limit",
+}
```
sophstad, line 4: These values are already defined in `src/components/Table/utils.ts`. Could we reuse those?

apps/spruce/src/components/Pagination/Pagination.stories.tsx
```diff
@@ -0,0 +1,53 @@
...
+export default {
+  component: Pagination,
+} satisfies CustomMeta<typeof Pagination>;
+
+const DefaultImpl = (args: React.ComponentProps<typeof Pagination>) => {
+  const { page } = useTablePagination();
+
+  return (
+    <>
+      <p>Query Params: {page}</p>
```
sophstad, line 15: [nit] I would say this is kind of redundant since it doesn't actually reflect the query params, just the same value that is passed to the Pagination component

sophstad (changes requested):

apps/spruce/src/components/Pagination/Pagination.stories.tsx
khelif96: I updated the verbiage to reflect the usage instead. lmk if that works or if you think I should remove it entirely!

apps/spruce/src/components/Table/utils.ts
```diff
@@ -1,8 +1,6 @@
 import { OnChangeFn } from "@tanstack/react-table";
 
 export enum TableQueryParams {
-  Limit = "limit",
-  Page = "page",
```
sophstad, line 5: Any reason you moved these values outside of this enum instead of using this enum for the hook? I feel like it makes sense to group them under one enum, or at least to colocate the enums.

khelif96: I moved these since pagination is not explicitly restricted to tables. We utilize the same component and logic on the patches pages so it didn't make sense to tie the pagination query params to the table component. I was hoping I could utilize some sort of inheritance with these 2 fields but it is not supported with enums. I could colocate them in the `queryParms` file if that makes sense?

sophstad: Ah good point! And I've also tried that before to the same result 😢 I think the `queryParams` file sounds perfect!

sophstad (approved):

PR #77, opened 2024-04-24
DEVPROD-5137: Combine ESLint & Prettier configs

sophstad:
DEVPROD-5137

### Description

- Add shared ESLint config in new `packages/` directory. This will be the home of our future shared code where we can add directories for things used by Spruce and Parsley.
- Extend custom ESLint config for each application
  - I updated the lint commands to run on all files now, so some previously unlinted files in non-`src/` directories have had fixes introduced.
  - I opted not to move to the [flat config](https://eslint.org/docs/latest/use/configure/configuration-files) yet because it would make sense to upgrade to v9 to leverage all new features, and that's [blocked](https://github.com/jsx-eslint/eslint-plugin-react/issues/3699#issuecomment-2071145657) on our various ESLint dependencies being compatible with v9.
- Update deploy logging to include any commits made to the `packages/` repo when listing commits included in the deploy.

### Testing

- Verified that the updated `git log` commits work locally


apps/parsley/package.json
```diff
@@ -112,28 +110,14 @@
...
     "@types/react": "18.0.15",
     "@types/react-dom": "18.2.25",
-    "@typescript-eslint/eslint-plugin": "6.19.1",
-    "@typescript-eslint/parser": "5.62.0",
     "@vitejs/plugin-react": "3.0.1",
     "babel-plugin-import-graphql": "2.8.1",
     "cypress": "12.7.0",
     "cypress-real-events": "1.11.0",
     "env-cmd": "10.1.0",
     "eslint": "8.31.0",
```
minnakt, line 118: can this eslint be deleted? I see it in `eslint-config-evg` but I'm not sure. Also for prettier, typescript and graphql

apps/spruce/package.json
```diff
@@ -170,21 +163,9 @@
     "babel-preset-react-app": "^10.0.1",
     "camelcase": "^6.1.0",
     "cypress": "12.17.1",
-    "eslint": "8.19.0",
-    "eslint-config-airbnb": "19.0.4",
-    "eslint-config-prettier": "9.1.0",
+    "eslint": "8.31.0",
```
minnakt, line 166: (same question)

apps/spruce/package.json
```diff
@@ -170,21 +163,9 @@
...
+    "eslint-config-evg": "*",
     "eslint-plugin-check-file": "2.6.2",
```
minnakt, line 168: Can we file a ticket to install `eslint-plugin-check-file` in Parsley so we can extract it out later?

apps/parsley/package.json
```diff
@@ -112,28 +110,14 @@
...
-    "eslint-plugin-import": "2.28.1",
-    "eslint-plugin-jest": "27.2.0",
-    "eslint-plugin-jsdoc": "^46.2.4",
-    "eslint-plugin-jsx-a11y": "6.7.1",
-    "eslint-plugin-prettier": "5.1.3",
-    "eslint-plugin-react": "7.33.2",
-    "eslint-plugin-react-hooks": "4.6.0",
-    "eslint-plugin-sort-destructure-keys": "^1.5.0",
+    "eslint-config-evg": "*",
     "eslint-plugin-sort-keys-plus": "^1.3.1",
```
minnakt, line 120: Can file a ticket to install `eslint-plugin-sort-keys-plus` in Spruce so we can extract it out later?

apps/parsley/package.json
```diff
@@ -24,7 +24,7 @@
     "prod": "env-cmd -e devProduction -r .env-cmdrc.local.json vite",
     "eslint:fix": "yarn eslint:strict --fix",
     "eslint:staged": "STRICT=1 eslint",
-    "eslint:strict": "STRICT=1 eslint '*.{js,ts,tsx}' 'src/**/*.ts?(x)' 'cypress/**/*.ts' 'src/gql/**/*.graphql'",
+    "eslint:strict": "STRICT=1 eslint .",
```
minnakt, line 27: When I run `eslint:strict`, I get this 😅:

Screenshot 2024-04-26 at 4 50 13 PM


I have a feeling that it's also scanning my files in the `/dist`, `/bin` folders etc and that we might need to exclude them

packages/eslint-config-evg/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "eslint-plugin-jest": "27.2.1",
+    "eslint-plugin-jsdoc": "48.0.2",
+    "eslint-plugin-jsx-a11y": "6.8.0",
+    "eslint-plugin-react": "7.30.1",
+    "eslint-plugin-react-hooks": "4.6.0",
+    "eslint-plugin-sort-destructure-keys": "^1.5.0",
+    "eslint-plugin-storybook": "0.6.12",
+    "eslint-plugin-testing-library": "^5.10.1",
+    "graphql": "16.8.1",
+    "prettier": "3.2.2",
```
minnakt, line 25: This is kind of odd but prettier is no longer working for me 🥲 I type a bunch of new lines but it won't clean it up for me. Do you observe the same?

apps/spruce/.eslintrc.js
```diff
@@ -120,150 +43,17 @@ module.exports = {
...
-        "@graphql-eslint/no-deprecated": WARN,
-        "@graphql-eslint/alphabetize": [
-          ERROR,
-          { selections: ["OperationDefinition", "FragmentDefinition"] },
-        ],
-        // Allows importing fragments without lint errors.
-        "spaced-comment": OFF,
-        "@typescript-eslint/spaced-comment": OFF,
+    "jest/max-expects": [
+      WARN,
```
minnakt, line 47: I think you can move this one to the main config without any consequence!

apps/spruce/.eslintrc.js
```diff
@@ -5,97 +5,20 @@ const OFF = "off";
...
-        extensions: [".js", ".jsx", ".ts", ".tsx"],
+  root: true,
+  extends: ["evg"],
+  overrides: [
+    {
+      files: ["src/gql/**/*.graphql"],
+      extends: "plugin:@graphql-eslint/operations-recommended",
+      rules: {
+        "@graphql-eslint/selection-set-depth": OFF,
+        "@graphql-eslint/no-deprecated": WARN,
```
minnakt, line 16: We might be able to move this one up as well (but I didn't explicitly check)

apps/parsley/package.json
sophstad: This `eslint` is required since it's used by the eslint Yarn scripts for each app

apps/spruce/package.json
sophstad: DEVPROD-6889

apps/parsley/package.json
sophstad: DEVPROD-6890

apps/parsley/package.json
minnakt: I tried running a [patch](https://spruce.mongodb.com/version/662c1b6c6282860007d799dd/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) after removing eslint in each package.json and I think it worked, but maybe I'm looking at the wrong thing. 😅 But also maybe another acceptable thing would be to put eslint in the parent package.json and use `yarn cwd`

packages/eslint-config-evg/config.js
```diff
@@ -0,0 +1,228 @@
...
+    node: true,
+  },
+  extends: [
+    "eslint:recommended",
+    "plugin:import/recommended",
+    "plugin:jsdoc/recommended-typescript-error",
+    // Airbnb includes some helpful rules for ESLint and React that aren't covered by recommended.
+    // See https://github.com/airbnb/javascript/tree/master/packages for specific rules.
+    "airbnb",
+    "prettier",
```
minnakt, line 21: prettier still does not work for me, and I wonder if it's because it should be this:
```suggestion
    "plugin:prettier/recommended",
```

I think we should fix it because it doesn't show up when I run `eslint:strict` and doesn't fix itself when I run `eslint:fix`. And because it doesn't fix itself for me when I save in VSCode 🥹

This is how it would look on main right now:
Screenshot 2024-04-29 at 12 25 30 PM

apps/parsley/package.json
minnakt: thank you for the explanation!!

packages/eslint-config-evg/config.js
sophstad: Thanks for flagging, I was trying `eslint-config-prettier` [as recommended](https://prettier.io/docs/en/integrating-with-linters.html) but turns out they're actually pretty different. I reintroduced the plugin so things should work again!

packages/eslint-config-evg/config.js
```diff
@@ -0,0 +1,228 @@
+const ERROR = "error";
+const WARN = "warn";
+const OFF = "off";
+
+const errorIfStrict = process.env.STRICT ? ERROR : WARN;
+
```
minnakt, line 6: Yay!! can we also add:

`"prettier/prettier": errorIfStrict`

to make them show up as warnings in local development?

minnakt (approved):
I was just curious, is there any way for us to track which plugins are ready for V9 and which are not? 😲

packages/eslint-config-evg/config.js
minnakt: interesting!! maybe we can take another look in the future 🤔 for now since we're just consolidating stuff this looks great to me!

sophstad:
> I was just curious, is there any way for us to track which plugins are ready for V9 and which are not? 😲

I don't think there's anything super straightforward 😢

PR #91, opened 2024-05-02
DEVPROD-6980: Edit partial download error

sophstad:
DEVPROD-6980

### Description

Remove reference to Evergreen tooling from partial download error

### DIFF
--- a/apps/parsley/src/constants/errors.ts
+++ b/apps/parsley/src/constants/errors.ts
@@ -6,4 +6,4 @@ export const LOG_LINE_TOO_LARGE_WARNING = `Parsley was unable to process one or 
 )}. They have been trimmed to fit the limit.`;
 
 export const LOG_FILE_DOWNLOAD_TOO_LARGE_WARNING =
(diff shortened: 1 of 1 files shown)


apps/parsley/src/constants/errors.ts
```diff
@@ -6,4 +6,4 @@ export const LOG_LINE_TOO_LARGE_WARNING = `Parsley was unable to process one or
 )}. They have been trimmed to fit the limit.`;
 
 export const LOG_FILE_DOWNLOAD_TOO_LARGE_WARNING =
-  "Parsley was only able to partially download this log. Use the Evergreen CLI command in the details menu to download the log onto your machine.";
+  "Parsley was only able to partially download this log. You can use a command line tool to download the file from the raw URL found in the 'Details' menu.";
```
khelif96, line 9: Should we say 
```suggestion
  "Parsley was only able to partially download this log. You can use `wget` or `curl`  to download the file from the raw URL found in the 'Details' menu.";
```
to eliminate any confusion on what command line tool they should use?

khelif96 (approved):
Mod comment

PR #101, opened 2024-05-02
DEVPROD-4268: Add link to GitHub merge queue in patch metadata

minnakt:
DEVPROD-4268

### Description
After speaking to Brian, we can't always link to the base commit for certain patches in the GitHub merge queue. In these cases, we should remove the hyperlink.

Geoff also requested a link to the project's merge queue on GitHub. I showed Geoff the changes and he said they would work for him.

### Screenshots

(Should I add an external link icon through `hideExternalIcon`?)

Screenshot 2024-05-02 at 3 09 44 PM

Screenshot 2024-05-02 at 3 11 19 PM

### Testing
* Test using `yarn prod`

### DIFF
--- a/apps/spruce/src/pages/version/Metadata.tsx
+++ b/apps/spruce/src/pages/version/Metadata.tsx
@@ -7,7 +7,11 @@ import {
   MetadataTitle,
 } from "components/MetadataCard";
 import { StyledLink, StyledRouterLink } from "components/styles";
-import { getGithubCommitUrl } from "constants/externalResources";
+import {
+  getGithubCommitUrl,
+  getGithubMergeQueueUrl,
+} from "constants/externalResources";
+import { githubMergeRequester } from "constants/patch";
 import {
   getCommitQueueRoute,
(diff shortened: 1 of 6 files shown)


apps/spruce/src/pages/version/Metadata.tsx
```diff
@@ -152,6 +155,16 @@ export const Metadata: React.FC<Props> = ({ loading, version }) => {
           </InlineCode>
         </MetadataItem>
       )}
+      {isGithubMergePatch && (
+        <MetadataItem>
+          <StyledLink
```
sophstad, line 160: I think an external icon would be great 👍

sophstad (approved):

apps/spruce/src/pages/version/Metadata.tsx
minnakt: added! for reasons I don't understand, it doesn't show up on development environment, but it'll show up when deploying to staging/beta/etc

PR #106, opened 2024-05-03
DEVPROD-7025: Update event diff table

minnakt:
DEVPROD-7025

### Description
Just one last quick win to remove V11 adapter from the event diff table. I believe this means we no longer have any usage of V11 adapter.

### Screenshots
A nice side effect of this is that things will no longer be cut-off in project & distro events logs:

Before:
Screenshot 2024-05-03 at 2 33 09 PM

After:
Screenshot 2024-05-03 at 2 43 36 PM


minnakt:
evergreen refresh

minnakt:
evergreen retry

sophstad (approved):
🥳

PR #113, opened 2024-05-07
DEVPROD-3188: Attach additional data to ui.click breadcrumbs in Sentry

minnakt:
DEVPROD-3188

### Description
This PR overwrites the breadcrumb message for `ui.click` events if a data-cy attribute is available. Otherwise, the breadcrumb message remains the same.

It also adds additional data to the breadcrumb.

### Caveats
Sometimes this doesn't work that well, and that could be because:
* We don't have a data-cy or any other helpful attributes on the element. We would have to go through and manually add them where missing.
* It is somewhat out of our control. For example, Sentry will pick up on this [hidden div](https://github.com/mongodb/leafygreen-ui/blob/a149857b2bd05dbae003f89ba2a8ba449ff7b213/packages/button/src/ButtonContent/ButtonContent.tsx#L93) that LeafyGreen has on its buttons and not the button itself. I feel like the solution for this would be for LeafyGreen to remove hidden divs.

### Screenshots
Here is what it looks like when the breadcrumb message gets replaced with the tag name and data-cy value. Additional data is also present:

Screenshot 2024-05-07 at 11 36 51 AM

Here is what it looks like when the breadcrumb message could not be replaced. We just add whatever additional data we have:
Screenshot 2024-05-07 at 11 38 28 AM

### Testing
* Manually throw errors in Parsley & Spruce beta

### DIFF
--- a/apps/parsley/src/components/ErrorHandling/Sentry.tsx
+++ b/apps/parsley/src/components/ErrorHandling/Sentry.tsx
@@ -14,6 +14,7 @@ import {
   isProduction,
 } from "utils/environmentVariables";
 import ErrorFallback from "./ErrorFallback";
+import { processHtmlAttributes } from "./utils";
 
 const initializeSentry = () => {
   try {
@@ -21,9 +22,9 @@ const initializeSentry = () => {
       beforeBreadcrumb: (breadcrumb, hint) => {
         if (breadcrumb?.category?.startsWith("ui")) {
           const { target } = hint?.event ?? {};
-          if (target.dataset.cy) {
+          if (target?.dataset?.cy) {
             // eslint-disable-next-line no-param-reassign
-            breadcrumb.message = `${target.tagName.toLowerCase()}[data-cy=${target.dataset.cy}]`;
+            breadcrumb.message = `${target.tagName.toLowerCase()}[data-cy="${target.dataset.cy}"]`;
           }
           // eslint-disable-next-line no-param-reassign
           breadcrumb.data = processHtmlAttributes(target);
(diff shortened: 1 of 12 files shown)


apps/parsley/src/components/ErrorHandling/Sentry.tsx
```diff
@@ -30,6 +42,17 @@ const initializeSentry = () => {
 
 const isInitialized = () => !!getClient();
 
+const processHtmlAttributes = (htmlElement: HTMLElement) => {
+  const { ariaLabel, dataset, title } = htmlElement ?? {};
+  return {
+    ...(ariaLabel && { ariaLabel }),
+    ...(dataset &&
+      Object.keys(dataset).length > 0 && {
+        dataset: htmlElement.dataset,
```
khelif96, line 51: nit.
```suggestion
        dataset
```

apps/parsley/src/components/ErrorHandling/Sentry.tsx
```diff
@@ -18,6 +18,18 @@ import ErrorFallback from "./ErrorFallback";
 const initializeSentry = () => {
   try {
     init({
+      beforeBreadcrumb: (breadcrumb, hint) => {
+        if (breadcrumb?.category?.startsWith("ui")) {
+          const { target } = hint?.event ?? {};
+          if (target.dataset.cy) {
```
khelif96, line 24: Oh this is interesting I didn't know data-* attributes get grouped to dataset!

apps/spruce/src/components/ErrorHandling/Sentry.tsx
```diff
@@ -28,6 +40,18 @@ const initializeSentry = () => {
 
 const isInitialized = () => !!getCurrentHub().getClient();
 
+const processHtmlAttributes = (htmlElement: HTMLElement) => {
+  const { ariaLabel, dataset, title } = htmlElement ?? {};
+  return {
+    ...(ariaLabel && { ariaLabel }),
+    ...(dataset &&
+      Object.keys(dataset).length > 0 && {
+        dataset: htmlElement.dataset,
```
khelif96, line 49: same comment as above!

apps/parsley/src/components/ErrorHandling/Sentry.tsx
```diff
@@ -30,6 +42,17 @@ const initializeSentry = () => {
 
 const isInitialized = () => !!getClient();
 
+const processHtmlAttributes = (htmlElement: HTMLElement) => {
```
khelif96, line 45: nit. can we add a comment and is there any way we can test this and the beforeBreadcrumb hook?

khelif96 (approved):
Great job! 🚀

PR #116, opened 2024-05-10
DEVPROD-5904: Mock clipboard usage in Cypress

minnakt:
DEVPROD-5904

### Description
Attempt to stop the bookmarks test from flaking by using a mock (`cy.spy`) for the clipboard.

### Testing
- Restarted the e2e tests in this PR patch five times without failure


apps/parsley/cypress/support/commands.ts
```diff
@@ -202,13 +206,14 @@ Cypress.Commands.add(
...
-    win.navigator.clipboard.readText().then((text) => {
-      expect(text).to.eq(value);
-    });
-  });
-  // This wait is necessary to ensure the clipboard has time to be read
-  // eslint-disable-next-line cypress/no-unnecessary-waiting
-  cy.wait(50);
+Cypress.Commands.overwrite("visit", (originalVisit, url, options = {}) => {
+  const opts = {
+    onBeforeLoad(win: Window): void {
```
khelif96, line 211: Can we do this in a beforeEach instead or does it need to be in this onBeforeLoad hook?

khelif96 (approved):
🤞 hope this works! 🚀

PR #121, opened 2024-05-14
DEVPROD-3578: Upgrade Vite to v5

sophstad:
DEVPROD-3578

### Description

I'm working on implementing Vitest in DEVPROD-6247 and Vite 5 is a prerequisite


apps/spruce/scripts/deploy/utils/git/git.test.ts
```diff
@@ -1,4 +1,3 @@
-import { execSync } from "child_process";
```
sophstad, line 1: I am not sure why this was only caught as a type error now... but `execSync` is unused in the file

apps/parsley/package.json
```diff
@@ -135,9 +135,9 @@
     "ts-jest": "29.1.2",
     "ts-node": "10.9.1",
     "typescript": "5.1.3",
-    "vite": "4.5.2",
-    "vite-plugin-checker": "0.6.2",
-    "vite-plugin-env-compatible": "1.1.1",
+    "vite": "5.2.11",
+    "vite-plugin-checker": "0.6.4",
+    "vite-plugin-env-compatible": "2.0.1",
     "vite-tsconfig-paths": "4.3.2",
```
minnakt, line 141: Is this something we should try to fix?
Screenshot 2024-05-14 at 3 18 49 PM

Text content:
```
The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.
```

sophstad: I looked into this and it's not just a one-line change, so I opened DEVPROD-7396 to fix (I think the CJS build will be removed entirely in Vite 6).

minnakt (approved):

PR #122, opened 2024-05-16
DEVPROD-6247: Use Vitest for Parsley

sophstad:
DEVPROD-6247

### Description

- Use Vitest instead of Jest for Parsley, so replace every instance of `jest.` with `vi.`
- Update snapshot testing so that each component within a story file gets its own snapshot file. This allows us to use Vitest's [`toMatchFileSnapshot`](https://vitest.dev/api/expect.html#tomatchfilesnapshot) function instead of the `jest-specific-snapshot` package.
- Configure Vitest [workpaces](https://vitest.dev/guide/workspace.html)
- Remove Jest from shared ES Lint config

### Testing

- Renamed Spruce's Evergreen tasks/functions to temporarily include `-jest` suffix so that the Spruce PR can just update the `test` and `snapshot` tasks


apps/parsley/src/components/Badge/TestStatusBadge/__snapshots__/TestStatusBadge_AllBadges.storyshot
```diff
@@ -30,16 +27,4 @@ exports[`Snapshot Tests TestStatusBadge.stories AllBadges 1`] = `
...
-<div>
-  <div
-    class="leafygreen-ui-n4itms"
-    data-cy="test-status-badge"
-  >
-    Pass
-  </div>
-</div>
-`;
+</div>
```
minnakt, line 30: Is it possible to make all of the snapshots end with new lines?

apps/parsley/src/test_utils/index.tsx
```diff
@@ -95,6 +95,41 @@ const renderWithRouterMatch = (
   };
 };
 
+/**
+ * `overwriteFakeTimers` is a workaround for a bug in @testing-libray/react.
+ * It prevents Vitest's fake timers from functioning with user-event.
+ * https://github.com/testing-library/react-testing-library/issues/1197
+ * @returns an anonymous function overwriting jsdon's Jest field
```
minnakt, line 102: nit. (I think these were typos)
```suggestion
 * `overwriteFakeTimers` is a workaround for a bug in @testing-library/react.
 * It prevents Vitest's fake timers from functioning with user-event.
 * https://github.com/testing-library/react-testing-library/issues/1197
 * @returns an anonymous function overwriting jsdom's Jest field
```

apps/parsley/src/test_utils/index.tsx
```diff
@@ -95,6 +95,41 @@ const renderWithRouterMatch = (
...
+
+/**
+ * `stubGetClientRects` fixes a fallbackFocus error introduced by focus-trap.
+ * focus-trap only offers legacy CommonJS exports so it can't be mocked by Vitest.
+ * Instead, spoof focus-trap into thinking there is a node attached.
+ * https://stackoverflow.com/a/75527964
+ */
+const stubGetClientRects = () => {
+  const { getClientRects } = HTMLElement.prototype;
+  HTMLElement.prototype.getClientRects = function () {
```
minnakt, line 125: focus-trap 😾 Could we can disable the eslint rule here? (unless we plan to change it)

apps/parsley/src/utils/highlightHtml/index.tsx
```diff
@@ -46,6 +46,7 @@ export const highlightHtml = (
           );
         }
 
+        // eslint-disable-next-line testing-library/render-result-naming-convention
```
minnakt, line 49: Should the `testing-library` rules apply to non-test files?

apps/parsley/src/utils/errorReporting/errorReporting.test.ts
```diff
@@ -9,13 +10,17 @@ import {
 const { cleanup, mockEnv } = mockEnvironmentVariables();
 
 describe("error reporting", () => {
+  beforeAll(() => {
+    overwriteFakeTimers();
+  });
```
minnakt, line 15: Is there a way to call `overwriteFakeTimers` at the global level so we never have to call this function within individual tests?

apps/parsley/src/components/LogPane/LogPane.test.tsx
```diff
@@ -119,7 +129,9 @@ describe("logPane", () => {
         route: "?shareLine=5",
         wrapper,
       });
-      jest.advanceTimersByTime(100);
+      act(() => {
+        vi.advanceTimersByTime(100);
+      });
```
minnakt, line 134: Is it possible to remove `act`? (The mirror tests above don't have them!)

.evergreen/attach.yml
```diff
@@ -164,6 +164,12 @@ functions:
         preserve_path: true
 
   attach-test-results:
+    command: attach.xunit_results
+    params:
+      files:
+        - "./ui/bin/vite/*.xml"
+
+  attach-test-results-jest:
```
minnakt, line 172: Could we have one command and do:

```
files:
  - "./ui/bin/vite/*.xml"
  - "./ui/bin/jest/*.xml"
```

apps/parsley/src/context/auth/index.tsx
```diff
@@ -94,14 +94,18 @@ const AuthProvider: React.FC<{ children: React.ReactNode }> = ({
...
+      .then(() => {
+        setIsAuthenticated(false);
+        if (isDevelopmentBuild()) {
+          navigate("/login");
+        } else {
+          window.location.href = `${evergreenURL}/login`;
+        }
+      })
+      .catch((err: Error) => {
+        console.error(err);
```
minnakt, line 107: Should we report a warning to Sentry here?

minnakt (commented):
very cool!

apps/parsley/src/components/Badge/TestStatusBadge/__snapshots__/TestStatusBadge_AllBadges.storyshot
sophstad: I think maybe if we implemented a custom snapshot serializer, but that doesn't seem worth it to me 😅

apps/parsley/src/snapshot.test.ts
```diff
@@ -116,9 +127,9 @@ describe(`${options.suite}`, () => {
           const snapshotPath = path.join(
             storyDirectory,
             options.snapshotsDirName,
-            `${storyBookFileBaseName}${options.snapshotExtension}`,
+            `${storyBookFileBaseName}_${name}${options.snapshotExtension}`,
           );
-          expect(container).toMatchSpecificSnapshot(snapshotPath);
+          expect(container).toMatchFileSnapshot(snapshotPath);
```
minnakt, line 132: Based on the [documentation](https://vitest.dev/api/expect.html#tomatchfilesnapshot), it seems like we should await this:

```suggestion
          await expect(container).toMatchFileSnapshot(snapshotPath);
```

minnakt (approved):

PR #123, opened 2024-05-16
DEVPROD-596 PoC: jq in Parsley

FGasper:
DEVPROD-596

### Description
Logs in lines-of-JSON format are commonplace, both at MongoDB and elsewhere. [jq](https://github.com/jqlang/jq) is a popular tool for manipulating such data. This changeset offers a proof-of-concept of integrating jq into Parsley.

### Screenshots
image

### Testing
I manually tested with a small JSON document. I verified a few simple transforms, and I verified that this behaves sensibly when the given filter is invalid.

I verified that Parsley’s existing (regexp-based) filters run _after_ JQ’s filtering.

### Evergreen PR
n/a

#### jq provenance
The JQ version that this includes is built from [my update to jq-web](https://github.com/FGasper/jq-web/tree/issue_33_modern_emscripten). `jq/index.js` has a special `export default jq;` added at the end so that it can be imported as an ES6 module. I would expect a production deployment to use an npm-provided JQ. (jq-web’s maintainer doesn’t seem very interested in the project; I may offer to take it over. A significant part of whipping this up involved just getting it working with modern Emscripten!)

#### Internal notes
This is a simple, proof-of-concept implementation that makes no claim to production-readiness.

Right now Parsley stores the jq-processed logs in memory. It’s probably more ideal to run jq in the renderer instead so that large files don’t cause trouble.

#### UI

It would be good, when/if we put this into production, to have a link handy that either a) goes to jq’s official docs, or b) goes to a “cheat sheet” of handy examples (which itself should probably link to jq’s docs).


khelif96:
Thanks for opening this Felipe this is really cool! We will close this PR for now as we plan on building out more official plugin support in the future which should enable this type of functionality. I attached a link to this pr in the related ticket so that we can reference it in the future. Great work!

PR #124, opened 2024-05-17
DEVPROD-5887: Enable TypeScript strict mode in Spruce

minnakt:
DEVPROD-5887

### Description
In Spruce, we have over 1800 TypeScript errors in strict mode. It would not be feasible to fix all at once. 

However, if we have strict mode disabled, we will continue to unknowingly introduce errors to the codebase.

In this PR, I've enabled strict mode to prevent further additions of any errors. I ran a script to autogenerate `// @ts-expect-error `comments that failed strict validation. I manually fixed a handful of the comments that ended up wonky after generation, but >95% was commented via automation. The errors can be fixed as they are encountered over time.

### How the script works
Generate a report of strict errors by writing the output of `yarn tsc -p tsconfig.json --strict true` to a file. The result looks something like this:
```
src/components/CodeChangesTable/index.tsx(39,7): error TS7031: Binding element 'getValue' implicitly has an 'any' type.
src/components/CodeChangesTable/index.tsx(41,21): error TS7031: Binding element 'diffLink' implicitly has an 'any' type.
src/components/CodeChangesTable/index.tsx(57,14): error TS7031: Binding element 'getValue' implicitly has an 'any' type.
src/components/CodeChangesTable/index.tsx(62,14): error TS7031: Binding element 'getValue' implicitly has an 'any' type.
src/components/CommitChartLabel/CommitChartLabel.test.tsx(7,35): error TS7031: Binding element 'version' implicitly has an 'any' type.
src/components/CommitChartLabel/CommitChartLabel.test.tsx(71,22): error TS2345: Argument of type 'HTMLElement | null' is not assignable to parameter of type 'Element'.
```
Loop through each of these errors and use `sed` to programmatically insert the `// @ts-expect-error` comment at the specified line number for the file.

### Testing
- Spruce should compile without errors


sophstad (commented):
Is there a particular reason not to use `// @ts-expect-error` for these 👁️👁️

minnakt:
> Is there a particular reason not to use `// @ts-expect-error` for these 👁️👁️

No it can be `ts-expect-error`! I changed it via find and replace

apps/spruce/scripts/check-schema-and-codegen/utils.ts
```diff
@@ -1,8 +1,8 @@
-import dns from "dns";
-import fs from "fs";
-import os from "os";
+// @ts-nocheck
```
sophstad, line 1: Why `nocheck`? Could add a comment

sophstad (approved):

PR #130, opened 2024-05-20
DEVPROD-6523: Patch Submission Filter

SupaJoon:
DEVPROD-6523

### Description
These code changes add a Requester filter to the My Patches page.

### Screenshots
Screenshot 2024-05-22 at 12 11 14 PM

### Evergreen PR
https://github.com/evergreen-ci/evergreen/pull/7777

### DIFF
--- a/apps/spruce/src/components/PatchesPage/RequesterSelector.tsx
+++ b/apps/spruce/src/components/PatchesPage/RequesterSelector.tsx
@@ -12,7 +12,7 @@ export const RequesterSelector: React.FC = () => {
     <Combobox
       label=""
       data-cy="requester-selector"
-      placeholder="Select patch types"
+      placeholder="Patch submission"
       value={statusVal}
       multiselect
       onChange={statusValOnChange}


apps/spruce/src/components/PatchesPage/RequesterSelector.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { noFilterMessage } from "constants/strings";
+import { useStatusesFilter } from "hooks";
+import { PatchPageQueryParams } from "types/patch";
+
+export const RequesterSelector: React.FC = () => {
+  const { inputValue: statusVal, setAndSubmitInputValue: statusValOnChange } =
+    useStatusesFilter({ urlParam: PatchPageQueryParams.Requesters });
+
+  return (
+    <Dropdown
```
sophstad, line 12: Would you be opposed to using [Combobox](https://www.mongodb.design/component/combobox/example/) instead of our custom component?

apps/spruce/src/components/PatchesPage/RequesterSelector.tsx
```diff
@@ -0,0 +1,47 @@
...
+        />
+      ))}
+    </Combobox>
+  );
+};
+
+enum Requesters {
+  GithubPRRequester = "github_pull_request",
+  PatchVersionRequester = "patch_request",
+}
```
sophstad, line 34: I think you can instead use `requesterSubscriberOptions` [here](https://github.com/evergreen-ci/ui/blob/38424427727d73834fe5159edc82a4088c248d08/apps/spruce/src/constants/triggers.ts#L56-L62) (imo "Patch" may also be clearer to users since it's terminology used elsewhere)

sophstad (changes requested):

sophstad (approved):
This is really handy, thanks Arjun!!

PR #132, opened 2024-05-21
DEVPROD-1770 Use modern query param hooks in spruce and consolidate sorting methods

khelif96:
DEVPROD-1770

### Description
This PR does a lot of refactoring of the query param methods we use to utilize our new useQueryParam(s) hooks. It also refactors the sorting method we use on some tables to ensure consistent usage of the sort string query params.

This is a large PR so sorry for that! I tried to split up the commits to make it easier to review

### Testing
Existing tests should hopefully cover all of the changes I made. 🤞


apps/spruce/src/hooks/useTableSort/index.ts
```diff
@@ -21,35 +19,24 @@ type CallbackType = (sorter: SortingState) => void;
...
-        .map(({ desc, id }) =>
-          getSortString(id, desc ? SortDirection.Desc : SortDirection.Asc),
-        )
-        .filter(Boolean)
-        .join(";");
+    if (sorter.length) {
+      const sortArray = sorter.map(({ desc, id }) => ({
+        id,
+        direction: desc ? SortDirection.Desc : SortDirection.Asc,
+      }));
```
minnakt, line 35: (Just leaving this one comment since it affects a bunch of files) I'm not sure that we should make this change since it affects so many different tables. For example, sorting on the hosts table doesn't seem to work with these changes applied.

We could change the sorting later in [DEVPROD-4236](https://jira.mongodb.org/browse/DEVPROD-4236) if that's a preferable option

PR #143, opened 2024-05-29
DEVPROD-3936: Enable sampling in Sentry

SupaJoon:
DEVPROD-3936

### Description
These code changes enable sampling in Sentry for Spruce and Parsley. I set the rate to 50% for now. After we get some results we can decide to increase, decrease or keep it.

Docs: https://docs.sentry.io/platforms/javascript/configuration/sampling/

### DIFF
--- a/apps/parsley/src/components/ErrorHandling/initialize.test.ts
+++ b/apps/parsley/src/components/ErrorHandling/initialize.test.ts
@@ -34,6 +34,7 @@ describe("should initialize error handlers according to release stage", () => {
       dsn: "fake-sentry-key",
       environment: "production",
       normalizeDepth: 5,
+      sampleRate: 0.5,
     });
   });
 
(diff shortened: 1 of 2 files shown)


sophstad (commented):
I left a comment in the ticket, but I think there's more work we can try besides sampling to reduce our event count. If you don't want to do it as part of this ticket could you open a new one for that?

apps/parsley/src/components/ErrorHandling/Sentry.tsx
sophstad, line 1: I think the unit tests need fixing

sophstad (changes requested):
> We will probably have to create our own utility if we need more specificity.

That sounds good to me. I think it may also be worth making a general ticket like "Audit Sentry errors" too though, because it seems like there is probably some low hanging fruit that we could stop reporting.

PR #153, opened 2024-06-04
DEVPROD-4200: Add temporary exemption selector

sophstad:
DEVPROD-4200

### Description

- On edit spawn host modal, add a date picker for setting a temporary sleep schedule exemption of up to one month

### Screenshots

image

### Testing

- Add unit and Cypress tests


sophstad:
> if it's possible to disable invalid months from the month dropdown
Good catch, I opened [LG-4322](https://jira.mongodb.org/browse/LG-4322) with LeafyGreen since we don't really have control over that

PR #171, opened 2024-06-11
DEVPROD-7814: Filter Sentry errors from unauthenticated users

SupaJoon:
DEVPROD-7814

### Description
Adjust Sentry config to not submit errors to Sentry unless the session has a valid user id.


khelif96 (changes requested):
Thank you for picking this up!

Conditionally enabling Sentry based on authentication would prevent us from tracking events before the initial authentication request. There might be a more effective approach; The Sentry documentation outlines alternative methods to implement filtering, which could help prevent the error spikes we've been seeing from these (_annoying_) security testing users. You can find more information here: [Sentry Inbound Filters](https://docs.sentry.io/pricing/quotas/#inbound-filters).
I think an effective combination of filter rules would help avoid a lot of the noise since most of these errors follow a predictable pattern. 

Sorry if this was already discussed or if I missed some other context while I was out please feel free to correct me.

khelif96:
> > Thank you for picking this up!
> > Conditionally enabling Sentry based on authentication would prevent us from tracking events before the initial authentication request. There might be a more effective approach; The Sentry documentation outlines alternative methods to implement filtering, which could help prevent the error spikes we've been seeing from these (_annoying_) security testing users. You can find more information here: [Sentry Inbound Filters](https://docs.sentry.io/pricing/quotas/#inbound-filters). I think an effective combination of filter rules would help avoid a lot of the noise since most of these errors follow a predictable pattern.
> > Sorry if this was already discussed or if I missed some other context while I was out please feel free to correct me.
> 
> @khelif96 It looks like Sentry is unable to filter by the sentry `user.id` property but is able to filter by error message. Recently, spam requests are all showing the error: `Cannot read properties of undefined (reading 'target')`. I can add this to the filter but we will miss this issue when usage is legitimate. Sentry is most helpful in surfacing issues that aren't apparent. An error thrown before authentication would likely prevent the user from logging in or using any application features and we would be notified by a human moments after the deploy. The code path and functionality before the authentication step is sparse, not in development and well tested in e2e_test and unit tests. Let me know what you think of this tradeoff.

The sentry docs [recommend](https://docs.sentry.io/platforms/javascript/guides/react/#configure) that it should be initialized before the rest of the application and moving it elsewhere would be against the best practices they recommend. I would prefer if we were to try to find a way to limit this using tools sentry provides. We could reach out to our sentry reps for guidance since we do have that option. We could leverage the #ext-mongodb-sentry slack channel for support from them on what the best approach would be here.

khelif96:
> > > > Thank you for picking this up!
> > > > Conditionally enabling Sentry based on authentication would prevent us from tracking events before the initial authentication request. There might be a more effective approach; The Sentry documentation outlines alternative methods to implement filtering, which could help prevent the error spikes we've been seeing from these (_annoying_) security testing users. You can find more information here: [Sentry Inbound Filters](https://docs.sentry.io/pricing/quotas/#inbound-filters). I think an effective combination of filter rules would help avoid a lot of the noise since most of these errors follow a predictable pattern.
> > > > Sorry if this was already discussed or if I missed some other context while I was out please feel free to correct me.
> > > 
> > > 
> > > @khelif96 It looks like Sentry is unable to filter by the sentry `user.id` property but is able to filter by error message. Recently, spam requests are all showing the error: `Cannot read properties of undefined (reading 'target')`. I can add this to the filter but we will miss this issue when usage is legitimate. Sentry is most helpful in surfacing issues that aren't apparent. An error thrown before authentication would likely prevent the user from logging in or using any application features and we would be notified by a human moments after the deploy. The code path and functionality before the authentication step is sparse, not in development and well tested in e2e_test and unit tests. Let me know what you think of this tradeoff.
> > 
> > 
> > The sentry docs [recommend](https://docs.sentry.io/platforms/javascript/guides/react/#configure) that it should be initialized before the rest of the application and moving it elsewhere would be against the best practices they recommend. I would prefer if we were to try to find a way to limit this using tools sentry provides. We could reach out to our sentry reps for guidance since we do have that option. We could leverage the #ext-mongodb-sentry slack channel for support from them on what the best approach would be here.
> 
> Since you think it's valuable to log unauthenticated requests to some level, the next option to try before reaching out to sentry reps would be Spike Protection. https://docs.sentry.io/pricing/quotas/spike-protection/#how-sentry-detects-spikes

Sorry for the misunderstanding, I don't necessarily think logging unauthenticated requests is valuable but I do think initializing sentry the way the docs recommend for it to be enabled is valuable since that ensures we are using the tool correctly and are maximizing our coverage. I think spike protection makes sense for our case assuming the thresholds work for our level of load. I would recommend leveraging our reps on the best approaches since it's pretty easy to get in touch with them.

PR #177, opened 2024-06-13
DEVPROD-5375: Add staging and beta deploy tasks to Evergreen

sophstad:
DEVPROD-5375 - This work facilitates DEVPROD-5141 by allowing me to test some deploy scripts without pushing anything to prod 😅

### Description

- Add `deploy-parsley-beta`, `deploy-parsley-spruce`, `deploy-spruce-beta`, and `deploy-spruce-staging` tasks to Evergreen in a new `deploy` variant
  - ~Beta tasks are currently broken, blocked on DEVPROD-8149~ ([working](https://spruce.mongodb.com/version/66709737986d5a00071f071a/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)!)
  - Added `setup-credentials` function and a `deploy` function that handles building and pushing to S3. I also considered two separate functions for build and push, let me know if you think that makes more sense 👀 
- Introduce new `deploy-utils` package for deploy code. This runs in a node environment so I thought it made sense to separate from @evg-ui/lib
- Move Spruce's output directory from `build/` to `dist/` so that code can be shared with Parsley

### Testing

- Add unit tests


packages/deploy-utils/tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+    "baseUrl": "src",
+    "module": "CommonJS",
+    "moduleResolution": "Node",
+    "paths": {
+      "@evg-ui/lib/*": ["../../lib/src/*"]
+    },
+    "types": [
+      "vitest/globals",
+      "vitest/jsdom"
+    ]
```
minnakt, line 13: I was wondering if these overrides not get inherited from the base tsconfig? :0 (excluding the `paths` field)

.evergreen/deploy.yml
```diff
@@ -0,0 +1,96 @@
+functions:
+  setup-credentials:
+    command: shell.exec
```
minnakt, line 3: Should we consider rewriting this with `subprocess.exec`?

packages/deploy-utils/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@evg-ui/deploy-utils",
```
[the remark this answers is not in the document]
sophstad: Kinda wild but I believe a ts-node bug prevents this: https://stackoverflow.com/a/63459149

.evergreen/deploy.yml
sophstad: Imo this isn't super straightforward so would prefer to leave it for the ticket that captures that work!

minnakt (approved):

PR #185, opened 2024-06-14
DEVPROD-7831: Prevent 422 when clicking on version/task action button dropdown

SupaJoon:
DEVPROD-7831

### Description
The SetPriorityButton called into task and version analytics hooks and made a request even when the required information was unavailable from then URL. The button is rendered after clicking the ellipsis button on the task and version page.


apps/spruce/src/analytics/task/useTaskAnalytics.ts
```diff
@@ -74,6 +74,7 @@ export const useTaskAnalytics = () => {
 
   const [execution] = useQueryParam(RequiredQueryParams.Execution, 0);
   const { data: eventData } = useQuery<TaskQuery, TaskQueryVariables>(TASK, {
+    skip: !taskId || execution === undefined,
```
sophstad, line 77: Does `execution === undefined` ever evaluate to true? It seems like the variable has a default of 0 from `useQueryParam`

sophstad (approved):

PR #187, opened 2024-06-17
DEVPROD-7396: Convert Spruce & Parsley to ESM

minnakt:
DEVPROD-7396

### Description
* Enable ESM in Spruce & Parsley
* Bump node to 20.11.0 to get access to `import.meta.dirname`

### Testing
You should no longer see the warning message
```
The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.
```
when starting each app.

### DIFF
--- a/apps/parsley/package.json
+++ b/apps/parsley/package.json
@@ -1,7 +1,7 @@
 {
   "name": "parsley",
+  "version": "2.1.25",
   "private": true,
-  "version": "2.1.24",
   "type": "module",
   "scripts": {
     "bootstrap-logkeeper": "./scripts/bootstrap-logkeeper.sh",
--- a/apps/spruce/package.json
+++ b/apps/spruce/package.json
@@ -1,6 +1,6 @@
 {
   "name": "spruce",
-  "version": "4.1.37",
+  "version": "4.1.38",
   "private": true,
   "type": "module",
   "scripts": {


apps/parsley/tsconfig.json
```diff
@@ -4,7 +4,8 @@
     "baseUrl": "src",
     "paths": {
       "@evg-ui/lib/*": ["../../../packages/lib/src/*"]
-    }
+    },
+    "module": "ESNext"
```
sophstad, line 8: Should we just be replacing the `module` field in the root `tsconfig.base.json`? I don't know that any package needs CommonJS (or if they do, they could override the base default)

apps/parsley/lint-staged.config.cjs
sophstad, line 1: Can we convert these files to use ES6 modules now?

apps/parsley/tsconfig.json
minnakt: this opened a whole new can of worms so i'm sorry about the additional file changes! Also I just have to call out that `ts-node` legit just doesn't work with Node 20 and ESM modules period ([GitHub issue here](https://www.github.com/TypeStrong/ts-node/issues/1997)) so if you see funky workarounds... that's why....

packages/deploy-utils/src/utils/s3.ts
```diff
@@ -9,6 +9,7 @@ export const pushToS3 = (bucket: string) => {
     );
     console.log("Successfully deployed to S3");
   } catch (e) {
-    throw Error("Deployment to S3 failed", { cause: e });
+    console.error(e);
+    throw Error("Deployment to S3 failed");
```
sophstad, line 13: Just curious, what's this change for?

minnakt: I have no idea. Must have been an artifact from when I was fighting for my life
Deleted!

sophstad (approved):
🥳

PR #204, opened 2024-06-25
DEVPROD-8143: Prevent searchbar error from pushing content 

SupaJoon:
DEVPROD-8143

### Description
The most recent upgrade to TextInput resulted in an error message underneath the text-input which pushed content, removed the error icon and adjusted which elements in the dom controlled styles. These code changes remove the error message copy, add in the error icon and adjust the text-input border when it's next a select component.

### DIFF
--- a/apps/parsley/src/components/Search/SearchBar/index.tsx
+++ b/apps/parsley/src/components/Search/SearchBar/index.tsx
@@ -179,7 +179,7 @@ const SearchBar: React.FC<SearchBarProps> = ({
               <Tooltip
                 justify="middle"
                 trigger={
-                  <div>
+                  <div data-cy="searchbar-error">
                     <RedIcon glyph="Warning" />
                   </div>
                 }
--- a/apps/parsley/src/components/Search/SearchBar/index.tsx
+++ b/apps/parsley/src/components/Search/SearchBar/index.tsx
@@ -16,6 +16,7 @@ import { useKeyboardShortcut } from "hooks";
 import { SentryBreadcrumb, leaveBreadcrumb } from "utils/errorReporting";
 import SearchPopover from "./SearchPopover";
 
+const { red } = palette;
 interface SearchBarProps {
   className?: string;
   disabled?: boolean;
@@ -180,7 +181,7 @@ const SearchBar: React.FC<SearchBarProps> = ({
                 justify="middle"
                 trigger={
                   <div data-cy="searchbar-error">
-                    <Icon fill={palette.red.base} glyph="Warning" />
+                    <Icon fill={red.base} glyph="Warning" />
                   </div>
                 }
                 triggerEvent="hover"
--- a/apps/parsley/src/components/Search/SearchBar/index.tsx
+++ b/apps/parsley/src/components/Search/SearchBar/index.tsx
@@ -226,6 +226,10 @@ const StyledInput = styled(TextInputWithGlyph)`
     border-left: 0;
     padding-left: 42px;
   }
+  div[data-lgid="lg-form_field-feedback"] {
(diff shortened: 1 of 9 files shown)


apps/parsley/src/components/Search/SearchBar/index.tsx
```diff
@@ -226,6 +226,10 @@ const StyledInput = styled(TextInputWithGlyph)`
     border-left: 0;
     padding-left: 42px;
   }
```
khelif96, line 228: Could we also fix the border rounding on this component since it was also broken when this broke? 
This component should look like this
image

khelif96 (changes requested):
Taking a look at the state of the component before this bug was introduced it looks like this isn't the only way this component broke. The error state was represented completely differently prior to the breaking [change](https://github.com/evergreen-ci/ui/pull/73). See the [old storybook](https://evg-bucket-evergreen-ui.s3.amazonaws.com/parsley/evergreen_ui_parsley_storybook_63d3ae9d56d28a5c7d0d6166916b12fd5b1fce43_24_05_29_16_32_51/0/storybook/ui/apps/parsley/storybook-static/index.html?path=/docs/components-search-searchbar--docs) from the commit prior to the breaking commit being introduced.

This bug also exists in spruce within the `TupleSelect` component.

apps/parsley/src/components/Search/SearchBar/index.tsx
```diff
@@ -220,12 +225,20 @@ const InputWrapper = styled.div`
...
     border-bottom-left-radius: 0;
-    border-left: 0;
+    border-left: none;
+  }
+  input[data-cy="searchbar-input"] {
     padding-left: 42px;
   }
+  div[data-lgid="lg-form_field-feedback"] {
+    display: none;
+  }
```
khelif96, line 238: I'm not a huge fan of depending on internal attributes of 3rd party components like this since we can't control what they will be which is partly why this broke in the first place. Have you considered the alternative approach of not using this component's "error" state? We are overriding the error state anyways so there doesn't seem to be much value in using it.

apps/parsley/src/components/Search/SearchBar/index.tsx
```diff
@@ -177,7 +178,11 @@ const SearchBar: React.FC<SearchBarProps> = ({
             ) : (
               <Tooltip
                 justify="middle"
-                trigger={<IconPlaceholder data-cy="searchbar-error" />}
+                trigger={
+                  <div data-cy="searchbar-error">
+                    <RedIcon glyph="Warning" />
```
khelif96, line 183:
```suggestion
                    <Icon fill={red.base} glyph="Warning" />
```

apps/parsley/src/components/Search/SearchBar/index.tsx
```diff
@@ -220,12 +225,20 @@ const InputWrapper = styled.div`
...
+  input[data-cy="searchbar-input"] {
     padding-left: 42px;
   }
+  div[data-lgid="lg-form_field-feedback"] {
+    display: none;
+  }
+`;
+const RedIcon = styled(Icon)`
+  color: ${palette.red.base};
 `;
```
khelif96, line 242:
```suggestion
```

apps/parsley/src/components/Search/SearchBar/index.tsx
SupaJoon: I opted to utilize the error state because it comes with the border-color. When it comes to hacking on packages, I rather take things away than add things in to follow the the original pattern as closely as possible.

khelif96 (approved):

PR #205, opened 2024-06-25
DEVPROD-7529 Include downstreamRevision in save project settings mutation

hadjri:
DEVPROD-7529

### Description
Add new `downstreamRevision` field to the UI so that when saving patch trigger aliases in project settings, the field doesn't get cleared. Not implementing a UI component for this field because it's going to be API-only to start, per the linked Evergreen PR.

### Testing
Tested locally

### Evergreen PR
https://github.com/evergreen-ci/evergreen/pull/8012


khelif96 (approved):
Makes sense to me 👍

PR #210, opened 2024-06-27
DEVPROD-5141: Merge deploy scripts

sophstad:
DEVPROD-5141

### Description

Sorry this is kinda chunky 😓 
- Add `src/deploy.sh` to `@evg-ui/deploy-utils` to route production vs staging/beta deploys
- Introduce deploy scripts using [package.json's `bin` field](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#bin). Each `script.ts` file represents one of these scripts, which allows for easily running commands in our Evergreen YAML and locally.
  - Run scripts with [vite-node](https://github.com/vitest-dev/vitest/tree/main/packages/vite-node#readme). I found this to be much more straightforward than ts-node since it inherits from Vite config files. This means things like tsconfig paths don't have to be configured twice.
  - Require `--force` flag to run local deploy scripts
- Keep existing deploy scripts around; in case the new one fails, the deployer can revert the git tag project settings and run `yarn deploy:prod:old` (will delete in DEVPROD-7027)
- I'll be fixing the email script and converting to TS for testability in DEVPROD-7027.

### Testing

- Add unit tests
- Patch testing [write-previous-deploy](https://spruce.mongodb.com/task/evergreen_ui_spruce_test_write_previous_deploy_patch_afd1d715aa6f89e8f3a8f5efecb1678c8bc3ca18_667d94375e8be70007848d83_24_06_27_16_33_10/logs?execution=0)


packages/deploy-utils/src/setup-credentials/script.ts
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env -S vite-node --script
+
+import { setupCredentials } from ".";
+import { isTarget } from "../utils/types";
+
+const target = process.argv[2];
+if (!isTarget(target)) {
```
SupaJoon, line 7: maybe allow the `isTarget` function parameter to be `string | undefined` in the function definition. I see a few places where we apply `?? ""` to the parameter

apps/spruce/lint-staged.config.js
```diff
@@ -2,5 +2,5 @@ import baseConfig from "@evg-ui/lint-staged";
 
 export default {
   ...baseConfig,
-  "*.{ts,tsx}": () => ["yarn check-types", "yarn check-schema-and-codegen"], // For TypeScript files, run tsc, and gql schema check
```
SupaJoon, line 5: why drop check-types?

packages/deploy-utils/src/utils/utils.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    expect(getAppToDeploy()).toEqual("spruce");
+  });
+
+  it("throws an error when run from directory inside app", () => {
+    vi.spyOn(process, "cwd").mockReturnValue(
+      "/Users/username/evergreen-ci/ui/apps/spruce/hello",
+    );
+
+    expect(() => getAppToDeploy()).toThrowError(
```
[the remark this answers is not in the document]
sophstad: This is actually the syntax for the [toThrowError matcher](https://vitest.dev/api/expect.html#tothrowerror). Removing `() =>` causes the test to fail.

apps/spruce/lint-staged.config.js
sophstad: It's already applied to all TS/TSX files in the base config!

packages/deploy-utils/src/utils/git/tag.test.ts
```diff
@@ -0,0 +1,14 @@
+import { getLatestTag } from ".";
+import { DeployableApp } from "../types";
+
+describe("getLatestTag", () => {
+  const currentlyDeployedTagRegex = (app: DeployableApp) =>
+    new RegExp(`${app}/v\\d+.\\d+.\\d+`);
```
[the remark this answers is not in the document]
sophstad: It's not possible to assert an exact string value since we aren't mocking the value returned by `execSync` in `getLatestTag`. IMO that would defeat the usefulness of this test since we want to assert that the git tags are filtered correctly. I did add a test to verify that the regex matcher works against a known-valid string!

SupaJoon (approved):

PR #211, opened 2024-06-28
DEVPROD-8159: Prevent Version Metadata Panel from accessing undefined

SupaJoon:
DEVPROD-8159

### Description
In-app navigation to the version page from a version page results in the Version Metadata panel processing an undefined version object. Resetting the version page state when the versionID URL field changes properly gates page rendering until new data is fetched. 
I'm unable to edit the diagram but this case is basically captured by the diagram entry statement "User Visits /version/{id}".

You can recreate this issue by clicking Previous Commit on this [page](https://spruce.mongodb.com/version/mongodb_mongo_v5.0_809167cbc5a70e1418d8090a7bb6a44395163c7a/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC).

### DIFF
--- a/apps/spruce/src/pages/Version.tsx
+++ b/apps/spruce/src/pages/Version.tsx
@@ -44,12 +44,6 @@ export const VersionPage: React.FC = () => {
   const [redirectURL, setRedirectURL] = useState(undefined);
   const [isLoadingData, setIsLoadingData] = useState(true);
 
-  // Reset state when version ID changes.
-  useEffect(() => {
-    setIsLoadingData(true);
-    setRedirectURL(undefined);
-  }, [versionId]);
-
   // This query is used to fetch the version data.
   const [getVersion, { data: versionData, error: versionError }] = useLazyQuery<
     VersionQuery,
--- a/apps/spruce/src/pages/Version.tsx
+++ b/apps/spruce/src/pages/Version.tsx
@@ -45,7 +45,6 @@ export const VersionPage: React.FC = () => {
   const [isLoadingData, setIsLoadingData] = useState(true);
 
   // Reset state when version ID changes.
-  // This triggers a re-fetch of the HasVersion query.
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


apps/spruce/src/pages/Version.tsx
```diff
@@ -44,6 +44,12 @@ export const VersionPage: React.FC = () => {
   const [redirectURL, setRedirectURL] = useState(undefined);
   const [isLoadingData, setIsLoadingData] = useState(true);
 
+  // Reset state when version ID changes.
+  useEffect(() => {
+    setIsLoadingData(true);
+    setRedirectURL(undefined);
+  }, [versionId]);
+
```
sophstad, line 52: I think adding keyArgs for the HasVersion query [here](https://github.com/evergreen-ci/ui/blob/92881dfbd8407451d20c7aaa589ac7f31d14323b/apps/spruce/src/gql/client/cache.ts#L6) also fixes the issue. IMO that's a little more elegant since it better leverages the Apollo cache!

sophstad (changes requested):

apps/spruce/src/pages/Version.tsx
SupaJoon: I found that the cache setting fixes the error but it's not completely clear to me how because the setting change represents the default behavior. Apollo Client docs mention [here](https://www.apollographql.com/docs/react/caching/cache-field-behavior/#example-1) that all of a field's arguments are keyArguments by default. I also tested hasVersion query with a network-only fetch-policy and still got the error.

sophstad: > Apollo Client docs mention [here](https://www.apollographql.com/docs/react/caching/cache-field-behavior/#example-1) that all of a field's arguments are keyArguments

This is a great point and has me confused! Glad it works though...

sophstad (approved):

PR #212, opened 2024-06-28
DEVPROD-4046: Trim lines split between chunks

sophstad:
DEVPROD-4046

### Description

The current approach to trimming lines failed to catch any long lines split over multiple chunks. When concatenating the first line with the previously processed line, check whether their combined length is too long and trim if so.

### Testing

- Add unit test that fails without code changes
- Log linked in ticket now shows warning toast and trims correctly


khelif96 (approved):
Works great!

PR #214, opened 2024-06-28
DEVPROD-7957 Spruce goes to the beach!

khelif96:
DEVPROD-7957

### Description
Removed some of the old logos and added the summer one

### Screenshots
image


sophstad (commented):
Any reason not to keep the old logos for next year?

khelif96:
> Any reason not to keep the old logos for next year?

I just cleaned up since they weren't used anymore. They are in the git history if we want to add them back in or I can revert as part of this pr, it's up to you!

sophstad (approved):
git history works for me! 🏖️

PR #215, opened 2024-07-01
DEVPROD-8220: Render Command Sections

SupaJoon:
DEVPROD-8220

### Description
These code changes render command sections by iterating over `SectionData.commands` in `filterLogs`. As of this PR, sections are only rendered when no filters are applied. That behavior will be update in [DEVPROD-8015](https://jira.mongodb.org/browse/DEVPROD-8015).


apps/parsley/src/components/LogRow/RowRenderer/index.tsx
```diff
@@ -87,7 +88,16 @@ const ParsleyRow: RowRendererFunction = ({ processedLogLines }) => {
     }
 
     if (isSubsectionHeaderRow(processedLogLine)) {
-      return <div>subsection start</div>;
+      return (
+        <SubsectionHeader
```
sophstad, line 92: What do you think of adding the command number to the header itself? E.g. something like "(1 of 3)" for this command:
image

apps/parsley/src/components/LogRow/RowRenderer/index.tsx
```diff
@@ -87,7 +88,16 @@ const ParsleyRow: RowRendererFunction = ({ processedLogLines }) => {
...
+          commandID={processedLogLine.commandID}
+          commandName={processedLogLine.commandName}
+          functionID={processedLogLine.functionID}
+          lineIndex={index}
+          onToggle={toggleCommandSection}
+          open={processedLogLine.isOpen}
```
sophstad, line 98: Did you consider auto-opening commands if there's only one present? If we don't include the command number as suggested above, this could be confusing because it would be hard to know if it's the only command. But if the count is included, it would be nice to avoid the extra step of opening this. Interested in your thoughts!
image

apps/parsley/src/components/LogRow/RowRenderer/index.tsx
```diff
@@ -87,7 +88,16 @@ const ParsleyRow: RowRendererFunction = ({ processedLogLines }) => {
     }
 
     if (isSubsectionHeaderRow(processedLogLine)) {
```
sophstad, line 90: It may be cool to add some type of tiny drop shadow to the top of the function bar for visual separation
image

apps/parsley/src/utils/filterLogs/index.ts
```diff
@@ -49,26 +73,69 @@ const filterLogs = (options: FilterLogsParams): ProcessedLogLines => {
...
+        } else if (isCommandStart) {
+          // A command start is detected.
+          const commandLine = getProccessedCommandLine(command, sectionState);
+          filteredLines.push(commandLine);
+          if (commandLine.isOpen) {
+            filteredLines.push(idx);
+          } else {
+            idx = command.range.end - 1;
           }
+          commandIndex += 1;
```
sophstad, line 136: This is the exact same logic implemented above in lines 97-110. Given that, is it even necessary to check `isFuncOpen` initially?

sophstad (commented):
Works great, I mostly had some UX thoughts!

apps/parsley/src/components/LogRow/RowRenderer/index.tsx
SupaJoon: It's a good idea. I just opened this ticket: https://jira.mongodb.org/browse/DEVPROD-8683

apps/parsley/src/components/LogRow/RowRenderer/index.tsx
SupaJoon: I opened this ticket to verify some default behavior. I will bring these up in design sync this week. https://jira.mongodb.org/browse/DEVPROD-8684

apps/parsley/src/components/LogRow/RowRenderer/index.tsx
SupaJoon: I think the visually distinct properties like the caret, icon and button and spacing create enough visual separation to not require a box-shadow that runs across the whole row. After this project is implemented, I will show Will the components to see if there are any tweaks we can make for better readability.

apps/parsley/src/utils/filterLogs/index.ts
SupaJoon: Good point, I flattened the if-statements to make the algorithm easier to read. `isFuncOpen` determines if the following commands should be skipped or processed so it should be checked initially.

sophstad (approved):

PR #221, opened 2024-07-01
DEVPROD-7917 Fix bug where adding a filter resets url params if HighlightFilters is enabled.

khelif96:
DEVPROD-7917

### Description
This bug was caused by the Search component incorrectly overridign the existing query params with the wrong value when ever a filter would be added when the `HighlightFilters` feature was used.

### Testing
Added some unit tests to this component to cover this use case

### DIFF
--- a/apps/parsley/src/components/Search/Search.test.tsx
+++ b/apps/parsley/src/components/Search/Search.test.tsx
@@ -60,7 +60,7 @@ describe("Search", () => {
   });
   it("adding a filter should update the url", async () => {
     const user = userEvent.setup();
-    const { hook, utils } = renderSearch();
+    const { utils } = renderSearch();
     expect(utils.router.state.location.search).toBe("");
     await user.type(screen.getByDataCy("searchbar-input"), "test");
     await user.type(


apps/parsley/src/components/Search/Search.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+    const user = userEvent.setup();
+    const { hook, utils } = renderSearch();
+    expect(utils.router.state.location.search).toBe("");
+    await user.type(screen.getByDataCy("searchbar-input"), "test");
+    await user.type(
+      screen.getByDataCy("searchbar-input"),
+      "{Meta>}{enter}",
+      {},
+    );
+    expect(hook.current.useQueryParams[0][QueryParams.Filters]).toBe("100test");
```
SupaJoon, line 85: If the main purpose of `useQueryParams` is to test the URL value, I think we should access that directly instead of involving an additional package in between. It may be accessible via `global.window.location`

SupaJoon (approved):

PR #228, opened 2024-07-03
DEVPROD-5141: Fix commit check for prod deploy

sophstad:
DEVPROD-5141 - Fix bug introduced in #210 blocking new deploy script from working

### Description

- Fix string check: commit fetched from remote file was not being trimmed and so was not 40 characters
- Also validate tag correctly if local commit is used, since it would not be 40 characters

### Testing

- Add unit tests

### DIFF
--- a/packages/deploy-utils/src/utils/git/git.test.ts
+++ b/packages/deploy-utils/src/utils/git/git.test.ts
@@ -71,10 +71,8 @@ describe("getCurrentlyDeployedCommit", () => {
 
   it("returns a valid local commit", async () => {
     vi.mocked(get).mockImplementation(errorMock);
-    vi.mocked(execSync).mockReturnValue(validCommitString);
-    expect(await getCurrentlyDeployedCommit("spruce")).toEqual(
-      validCommitString,
-    );
+    vi.mocked(execSync).mockReturnValue("spruce/v1.0.0");
+    expect(await getCurrentlyDeployedCommit("spruce")).toEqual("spruce/v1.0.0");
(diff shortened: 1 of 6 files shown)


packages/deploy-utils/src/utils/git/git.test.ts
```diff
@@ -71,10 +71,8 @@ describe("getCurrentlyDeployedCommit", () => {
 
   it("returns a valid local commit", async () => {
```
minnakt, line 72: It might be interesting to see if `getCurrentlyDeployedCommit` could be tested by executing it for real, I think the `getCurrentlyDeployedCommit` from `apps/spruce/scripts/deploy/utils/git` is tested without any mocking :O just an observation though I'm not sure if that's difficult

packages/deploy-utils/src/utils/git/tag.test.ts
```diff
@@ -1,31 +1,25 @@
...
-    expect(tag).toBeTruthy();
+describe("tagIsValid", () => {
+  it("should match on a known valid tag", () => {
+    expect(tagIsValid("parsley", "parsley/v1.2.3")).toEqual(true);
   });
 
-  it("currentlyDeployedTagRegex should not match on the wrong app's tag tag", () => {
-    const tag = currentlyDeployedTagRegex("parsley").test("spruce/v1.2.3");
-    expect(tag).toBeFalsy();
+  it("should not match on the wrong app's tag tag", () => {
```
minnakt, line 8:
```suggestion
  it("should not match on the wrong app's tag", () => {
```

minnakt (approved):

PR #231, opened 2024-07-08
DEVPROD-5295: Display function status on Section Header

SupaJoon:
DEVPROD-5295


apps/parsley/src/components/LogRow/RowRenderer/index.tsx
```diff
@@ -74,14 +75,17 @@ const ParsleyRow: RowRendererFunction = ({ processedLogLines }) => {
...
+        : SectionStatus.Pass;
       return (
         <SectionHeader
           functionID={processedLogLine.functionID}
           functionName={processedLogLine.functionName}
           lineIndex={index}
           onToggle={toggleFunctionSection}
           open={processedLogLine.isOpen}
-          status={SectionStatus.Pass} // TODO: Update in DEVPROD-5295
+          status={status}
```
minnakt, line 88: one thing I was wondering was if it makes sense to also store the `status` in `processedLogLine` like some of these other properties, I was just curious about the choice to evaluate it via a function here

apps/parsley/src/components/LogRow/RowRenderer/index.tsx
```diff
@@ -74,14 +75,17 @@ const ParsleyRow: RowRendererFunction = ({ processedLogLines }) => {
     }
 
     if (isSectionHeaderRow(processedLogLine)) {
+      const status = includesLineNumber(processedLogLine, failingLine)
```
minnakt, line 78: maybe this could be inlined below since `status` isn't reused anywhere

minnakt (approved):
yippee!

apps/parsley/src/components/LogRow/RowRenderer/index.tsx
SupaJoon: I chose to evaluate it during render to keep ProcessedLogLine simpler. If the value was used in a util function that uses ProcessedLogLines, then I would have added it in there. I could have done a similar pattern for the step information also but since it requires regex matching, I thought it's best to do that in the parsing step so parsing is all in one place.

PR #238, opened 2024-07-10
DEVPROD-8537: Require minimum length for image field

minnakt:
DEVPROD-8537

### Description
Require minimum length for the distro's `image_id` field. I have checked the distros on staging & production and filled in this field where it was missing.


minnakt:
evergreen retry

khelif96 (approved):

PR #241, opened 2024-07-10
DEVPROD-7910: Fix unresponsive notifications page bug

sophstad:
DEVPROD-7910

### Description

On the notifications page, making any change while not having any individual subscriptions would cause the page to become unresponsive. This was a result of the internal Tanstack table within the subscriptions table attempting to perform operations while empty. The easiest way to get around this is just to split out the table into a separate component that is not rendered if the user has no subscriptions. 

### Testing

Editing the page with no subscriptions is successful


minnakt (approved):
waaa good one 🙌

SupaJoon (approved):

PR #254, opened 2024-07-16
DEVPROD-6861: Show error toast when fetching Task.taskLogs

SupaJoon:
DEVPROD-6861

https://github.com/evergreen-ci/evergreen/pull/8105 introduces a permissions check for the `Task.taskLogs` GQL field and this PR surfaces those errors in the UI when they occur.

### DIFF
--- a/apps/spruce/src/pages/task/taskTabs/logs/LogTypes.tsx
+++ b/apps/spruce/src/pages/task/taskTabs/logs/LogTypes.tsx
@@ -7,6 +7,7 @@ import { useParams, useLocation } from "react-router-dom";
 import { DEFAULT_POLL_INTERVAL } from "constants/index";
 import { slugs } from "constants/routes";
 import { size, fontSize } from "constants/tokens";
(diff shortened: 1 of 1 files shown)


apps/spruce/src/pages/task/taskTabs/logs/LogTypes.tsx
```diff
@@ -53,14 +54,19 @@ export const AllLog: React.FC<Props> = (props) => {
   const location = useLocation();
   const parsed = parseQueryString(location.search);
   const selectedExecution = Number(parsed[RequiredQueryParams.Execution]);
```
khelif96, line 56: nit. Not a requirement but it would be ⭐ _amazing_ ⭐ if we could replace these with `useQueryParam` while we are here.
```suggestion
  const [selectedExecution] = useQueryParam(RequiredQueryParams.Execution, 0);
```

khelif96 (approved):
LGTM!

PR #255, opened 2024-07-17
DEVPROD-1130: Update flow of redeploying tag

sophstad:
DEVPROD-1130

### Description

- Add a little friction to prevent deployers from accidentally redeploying a tag: first require them to select "no" for cancellation, then confirm "yes" to redeploy the existing tag.
- Bonus: Use postversion script from deploy-utils package instead of local shell script for each app
- Fix documentation mentioning `--local` flag; it's now `--force`

### Testing

- Add unit test for deploy flow


packages/deploy-utils/src/prepare-prod-deploy/index.ts
```diff
@@ -33,10 +33,22 @@ export const prepareProdDeploy = async () => {
   // If there are no commit messages, ask the user if they want to delete and re-push the latest tag, thereby forcing a deploy with no new commits.
   if (commitMessages.length === 0) {
     const latestTag = getLatestTag(app);
+    const { value: cancelDeploy } = await prompts({
+      type: "confirm",
+      name: "value",
+      message: `No new commits. Do you want to cancel the deploy?`,
```
SupaJoon, line 39:
```suggestion
      message: "No new commits. Do you want to cancel the deploy?",
```

packages/deploy-utils/src/prepare-prod-deploy/prepare-prod-deploy.test.ts
```diff
@@ -93,9 +93,11 @@ my commit messages`,
     });
 
     it("redeploys tag when user confirms prompt", async () => {
-      vi.mocked(prompts).mockResolvedValueOnce({
-        value: true,
-      });
+      vi.mocked(prompts)
+        .mockResolvedValueOnce({ value: false })
```
SupaJoon, line 97: nit: would be helpful if there was a comment describing which prompt the resolved value refers to.

SupaJoon (approved):

PR #256, opened 2024-07-17
DEVPROD-8145: Fix project settings navigation

sophstad:
DEVPROD-8145

### Description

- To fix issues of state persisting between project and repo settings pages, set a `key` on the top-level `ProjectSettings` component. This prevents using loadable, but I think it's a worthwhile tradeoff. There were various states persisting between the components and this is the most straightforward way to get a clean slate
- Also fix a personal pet peeve regarding the location of the "Go to repo settings" button. It was not located next to the rest of the project navigation which I found confusing. It also links to the current tab now.
- Fix button overflow issues in header

### Screenshots

image


SupaJoon (commented):
Do you think there could be an issue could be with how `isRepo` is set or how `useProjectRedirect` works? Is this expected behavior from the lazy api?

sophstad:
> Do you think there could be an issue could be with how isRepo is set or how useProjectRedirect works?

Great callout, the bug was related to `isRepo` still being `true` when the user navigated back to a project. Updated with a simpler fix!

SupaJoon (approved):

PR #266, opened 2024-07-23
DEVPROD-8550: Fix overnight sleep schedule calculation

sophstad:
DEVPROD-8550

### Description

Fix calculation of sleep schedule when host is set to run overnight

### Testing

Add unit tests for calculator function


apps/spruce/src/components/Spawn/utils/hostUptime.test.ts
```diff
@@ -549,3 +550,89 @@ describe("getNextHostStart", () => {
     });
   });
 });
+
+describe("getEnabledHoursCount", () => {
```
khelif96, line 554: Are there any possible invalid inputs this function could get passed for example `stopTime` is before `startTime` that we should test against.

sophstad: `stopTime` before `startTime` is exactly what this PR implements to allow overnight schedules 😅

apps/spruce/src/components/Spawn/utils/hostUptime.ts
```diff
@@ -161,7 +161,17 @@ const getDailyUptime = ({
 }: {
   startTime: string;
   stopTime: string;
-}) => differenceInHours(new Date(stopTime), new Date(startTime));
+}) => {
+  const startDate = new Date(startTime);
+  const stopDate = new Date(stopTime);
+
+  // If this is an overnight schedule, set stop date to the following day so that uptime is correctly calculated.
+  if (stopDate < startDate) {
```
SupaJoon, line 169: How does this condition indicate an overnight schedule?

sophstad: If the stopTime occurs before the startTime, the back end interprets the start as occurring on the following day

SupaJoon (approved):

PR #268, opened 2024-07-24
DEVPROD-5302: Allow log search to work with sectioning

SupaJoon:
DEVPROD-5302

### Description
Typing a search term will initially open all sections where the term exists and jump to the first occurrence. Paginating through search results will open the section if it's closed.
`searchLogs` has been updated to consider lines that exist inside of closed sections and now returns the raw log index instead of the render index.
Two main helper functions: `useOpenSectionAndScrollToLine` and `openSectionContainingLine` contain the main logic for this feature.


apps/parsley/src/components/LogRow/BaseRow/index.tsx
```diff
@@ -73,7 +73,7 @@ const BaseRow: React.FC<BaseRowProps> = ({
 
   const bookmarked = bookmarks.includes(lineNumber);
   const failed = failingLine === lineNumber;
-  const highlighted = searchLine === lineIndex;
+  const highlighted = searchLine === lineNumber;
```
SupaJoon, line 76: log search now represent raw log index instead of render index

apps/parsley/src/utils/searchLogs/index.ts
```diff
@@ -13,34 +13,53 @@ interface searchOptions {
...
-      // Since processLogLines is ordered by line number, we can stop searching if we are out of range for our upper bound
-      if (upperBound && lineIndex > upperBound) {
-        break;
-      }
-      // If we are in range for our lower bound, we can start searching
-      if (lineIndex >= lowerBound) {
-        const line = getLine(lineIndex);
+      if (
+        rawLogIndex >= lowerBound &&
+        (upperBound ? upperBound >= rawLogIndex : true)
```
minnakt, line 51: Let's swap this because it's more intuitive:
```suggestion
        (upperBound ? rawLogIndex <= upperBound : true)
```

apps/parsley/src/utils/searchLogs/index.ts
```diff
@@ -13,34 +13,53 @@ interface searchOptions {
...
+ * @param options Object containing search parameters and helpers.
+ * @param options.searchRegex The regular expression to search for in the logs.
+ * @param options.upperBound Inclusive upper bound for the search.
+ * @param options.lowerBound Inclusive lower bound for the search.
+ * @param options.getLine Function to get the raw log string at the given index.
+ * @param options.processedLogLines Processed log lines that are currently rendered.
+ * @returns An array of sorted raw log indices that match the search criteria. SkippedLines are not included in the result.
+ */
 const searchLogs = (options: searchOptions): number[] => {
   const { getLine, lowerBound, processedLogLines, searchRegex, upperBound } =
```
minnakt, line 27: Maybe we should set the upperBound to make the if statements less tedious. even like `Number.MAX_VALUE`

apps/parsley/src/utils/searchLogs/index.ts
```diff
@@ -13,34 +13,53 @@ interface searchOptions {
...
-  for (let i = 0; i < processedLogLines.length; i++) {
-    const lineIndex = processedLogLines[i];
-    if (
+  const matchingLogIndex = new Set<number>();
+  for (let pLLIndex = 0; pLLIndex < processedLogLines.length; pLLIndex++) {
+    const rawLogIndex = processedLogLines[pLLIndex];
+    if (isSectionHeaderRow(rawLogIndex)) {
+      for (
+        let i = rawLogIndex.range.start;
+        i < rawLogIndex.range.end && (upperBound ? i <= upperBound : true);
```
minnakt, line 35: it might be more readable to break from the for-loop for surpassing the upper bound rather than as a condition in the for-loop itself.

apps/parsley/src/utils/searchLogs/index.ts
```diff
@@ -13,34 +13,53 @@ interface searchOptions {
...
-          matchingIndices.push(i);
+          matchingLogIndex.add(rawLogIndex);
         }
       }
     }
   }
-  return matchingIndices;
+  const result = Array.from(matchingLogIndex);
+  result.sort((a, b) => a - b);
+  return result;
```
minnakt, line 62: Could we return directly?
```suggestion
  return result.sort((a, b) => a - b);
```

apps/parsley/src/utils/searchLogs/index.ts
```diff
@@ -13,34 +13,53 @@ interface searchOptions {
...
-        const line = getLine(lineIndex);
+      if (
+        rawLogIndex >= lowerBound &&
+        (upperBound ? upperBound >= rawLogIndex : true)
+      ) {
+        const line = getLine(rawLogIndex);
         if (searchRegex.test(line)) {
-          // We want to match the index of the processedLogLines array, not the line number
-          matchingIndices.push(i);
+          matchingLogIndex.add(rawLogIndex);
```
minnakt, line 55: Could simplify to:
```
searchRegex.test(getLine(rawLogIndex))
```
(like you did above)

apps/parsley/src/context/LogContext/index.tsx
```diff
@@ -227,19 +229,20 @@ const LogContextProvider: React.FC<LogContextProviderProps> = ({
...
-  }, [
-    state.searchState.searchTerm,
-    processedLogLines,
-    upperRange,
-    lowerRange,
-    getLine,
-    dispatch,
-  ]);
+    // processedLogLines doesn't need to be in the dependency array because the search results would
+    // only change if a filter is applied -- and a filter and search term can't be applied at the same time.
```
minnakt, line 233: comment might be misleading because you can apply a filter and then apply a search on top of that at the same time

minnakt (changes requested):
thanks for writing a lot of tests!

apps/parsley/src/hooks/useSections/index.ts
```diff
@@ -108,32 +116,37 @@ export const useSections = ({
...
+    if (!sectionData || !sectionState) {
+      return false;
+    }
+    const nextState = openSectionContainingLineNumberHelper({
+      lineNumber,
+      sectionData,
+      sectionState,
+    });
+    setSectionState(nextState);
+    return nextState !== sectionState;
```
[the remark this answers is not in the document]
SupaJoon: I like it this way mainly because of code style/avoid an extra condition. This doesn't have a perf implication since the equality check is reference based.

apps/parsley/src/hooks/useSections/index.ts
```diff
@@ -108,32 +116,37 @@ export const useSections = ({
...
+   */
+  const openSectionContainingLineNumber = ({
+    lineNumber,
+  }: {
+    lineNumber: number | number[];
+  }): boolean => {
+    if (!sectionData || !sectionState) {
+      return false;
+    }
+    const nextState = openSectionContainingLineNumberHelper({
```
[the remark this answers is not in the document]
SupaJoon: This logic could actually exist inside of the `openSectionContainingLineNumber` hook function but it was too big and I wanted to test it. It's not really meant to be used outside of that hook function so that's why I added `...Helper`. I think `calculateNextSectionState` wouldn't capture that this function will return the next state with an open section based on the input line number. 
There is a function called `populateSections` and I thought about combining these two but I think it may complicate use cases and testing.

apps/parsley/src/hooks/useSections/utils.ts
```diff
@@ -150,5 +152,79 @@ const parseSections = (logs: string[]): SectionData => {
...
+        sectionState[functionID].commands[commandID].isOpen !== true ||
+        sectionState[functionID].isOpen !== true
+      ) {
+        diff = true;
+      }
+      nextState[functionID].commands[commandID].isOpen = true;
+      nextState[functionID].isOpen = true;
+    });
+  }
+  return diff ? nextState : sectionState;
```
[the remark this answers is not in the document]
SupaJoon: This function needs to return sectionState so it can be set with `setSectionState`

apps/parsley/src/hooks/useOpenSectionAndScrollToLine/index.ts
```diff
@@ -0,0 +1,54 @@
...
+        );
+        scroll(scrollIndex);
+        setLineNumber(undefined);
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [lineNumber]);
+
+  useEffect(() => {
+    if (lineNumber !== undefined) {
```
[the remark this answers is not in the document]
SupaJoon: they can't be combined unless previous value is tracked.

apps/parsley/src/hooks/useSections/utils.ts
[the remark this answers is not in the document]
SupaJoon: Ah, I understand, that will make this function easier to use.

apps/parsley/src/hooks/useSections/utils.ts
```diff
@@ -150,5 +152,79 @@ const parseSections = (logs: string[]): SectionData => {
...
+  sectionData: SectionData;
+  sectionState: SectionState;
+  lineNumber: number | number[];
+}) => {
+  const lineNumberArray = Array.isArray(lineNumber) ? lineNumber : [lineNumber];
+  const sectionContainingLine = lineNumberArray
+    .map((n) =>
+      sectionData?.commands.find((section) => includesLineNumber(section, n)),
+    )
+    .filter((v) => v) as CommandEntry[];
```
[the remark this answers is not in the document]
SupaJoon: fixed with your refactor suggestion

apps/parsley/src/hooks/useSections/utils.ts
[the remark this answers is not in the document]
SupaJoon: I updated the code to not return the same reference as the previous state. This is probably safer to avoid mutating data structures accidentally.

apps/parsley/src/context/LogContext/index.tsx
```diff
@@ -227,19 +229,21 @@ const LogContextProvider: React.FC<LogContextProviderProps> = ({
...
+  const openSectionAndScrollToLine = useOpenSectionAndScrollToLine(
+    processedLogLines,
+    sectioning.openSectionContainingLineNumber,
+    scrollToLine,
+  );
   const stringifiedSearchResults = searchResults.toString();
   useEffect(() => {
     if (searchResults.length > 0) {
-      scrollToLine(searchResults[0]);
+      openSectionAndScrollToLine(searchResults);
```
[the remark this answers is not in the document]
SupaJoon: `openSectionAndScrollToLine` needs all of the search results so it can open all of the sections that contain the search results.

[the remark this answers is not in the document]
minnakt: Thanks for explaining, now I understand why I was so confused by `useOpenSectionAndScrollToLine`. I think it's pretty confusing how `lineNumber` can be equal to an array of numbers (to open the sections) and also a single number (to scroll to). Besides the singular naming of `lineNumber`, it seems like to open the sections we could always expect an array of numbers and to scroll we could always expect a single number. Is it possible to separate that logic by using two separate variables or something otherwise?

minnakt (commented):
my only remaining confusion is with `useOpenSectionAndScrollToLine`! the other functions make sense to me

apps/parsley/src/context/LogContext/index.tsx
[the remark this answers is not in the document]
SupaJoon: Opening sections won't always expect an array of numbers. These are the use cases:
-clicking on a bookmark opens 1 section and jumps to the bookmark line
-paginating search opens 1 section and jumps to the search result
-new search input opens all sections that contain the search result and jumps to the first search result. 
It's possible to separate the two, but the interface will become less clean in my opinion and the functions `UseSectionsResult.openSectionContainingLineNumber` is documented with JSDoc and an interface. 
Updating the interface will result in maintaining an object that looks like this: `{lineNumberToScrollTo: number, lineNumbersToOpen: number[]}` where the usage would look like this most the time:
```openSectionAndScrollToLine({lineNumbersToOpen: [searchResults[nextPage]], lineNumberToScrollTo:searchResults[nextPage]})````
instead of:
```openSectionAndScrollToLine(searchResults[nextPage])```

File: apps/parsley/src/hooks/useOpenSectionAndScrollToLine/index.ts [hunk 14]
```diff
@@ -0,0 +1,64 @@
...
+ * @param scrollToLine This function is used to scroll to the line number.
+ * @returns A setter function that accepts one or many line numbers so scroll to. If multiple lines numbers are passed, open
+ * their corresponding sections and scroll to the first line.
+ */
+const useOpenSectionAndScrollToLine = (
+  processedLogLines: ProcessedLogLines,
+  openSectionContainingLineNumber: UseSectionsResult["openSectionContainingLineNumber"],
+  scrollToLine: (lineNumber: number) => void,
+) => {
+  const [lineNumber, setLineNumber] = useState<number | number[] | undefined>();
```
minnakt, line 21: Could we rename this variable? It's quite confusing that it can be singular or plural. `lineNumber` implies that it would be a single number

apps/parsley/src/hooks/useSections/utils.ts
```diff
@@ -150,5 +153,76 @@ const parseSections = (logs: string[]): SectionData => {
...
+  sectionState,
+}: {
+  sectionData: SectionData;
+  sectionState: SectionState;
+  lineNumber: number | number[];
+}): [boolean, SectionState] => {
+  const lineNumberArray = conditionalToArray(lineNumber, true);
+  const sectionContainingLine = sectionData.commands.filter((section) =>
+    lineNumberArray.some((n) => includesLineNumber(section, n)),
+  );
```
minnakt, line 178: is it possible for this function to just accept a `number[]` and do the conversion in the parent function?
```suggestion
  lineNumbers: number[];
}): [boolean, SectionState] => {
  const sectionContainingLine = sectionData.commands.filter((section) =>
    lineNumbers.some((n) => includesLineNumber(section, n)),
  );
```

apps/parsley/src/hooks/useOpenSectionAndScrollToLine/index.ts
minnakt: Based on the conversation above, is there a reason why it can't always just be `number[]`?

SupaJoon: This can be set to array. I think it's convenient that this function accepts a single number or number[] to avoid having to cast to an array in other parts of the application. I pushed up the array casting into this hook as separate setter function. Let me know what you think about this commit: https://github.com/evergreen-ci/ui/pull/268/commits/67f70f89ed798e315b6dc42549e292f4c0f80681

apps/parsley/src/hooks/useSections/utils.ts
SupaJoon: I pushed the conversion all the way up into the hook in a setting function separate from the `useState`

apps/parsley/src/hooks/useOpenSectionAndScrollToLine/index.ts
```diff
@@ -0,0 +1,67 @@
...
+  scrollToLine: (lineNumber: number) => void,
+) => {
+  const [lineNumbers, setLineNumbers] = useState<number[] | undefined>();
+
+  /**
+   * When the lineNumber is set from the callback, calculate the next section state
+   * from the lineNumber. If the next state is the same as the current state,
+   * scroll to the lineNumber and reset lineNumber. If the next state is different,
+   * wait for processedLogLines to update in the following useEffect.
+   */
```
minnakt, line 29: nit. update comment `lineNumber` => `lineNumbers`

apps/parsley/src/hooks/useOpenSectionAndScrollToLine/index.ts
```diff
@@ -0,0 +1,67 @@
...
+        scrollToLine(scrollIndex);
+        setLineNumbers(undefined);
+      }
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [lineNumbers]);
+
+  /**
+   * When processedLogLines updates and lineNumber is defined,
+   * scroll to the lineNumber and reset lineNumber.
```
minnakt, line 46: nit. update comment `lineNumber` => `lineNumbers`

apps/parsley/src/hooks/useOpenSectionAndScrollToLine/index.ts
```diff
@@ -0,0 +1,67 @@
...
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [processedLogLines]);
+
+  const openSectionAndScrollToLine = useCallback(
+    (lineNumberInput: number | number[]) => {
+      setLineNumbers(conditionalToArray(lineNumberInput, true));
+    },
+    [],
+  );
```
minnakt, line 62: omg yay! I think this is a lot easier to understand :)

apps/parsley/src/hooks/useSections/index.ts
```diff
@@ -108,32 +114,39 @@ export const useSections = ({
     [],
   );
 
+  /**
+   * openSectionsContainingLineNumbers Will update the current section state. If the next state is the
+   * same as the current state the function will return true and false otherwise.
+   * @param param0 is an object with a lineNumber key(s).
+   * @param param0.lineNumbers is a number or an array of numbers that represent raw log line numbers.
```
minnakt, line 121: (maybe this is a good way to name the parameter)
```suggestion
   * @param options is an object with a lineNumber key(s).
   * @param options.lineNumbers is a number or an array of numbers that represent raw log line numbers.
```

apps/parsley/src/hooks/useSections/index.ts
```diff
@@ -108,32 +114,39 @@ export const useSections = ({
...
+  const openSectionsContainingLineNumbers = ({
+    lineNumbers,
+  }: {
+    lineNumbers: number[];
+  }): boolean => {
+    if (!sectionData || !sectionState) {
+      return false;
+    }
+    const [hasDiff, nextState] =
+      getSectionStateWithOpenSectionBasedOnLineNumber({
```
minnakt, line 133: `lineNumbers` maybe!
```suggestion
      getSectionStateWithOpenSectionBasedOnLineNumbers({
```
another good name for this function could be `getOpenSectionStateBasedOnLineNumbers`

minnakt (approved):
thanks for the refactor! I think it looks great!

PR #270, opened 2024-07-25
DEVPROD-9224: Enable Cypress local development with image visibility API

minnakt:
DEVPROD-9224

### Description
Makes it possible to run Cypress tests that interact with the Image Visibility API locally.

### Context
The Image Visibility API credentials are stored in the admin collection in the database, which is read from `testdata/local/admin.json` in the Evergreen repo. We are unable to add the credentials to this file because that would mean we have to publish the credentials to the repo, which would leak them.

As a stopgap, we can set environment variables and use those to write the credentials directly to the database. Writing the credentials directly to the database means that we don't have to update the `admin.json` file and the credentials will be available for testing. 

Maybe in the future when we have a secrets manager we can re-evaluate how to do this in a way that doesn't require you to manually update your `zsh` with environment variables.

### Screenshots

How it looks in the shell:

Screenshot 2024-07-25 at 10 13 07 AM

### Testing
- Confirmed I can run Sophia's existing tests locally

### FYI
The Image Visibility API is fully configured on staging so you can interact with it via `yarn staging`, even if you don't have any credentials.


scripts/evg-db-ops.sh
```diff
@@ -20,8 +22,21 @@ reseed_database() {
...
+
     # Load test data into the database.
     ../bin/load-smoke-data -path ../testdata/local -dbName evergreen_local -amboyDBName amboy_local
+
+    cd -
+
+    # Load credentials for the Image Visibility API.
+    echo "Adding runtime environments credentials..."
+    if [ "$CI" != 'true' ]; then
+        mongosh $DB_NAME --quiet $RUNTIME_ENVIRONMENTS_SCRIPT_PATH
```
khelif96, line 34: Can we add some error handling here in case the user does not have mongosh?

khelif96 (approved):

PR #269, opened 2024-07-24
DEVPROD-5301: Allow Bookmarks to work with sectioning

SupaJoon:
DEVPROD-5301

### Description
Clicking a Bookmark will now open a section if necessary by utilizing `logContext.openSectionAndScrollToLine`

PR #271, opened 2024-07-25
DEVPROD-5297: Allow multiple selection menu to work with sections

SupaJoon:
DEVPROD-5297

### Description
These code changes are based off of https://github.com/evergreen-ci/ui/pull/269.
This PR works out the SharingMenu open/close state. Originally, the SharingMenu component was using a MultiLineSelectContext hook to manage default open state and then an internal component hook to manage state further. When the context state is open and internal state is closed, edge case occurs where the SharingMenu is opened every time ProcessedLogLines changed (when a section is toggled) since the default open/close state applied every time the log line containing the SharingMenu mounted (SharingMenu is mounted within a log line).
I updated the code to rely on the MultiLineSelectContext directly and removed the internal component state.


sophstad (approved):
Nice!

PR #283, opened 2024-07-31
DEVPROD-7656: Create Packages table on Build Information page

sophiabarness:
DEVPROD-7656

### Description

Created the Packages table on the Build Information page. It supports pagination and filtering.

### Screenshots

Packages Table

### Testing

Added Jest tests to ensure table is properly rendered.


apps/spruce/src/pages/container/EventsTable/index.tsx
```diff
@@ -79,7 +79,6 @@ const EventsTable: React.FC<{}> = () => {
     defaultColumn: {
       enableColumnFilter: false,
     },
-    manualPagination: true,
```
minnakt, line 82: Was this change intentional? It looks like the page is unrelated to the packages table

apps/spruce/src/pages/image/PackagesTable/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+  ImagePackagesQuery,
+  ImagePackagesQueryVariables,
+} from "gql/generated/types";
+import { IMAGE_PACKAGES } from "gql/queries";
+import { useFirstImage } from "hooks";
+
+export const PackagesTable: React.FC = () => {
+  const { [slugs.imageId]: imageId } = useParams();
+  const { image: firstImage } = useFirstImage();
+  const selectedImage = imageId ?? firstImage;
```
minnakt, line 27: let's pass down imageId as props so you don't have to call `useFirstImage`

apps/spruce/src/pages/image/PackagesTable/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const table = useLeafyGreenTable<Package>({
+    columns,
+    data: packageEntries,
+    containerRef: tableContainerRef,
+    defaultColumn: {
+      enableColumnFilter: false,
+    },
+    manualPagination: true,
+    manualFiltering: true,
+    rowCount: totalNumPackages,
```
minnakt, line 98: just curious, does this `rowCount` do anything? It seems like it's handled by `Pagination`

apps/spruce/src/pages/image/PackagesTable/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+      columnFilters,
+    },
+    onColumnFiltersChange: setColumnFilters,
+    onPaginationChange: setPagination,
+  });
+
+  return (
+    <>
+      <SettingsCardTitle>Packages</SettingsCardTitle>
+      <SettingsCard data-cy="packages-table-card">
```
minnakt, line 110: let's remove `SettingsCardTitle` and `SettingsCard` from this component; I think this component should just contain the table

apps/spruce/src/pages/image/PackagesTable/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+        <BaseTable
+          data-cy-row="packages-table-row"
+          table={table}
+          shouldAlternateRowColor
```
minnakt, line 114: can we adding `loading` and `loadingRows`?

apps/spruce/src/pages/image/PackagesTable/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+            }}
+            numTotalItems={totalNumPackages}
+            currentPage={table.getState().pagination.pageIndex + 1}
+            onCurrentPageOptionChange={(value: string) => {
+              table.setPageIndex(Number(value) - 1);
+            }}
+            onBackArrowClick={() => table.previousPage()}
+            onForwardArrowClick={() => table.nextPage()}
+          />
+        </PaginationWrapper>
```
minnakt, line 130: it would be nice if we added this component to the `BaseTable` component so it could be reused for future tables (and then we could remove this code from this component)

apps/spruce/src/pages/image/PackagesTable/PackagesTable.test.tsx
```diff
@@ -0,0 +1,315 @@
...
+} from "test_utils";
+import { ApolloMock } from "types/gql";
+import { PackagesTable } from ".";
+
+describe("packages table", () => {
+  it("shows name field data", async () => {
+    const { Component } = RenderFakeToastContext(
+      <MockedProvider mocks={[imagePackagesPageOneMock]}>
+        <PackagesTable />
+      </MockedProvider>,
```
minnakt, line 23: instead of redeclaring this in every test, can we have a `wrapper` component that gets passed to the `render` function? Similar to what you did in the tests for the Event Log

apps/spruce/src/pages/image/PackagesTable/PackagesTable.test.tsx
```diff
@@ -0,0 +1,315 @@
...
+    );
+    render(<Component />);
+    await waitFor(() => {
+      expect(screen.getByDataCy("packages-table-card")).toBeInTheDocument();
+    });
+    const card = screen.getByDataCy("packages-table-card");
+    const expectedNames = [
```
minnakt, line 80:
```suggestion
    const expectedVersions = [
```

minnakt (changes requested):
great tests!

apps/spruce/src/pages/image/PackagesTable/index.tsx
sophiabarness: The docs seem to suggest passing it in if we know the rowCount because it makes calculating the total number of entries easier for tanstack.
[Docs](https://tanstack.com/table/latest/docs/guide/pagination#pagination-row-model)

apps/spruce/src/pages/image/PackagesTable/index.tsx
sophiabarness: This seems like a good idea. I'm just a bit worried about how it would impact the BaseTables already in the code. For instance, the User Subscriptions table does pagination on top of the table, while this component does pagination below the table. Would there be a simple way of defining the order of the table and pagination?

minnakt: You can use pass a prop that defines whether to use the pagination comment e.g. `usePagination` and it can be set to `false` by default to not impact any existing tables

apps/spruce/src/pages/image/PackagesTable/PackagesTable.test.tsx
```diff
@@ -0,0 +1,316 @@
...
+    }
+  });
+
+  it("shows version field data", async () => {
+    const { Component } = RenderFakeToastContext(
+      <MockedProvider mocks={[imagePackagesPageOneMock]}>
+        <PackagesTable imageId="ubuntu2204" />
+      </MockedProvider>,
+    );
+    render(<Component />);
```
minnakt, line 83: You can reuse the wrapper defined above

apps/spruce/src/pages/image/PackagesTable/PackagesTable.test.tsx
```diff
@@ -0,0 +1,316 @@
...
+  it("supports name filter", async () => {
+    const user = userEvent.setup();
+    const { Component } = RenderFakeToastContext(
+      <MockedProvider
+        mocks={[imagePackagesPageOneMock, imagePackagesNameFilterMock]}
+      >
+        <PackagesTable imageId="ubuntu2204" />
+      </MockedProvider>,
+    );
+    render(<Component />);
```
minnakt, line 116: same comment

apps/spruce/src/pages/image/PackagesTable/PackagesTable.test.tsx
```diff
@@ -0,0 +1,316 @@
...
+  it("supports pagination", async () => {
+    const user = userEvent.setup();
+    const { Component } = RenderFakeToastContext(
+      <MockedProvider
+        mocks={[imagePackagesPageOneMock, imagePackagesPageTwoMock]}
+      >
+        <PackagesTable imageId="ubuntu2204" />
+      </MockedProvider>,
+    );
+    render(<Component />);
```
minnakt, line 139: same comment

apps/spruce/src/pages/image/PackagesTable/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+  ImagePackagesQuery,
+  ImagePackagesQueryVariables,
+} from "gql/generated/types";
+import { IMAGE_PACKAGES } from "gql/queries";
+
+type PackagesTableProps = {
+  imageId: string;
+};
+
+const PAGE_SIZE = 10;
```
minnakt, line 24: There's an existing constant called `DEFAULT_PAGE_SIZE` which you can import here

apps/spruce/src/pages/image/PackagesTable/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+    },
+    {
+      header: "Manager",
+      accessorKey: "manager",
+    },
+    {
+      header: "Version",
+      accessorKey: "version",
+    },
+  ];
```
minnakt, line 85: We should move the columns outside of the component, since it doesn't depend on any of the code within the component

apps/spruce/src/pages/image/PackagesTable/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+  });
+
+  return (
+    <>
+      <BaseTable
+        data-cy-row="packages-table-row"
+        table={table}
+        shouldAlternateRowColor
+        loading={loading}
+        loadingRows={PAGE_SIZE}
```
minnakt, line 113: I think this should possibly be `pagination.pageSize`; there's no guarantee the user will always be using the default page size of 10

apps/spruce/src/pages/image/PackagesTable/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+      accessorKey: "manager",
+    },
+    {
+      header: "Version",
+      accessorKey: "version",
+    },
+  ];
+
+  const tableContainerRef = useRef<HTMLDivElement>(null);
+  const table = useLeafyGreenTable<Package>({
```
minnakt, line 88: There's a bug where if you go forward a few pages, and then search for a valid package, it doesn't reset you to the 0th page and it seems like there's no results for the search. 

There might be a setting on the table that you can add to handle this, or you can add some custom code to handle resetting the page index

apps/spruce/src/pages/image/PackagesTable/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+  const packageEntries = useMemo(
+    () => packagesData?.image?.packages.data ?? [],
+    [packagesData?.image?.packages.data],
+  );
+
+  const totalNumPackages = useMemo(
+    () => packagesData?.image?.packages.totalCount ?? 0,
+    [packagesData?.image?.packages.totalCount],
+  );
```
minnakt, line 61: The count doesn't appear to be correct when a filter is applied:

Screenshot 2024-08-15 at 7 36 07 PM

(Specifically there shouldn't be 103 pages, and we shouldn't be able to click the Next button)

minnakt (changes requested):
(responded to your question about how to add `Pagination` to `BaseTable` in the thread above)

apps/spruce/src/pages/image/PackagesTable/PackagesTable.test.tsx
```diff
@@ -0,0 +1,316 @@
...
+            {
+              manager: "pip 22.0.2 from (python 3.10)",
+              name: "bcrypt",
+              version: "3.2.0",
+            },
+          ],
+          filteredCount: 1,
+          totalCount: 15,
+        },
+      },
```
minnakt, line 259: The reason why you're seeing GraphQL errors when you run the tests is because you need to include `__typename` in your mocks, like this:
```suggestion
 image: {
        __typename: "Image",
        id: "ubuntu2204",
        packages: {
           __typename: "ImagePackagesPayload",
          data: [
            {
              __typename: "Package",
              manager: "pip 22.0.2 from (python 3.10)",
              name: "bcrypt",
              version: "3.2.0",
            },
          ],
          filteredCount: 1,
          totalCount: 15,
        },
      },
```

apps/spruce/src/pages/image/PackagesTable/index.tsx
sophiabarness: Got it - added it along with a total count parameter to BaseTable

minnakt (changes requested):

apps/spruce/src/components/Table/BaseTable.tsx
```diff
@@ -253,3 +286,10 @@ const StyledExpandedContent = styled(ExpandedContent)`
     flex-grow: 1;
   }
 ` as typeof ExpandedContent;
+
+// @ts-expect-error
+const StyledPagination = styled(Pagination)`
+  display: flex;
+  justify-content: space-between;
```
minnakt, line 293: Seems like these styles are not necessary:
```suggestion
```

apps/spruce/src/components/Table/BaseTable.tsx
```diff
@@ -80,6 +86,18 @@ export const BaseTable = forwardRef(
...
+    // Auto-redirect user to first page when filter applied.
+    useEffect(() => {
+      if (
+        table &&
+        table.getRowModel().rows.length === 0 &&
+        table.getRowCount() !== 0
+      ) {
+        table.firstPage();
+      }
+    }, [table]);
```
minnakt, line 99: rather than using `useEffect`, can you return to the first page exactly when the filter is applied?

apps/spruce/src/components/Table/BaseTable.tsx
```diff
@@ -57,6 +58,9 @@ type SpruceTableProps = {
   loading?: boolean;
   /** estimated number of rows the table will have */
   loadingRows?: number;
+  usePagination?: boolean;
+  /** number of total items the table will have */
+  numTotalItems?: number;
   /** rows that will have a blue tint to represent that they are selected */
   selectedRowIndexes?: number[];
```
minnakt, line 65: nit. Let's keep the props in alphabetical order

apps/spruce/src/pages/image/PackagesTable/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+      pagination,
+      columnFilters,
+    },
+    onColumnFiltersChange: setColumnFilters,
+    onPaginationChange: setPagination,
+  });
+
+  if (table.getRowModel().rows.length === 0 && table.getRowCount() !== 0) {
+    table.firstPage();
+  }
```
minnakt, line 81: It looks like you forgot to delete this:
```suggestion
```

apps/spruce/src/pages/image/PackagesTable/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+  const packageEntries = useMemo(
+    () => packagesData?.image?.packages.data ?? [],
+    [packagesData?.image?.packages.data],
+  );
+
+  const numPackages = useMemo(
+    () => packagesData?.image?.packages.filteredCount ?? 0,
+    [packagesData?.image?.packages.filteredCount],
+  );
```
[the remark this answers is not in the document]
minnakt: Did you apply the fall back like this?
`packagesData?.image?.packages.filteredCount ?? packagesData?.image?.packages.totalCount`

apps/spruce/src/pages/image/PackagesTable/PackagesTable.test.tsx
```diff
@@ -0,0 +1,318 @@
...
+      expect(screen.queryAllByDataCy("packages-table-row")).toHaveLength(10);
+    });
+    const expectedVersions: string[] = [];
+    for (let i = 0; i < 10; i++) {
+      const packageData =
+        imagePackagesPageOneMock.result?.data?.image?.packages.data[i];
+      if (packageData && packageData.version) {
+        expectedVersions.push(packageData.version);
+      }
+    }
```
minnakt, line 86: I think you could make this look a little cleaner by using the [`map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map) function.

apps/spruce/src/pages/image/PackagesTable/PackagesTable.test.tsx
```diff
@@ -0,0 +1,318 @@
...
+      expect(screen.queryAllByDataCy("packages-table-row")).toHaveLength(10);
+    });
+    const expectedNames: string[] = [];
+    for (let i = 0; i < 10; i++) {
+      const packageData =
+        imagePackagesPageOneMock.result?.data?.image?.packages.data[i];
+      if (packageData && packageData.name) {
+        expectedNames.push(packageData.name);
+      }
+    }
```
minnakt, line 46: same comment

apps/spruce/src/components/Table/BaseTable.tsx
```diff
@@ -1,6 +1,7 @@
-import { ForwardedRef, forwardRef } from "react";
+import { ForwardedRef, forwardRef, useEffect } from "react";
```
minnakt, line 1: (comment placement is random) Could we write a Cypress integration test?

minnakt (changes requested):

apps/spruce/src/pages/image/PackagesTable/index.tsx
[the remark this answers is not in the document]
sophiabarness: Oops got confused - this comment makes more sense now

apps/spruce/cypress/integration/image/build_information.ts
```diff
@@ -6,4 +6,11 @@ describe("build information", () => {
       cy.contains("ubuntu1804-workstation");
     });
   });
+
+  describe("packages", () => {
+    it("should show the corresponding packages", () => {
+      cy.visit("/image/ubuntu2204");
+      cy.dataCy("packages-table-row").should("have.length", 10);
+    });
```
minnakt, line 14: Is there anything else we can test here like applying a search, pagination? The reason the distros table test is so short is because it has no search or pagination functionalities

apps/spruce/src/components/Table/BaseTable.tsx
```diff
@@ -57,6 +58,9 @@ type SpruceTableProps = {
   loading?: boolean;
   /** estimated number of rows the table will have */
   loadingRows?: number;
+  /** number of total items the table will have */
+  numTotalItems?: number;
+  usePagination?: boolean;
   /** rows that will have a blue tint to represent that they are selected */
   selectedRowIndexes?: number[];
```
minnakt, line 65: nit. to be in alphabetical order, `usePagination` should come after `selectedRowIndexes`

apps/spruce/src/pages/image/PackagesTable/PackagesTable.test.tsx
```diff
@@ -0,0 +1,314 @@
...
+    );
+    render(<Component />, { wrapper });
+    await waitFor(() => {
+      expect(screen.queryAllByDataCy("packages-table-row")).toHaveLength(10);
+    });
+    const expectedNames = (
+      imagePackagesPageOneMock.result?.data?.image?.packages.data || []
+    )
+      .slice(0, 10)
+      .map((packageData) => packageData?.name);
```
minnakt, line 43:
```suggestion
    const expectedNames = (
      imagePackagesPageOneMock.result?.data?.image?.packages.data || []
    ).map(({ name }) => name);
```

apps/spruce/src/pages/image/PackagesTable/PackagesTable.test.tsx
```diff
@@ -0,0 +1,314 @@
...
+    );
+    render(<Component />, { wrapper });
+    await waitFor(() => {
+      expect(screen.queryAllByDataCy("packages-table-row")).toHaveLength(10);
+    });
+    const expectedVersions = (
+      imagePackagesPageOneMock.result?.data?.image?.packages.data || []
+    )
+      .slice(0, 10)
+      .map((packageData) => packageData?.version);
```
minnakt, line 81:
```suggestion
    const expectedVersions = (
      imagePackagesPageOneMock.result?.data?.image?.packages.data || []
    ).map(({ version }) => version);
```

apps/spruce/src/pages/image/PackagesTable/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+      packagesData?.image?.packages.filteredCount <
+        packagesData?.image?.packages.totalCount
+    ) {
+      table.firstPage();
+    }
+  }, [
+    table,
+    packagesData?.image?.packages.filteredCount,
+    packagesData?.image?.packages.totalCount,
+  ]);
```
minnakt, line 99: I think that the logic to reset the page should be inside of `BaseTable`, not outside of it. If the logic was kept outside, that means every time we use this component we would need to rewrite this code every time.

In `BaseTable`, the filter logic is present in the headers. Can we reset the page when a filter is applied, somewhere in the code in `BaseTable`?

minnakt (changes requested):

apps/spruce/src/components/Table/BaseTable.tsx
```diff
@@ -123,9 +129,12 @@ export const BaseTable = forwardRef(
...
                           <TableSearchPopover
                             data-cy={meta?.search?.["data-cy"]}
-                            onConfirm={(value) =>
-                              header.column.setFilterValue(value)
-                            }
+                            onConfirm={(value) => {
+                              header.column.setFilterValue(value);
+                              if (table) {
+                                table.firstPage();
+                              }
```
minnakt, line 136: I think it would be:
```suggestion
                              if (usePagination) {
                                table?.firstPage();
                              }
```
Could you apply this to the `TableFilterPopover` as well?

apps/spruce/cypress/integration/image/build_information.ts
```diff
@@ -6,4 +6,41 @@ describe("build information", () => {
...
+          cy.dataCy("packages-table-row").should("have.length", 10);
+          cy.dataCy("packages-table-row")
+            .first()
+            .find("td")
+            .first()
+            .invoke("text")
+            .should((nextPackageName) => {
+              expect(nextPackageName).not.to.equal(firstPackageName);
+            });
+        });
```
minnakt, line 35: I think it might be clearer to write it like this, including comments
```suggestion
      // First package name on first page.
      cy.dataCy("packages-table-row")
        .first()
        .find("td")
        .first()
        .invoke("text")
        .as("firstPackageName", { type: "static" });

      cy.get("[data-testid=lg-pagination-next-button]").click();

      // First package name on second page.
      cy.dataCy("packages-table-row")
        .first()
        .find("td")
        .first()
        .invoke("text")
        .as("nextPackageName", { type: "static" });

      // Package names should not be equal.
      cy.get("@firstPackageName").then((firstPackageName) => {
        cy.get("@nextPackageName").then((nextPackageName) => {
          expect(nextPackageName).not.to.equal(firstPackageName);
        });
      });
```

minnakt (approved):

PR #287, opened 2024-08-01
DEVPROD-6351: Provide visual feedback for downstream projects

sophiabarness:
DEVPROD-6351

### Description
Added badge to Downstream Tasks tab showing how many child patches have failed, started, or succeeded. 

### Screenshots

Succeeded
Started


apps/spruce/src/pages/version/VersionTabs.tsx
```diff
@@ -20,20 +20,59 @@ const { parseQueryString } = queryString;
 interface Props {
   taskCount: number;
   isPatch: boolean;
-  // @ts-expect-error: FIXME. This comment was added by an automated script.
-  childPatches: VersionQuery["version"]["patch"]["childPatches"];
+  childPatches: Array<Patch>;
```
khelif96, line 23: This isn't actually the type this query is getting. `Patch` represents all of the fields that exist in the patch graphql struct. While `VersionQuery["version"]["patch"]["childPatches"]` represents the fields we are actually querying.

khelif96: You can type ChildPatches like 
```ts
type ChildPatches = NonNullable<
  VersionQuery["version"]["patch"]
>["childPatches"];
```

apps/spruce/src/pages/Version.tsx
```diff
@@ -229,7 +229,7 @@ export const VersionPage: React.FC = () => {
         <PageLayout>
           <PageContent>
             <VersionTabs
-              childPatches={childPatches}
+              {...(childPatches ? { childPatches } : {})}
```
khelif96, line 232: This is a little confusing to read. I think if you use the typing from below you can avoid this logic.

khelif96 (commented):
Great job just had a few suggestions on typing

apps/spruce/src/pages/version/VersionTabs.tsx
```diff
@@ -20,20 +20,59 @@ const { parseQueryString } = queryString;
 interface Props {
   taskCount: number;
   isPatch: boolean;
-  // @ts-expect-error: FIXME. This comment was added by an automated script.
-  childPatches: VersionQuery["version"]["patch"]["childPatches"];
+  childPatches: NonNullable<VersionQuery["version"]["patch"]>["childPatches"];
```
khelif96, line 23: nit. You can extract this into its own type and reuse it below instead of repeating the code

khelif96 (approved):
Great work! Just had a small nit 🚀

PR #288, opened 2024-08-02
DEVPROD-9195: Enlarge patch parameters

sophiabarness:
DEVPROD-9195

### Description

Enlarged patch parameters by changing CSS properties. 

### Screenshots
TextBoxesArea


minnakt (commented):
Can we turn the Value field into a Textbox? Sometimes it can be quite long, like a URL

apps/spruce/src/components/EditableTagField/TagRow.tsx
```diff
@@ -138,6 +138,7 @@ const FlexColumnContainer = styled.div`
   display: flex;
   flex-direction: column;
   margin-right: ${size.xs};
+  flex-grow: 1;
 `;
```
minnakt, line 143: let's change the button text from title case to sentence case: `Add Parameter` => `Add parameter`. This is based on the guidance of the design team

apps/spruce/src/components/EditableTagField/TagRow.tsx
```diff
@@ -122,12 +114,13 @@ export const TagRow: React.FC<TagRowProps> = ({
...
         </ButtonContainer>
       )}
-    </FlexColumnContainer>
+    </>
   );
 };
 
 const ButtonContainer = styled.div`
   margin-top: ${size.m};
+  margin-bottom: ${size.xxs};
```
minnakt, line 123: The margin underneath seems to not be needed
```suggestion
```

minnakt (approved):
thanks for cleaning up the HTML!

nit about PR title: Let's use sentence case: `Enlarge patch parameters`

PR #289, opened 2024-08-02
DEVPROD-3969: Make project settings header sticky

sophstad:
DEVPROD-3969

### Description

Add sticky header to project settings pages. The shadow appears once a user has scrolled down a bit. This should reduce the amount of scrolling back and forth, especially on lengthy pages like variables.

If we get no/positive feedback, we can apply the same on distro settings pages!

### Screenshots

image


apps/spruce/src/pages/projectSettings/index.tsx
```diff
@@ -49,6 +51,24 @@ const ProjectSettings: React.FC = () => {
...
+      if (!pageWrapperRef?.current) return;
+      if (pageWrapperRef?.current?.scrollTop < 40) {
+        setAtTop(true);
+      } else {
+        setAtTop(false);
+      }
+    }, 250);
+    pageWrapperRef?.current?.addEventListener("scroll", onScroll);
+    return () =>
+      pageWrapperRef?.current?.removeEventListener("scroll", onScroll);
```
khelif96, line 69: I see the following react-hooks warning
image
Doing what it suggests seems to work and I was able to confirm the behavior

khelif96 (approved):
Looks great!

PR #302, opened 2024-08-09
DEVPROD-9562: Create distros table on Build Information page

minnakt:
DEVPROD-9562

### Description
Adds the distros table to the image page. 

I think we're missing some logic in the backend to hide admin-only distros for non-admin users — filed [DEVPROD-9743](https://jira.mongodb.org/browse/DEVPROD-9743) to fix that.

### Screenshots

Screenshot 2024-08-09 at 9 25 00 AM

### Testing
- Added unit test for `ImageSelect` because it was missing
- Added unit test for `DistrosTable`
- Added small Cypress test
- Some of the tests are failing because the `BaseTable` was not correctly setting the `data-cy-row` prop. I'll fix these after getting a review, since they introduce a lot of unrelated changes.

### DIFF
--- a/apps/spruce/src/pages/image/Header.tsx
+++ b/apps/spruce/src/pages/image/Header.tsx
@@ -1,0 +1,32 @@
+import styled from "@emotion/styled";
+import { H2, Overline } from "@leafygreen-ui/typography";
+import { ImageTabRoutes } from "constants/routes";
+import { size } from "constants/tokens";
+import { getTabTitle } from "./getTabTitle";
+
+interface HeaderProps {
+  imageId: string;
+  tab: ImageTabRoutes;
+}
+
+export const Header: React.FC<HeaderProps> = ({ imageId, tab }) => {
+  const { title } = getTabTitle(tab);
+
+  return (
+    <Container data-cy="image-tab-title">
+      <Overline>{title}</Overline>
+      <TitleContainer>
+        <H2>{imageId}</H2>
+      </TitleContainer>
+    </Container>
+  );
+};
+
+const Container = styled.div`
+  display: flex;
+  flex-direction: column;
+`;
+
+const TitleContainer = styled.div`
+  margin-right: ${size.s};
+`;
--- a/apps/spruce/src/pages/image/Tabs.tsx
+++ b/apps/spruce/src/pages/image/Tabs.tsx
@@ -35,5 +35,5 @@ export const ImageTabs: React.FC<ImageTabsProps> = ({
 
 const Container = styled.div`
   min-width: 600px;
-  width: 60%;
+  max-width: 75%;
 `;
--- a/apps/spruce/src/pages/image/Tabs.tsx
+++ b/apps/spruce/src/pages/image/Tabs.tsx
@@ -1,0 +1,39 @@
+import styled from "@emotion/styled";
+import { Routes, Route, Navigate } from "react-router-dom";
+import { ImageTabRoutes } from "constants/routes";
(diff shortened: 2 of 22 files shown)


apps/spruce/src/pages/image/index.tsx
```diff
@@ -18,7 +24,10 @@ const Image: React.FC = () => {
 
   const selectedImage = imageId ?? firstImage;
 
-  if (!Object.values(ImageTabRoutes).includes(currentTab as ImageTabRoutes)) {
+  if (
+    !Object.values(ImageTabRoutes).includes(currentTab as ImageTabRoutes) ||
+    currentTab === undefined
```
sophstad, line 29: Just curious how the initial case wouldn't catch `currentTab === undefined`

apps/spruce/src/pages/image/Tabs.tsx
```diff
@@ -0,0 +1,39 @@
...
+          element={<BuildInformationTab imageId={imageId} />}
+        />
+      </Routes>
+    </Container>
+  );
+};
+
+const Container = styled.div`
+  min-width: 600px;
+  width: 60%;
```
sophstad, line 38: What do you think of increasing the min width here? It's annoying that making the window smaller initially reduces the width of content instead of the width of whitespace

(would also be open to making this change for other settings pages but no need to address rn)

apps/spruce/src/pages/image/Header.tsx
```diff
@@ -0,0 +1,32 @@
...
+  imageId: string;
+  tab: ImageTabRoutes;
+}
+
+export const Header: React.FC<HeaderProps> = ({ imageId, tab }) => {
+  const { title } = getTabTitle(tab);
+
+  return (
+    <Container data-cy="image-tab-title">
+      <Overline>{title}</Overline>
```
sophstad, line 17: Using `Overline` the way it's actually meant to be used 🫡 

That being said, we usually don't use the "entity" being viewed as the heading on these pages. Is there a particular reason to deviate from that pattern here? i.e. I would expect the heading to say "Build Information" and the image in question is present in the sidebar selector

apps/spruce/src/pages/image/index.tsx
minnakt: Tbh it should catch it... But I get a TypeScript error if I remove it because of the way `useParams` works 😔 I can swap the order of the statements so it makes a little more "sense" haha

apps/spruce/src/pages/image/Header.tsx
minnakt: It comes from the feedback during the design phases, where users were having trouble orienting themselves in the app. They couldn't tell what image they were looking at (even with the name written in the dropdown).

From the feedback:
Screenshot 2024-08-13 at 1 43 37 PM

I think maybe we should rework the project one too tbh, there's been some history of people not noticing they're on the wrong project and making edits to it. Example in [DEVPROD-3137](https://jira.mongodb.org/browse/DEVPROD-3137)

apps/spruce/src/pages/image/Tabs.tsx
minnakt: not sure if I did quite what you asked but I made the table bigger!

apps/spruce/src/pages/image/Header.tsx
sophstad: Great context, thanks!

sophstad (approved):

PR #306, opened 2024-08-12
DEVPROD-4958: Allow submitting 0 to batch time input

SupaJoon:
DEVPROD-4958

### Description
These code changes allow submitting 0 to the batch time input. If the field is cleared and submitted then the input value will be casted to 0 in for GQL payload as well as well as text input value. Attached projects will cast 0 to null to show the placeholder value indicating that a default repo value is applied.


apps/spruce/src/pages/projectSettings/tabs/GeneralTab/transformers.ts
```diff
@@ -36,8 +36,7 @@ export const gqlToForm = ((data, options = {}) => {
           "identifier" in projectRef && {
             identifier: projectRef.identifier,
```
minnakt, line 37: I think you might want to make some adjustments to this PR because I can't see this placeholder anymore:

Screenshot 2024-08-14 at 5 35 59 PM

Now it says the batch time is 0, even though it should be 5 since it inherited from repo

minnakt (approved):

PR #308, opened 2024-08-13
DEVPROD-924: Make Spruce & Parsley test names group alphabetically 

SupaJoon:
DEVPROD-924

### Description
These code changes append the top-most test spec description to the beginning of Spruce and Parsley e2e test names. The result is that tests that belong to the same file begin with the same words. This make Spruce and Parsley tests group by spec file when sorting alphabetically. 
The original goal of this ticket was to append the spec file name to the beginning of the test name. This isn't feasible without an out-of-scope custom reporting solution because of this [issue](https://github.com/cypress-io/cypress/issues/8237) with mocha-junit-report.
The tests were originally structured like this:  `{it_statement}.{describe_statement}{_}{it_statement}` and now they are structured like this: `{describe_statement}{_}{it_statement}.{it_statement}`. This is possible with the [testCaseSwitchClassnameAndName](https://github.com/michaelleeallen/mocha-junit-reporter?tab=readme-ov-file#full-configuration-options) option. There is no Evergreen or mocha-junit-reporter option to clear the Classname entirely and remove the repeated copy.
I updated some test files to add a top level describe statement that covers all test assertions if one didn't exist and discovered this [bug (DEVPROD-9839)](https://jira.mongodb.org/browse/DEVPROD-9839) when testing the code changes by clicking around the tests table.


khelif96 (commented):
For consistency could we also apply the same name formatting for the parsley cypress tests

PR #309, opened 2024-08-13
DEVPROD-7444 Add task scheduling warnings

hadjri:
DEVPROD-7444

🔵Spruce

### Description
Followup from linked Evergreen PR. If > 1000 tasks are about to be scheduled by an operation, display a banner warning suggesting to confirm if the operation is necessary for testing, along with a link to our docs.

The new resolver from the linked Evergreen PR is needed to estimate the number of tasks that generator tasks are going to create, so that their load can be accounted for in the warning banner.
### Screenshots
Ignore the numbers. This will only actually show if N > 1000. Screenshots are from when I hardcoded the limit down to just 1.

Screenshot 2024-07-31 at 12 27 25 PM

Screenshot 2024-07-31 at 12 25 39 PM

Screenshot 2024-07-31 at 12 26 04 PM

### Testing
Tested configure patch, restart version, and schedule tasks locally on versions with generator tasks that had `estimated_num_generated_tasks` set on their documents and confirmed that the number displayed in the warning was accurate.
### Evergreen PR
https://github.com/evergreen-ci/evergreen/pull/8169


apps/spruce/src/components/ScheduleTasksModal/index.tsx
```diff
@@ -70,6 +74,12 @@ export const ScheduleTasksModal: React.FC<ScheduleTasksModalProps> = ({
     dispatch({ type: "ingestData", taskData });
   }, [taskData]);
 
+  const numEstimatedActivatedGeneratedTasks =
+    getNumEstimatedActivatedGeneratedTasks(
```
sophstad, line 78: [nit] We tend to use "count" instead of "num" for naming
```suggestion
  const estimatedActivatedGeneratedTasksCount =
    getEstimatedActivatedGeneratedTasksCount(
```

apps/spruce/src/components/VersionRestartModal/VersionRestartModal.tsx
```diff
@@ -167,6 +177,19 @@ const VersionRestartModal: React.FC<VersionRestartModalProps> = ({
...
+            <Banner data-cy="disabled-webhook-banner" variant="warning">
+              This is a large operation, expected to schedule{" "}
+              {selectedTotal + numEstimatedActivatedGeneratedTasks} tasks.
+              Please confirm that this number of tasks is necessary before
+              continuing. For more information, please refer to our{" "}
+              <StyledLink href={taskSchedulingLimitsDocumentationUrl}>
+                docs.
+              </StyledLink>
+            </Banner>
+          )}
```
sophstad, line 192: If it's not too much trouble I think moving this banner into its own component (like `components/TaskSchedulingWarning`) would be ideal since it's exactly the same across the three components!

apps/spruce/src/constants/task.ts
```diff
@@ -283,3 +283,5 @@ export const finishedTaskStatuses = [
 // Task name and build variant for the commit queue. Both are owned by Evergreen.
 export const mergeTaskName = "merge-patch";
 export const mergeTaskVariant = "commit-queue-merge";
+
+export const largeNumFinalizedTasksThreshold = 1000;
```
sophstad, line 287: If you do move the banner into one component I'd perhaps colocate this value since it seems unlikely to be used elsewhere

sophstad (changes requested):
Thanks for taking this on! I'm on PTO starting tomorrow so feel free to request the team for re-review.

apps/spruce/src/components/Banners/TaskSchedulingWarningBanner.tsx
```diff
@@ -0,0 +1,20 @@
+import Banner from "@leafygreen-ui/banner";
+import { taskSchedulingLimitsDocumentationUrl } from "../../constants/externalResources";
+import { StyledLink } from "../styles";
```
khelif96, line 3: Same comment Sophie had about these auto-resolving as abs imports
```suggestion
import { taskSchedulingLimitsDocumentationUrl } from "constants/externalResources";
import { StyledLink } from "styles";
```

apps/spruce/src/components/Banners/TaskSchedulingWarningBanner.tsx
```diff
@@ -0,0 +1,20 @@
...
+interface TaskSchedulingWarningBannerProps {
+  totalTasks: number;
+}
+const largeNumFinalizedTasksThreshold = 1000;
+
+export const TaskSchedulingWarningBanner: React.FC<
+  TaskSchedulingWarningBannerProps
+> = ({ totalTasks }) =>
+  totalTasks >= largeNumFinalizedTasksThreshold ? (
+    <Banner data-cy="disabled-webhook-banner" variant="warning">
```
khelif96, line 14: I think this `data-cy` attribute is wrong and probably unnecessary. We typically use these attributes for our e2e tests so that we can easily reference components but I don't see any e2e tests for this so you can omit it.

apps/spruce/src/pages/configurePatch/configurePatchCore/index.tsx
```diff
@@ -39,9 +39,11 @@ import {
   VariantTasksState,
   useConfigurePatch,
 } from "hooks/useConfigurePatch";
+import { TaskSchedulingWarningBanner } from "../../../components/Banners/TaskSchedulingWarningBanner";
```
khelif96, line 42: Same comment about abs imports

apps/spruce/src/components/VersionRestartModal/VersionRestartModal.tsx
```diff
@@ -209,6 +219,29 @@ const getTaskIds = (selectedTasks: versionSelectedTasks) =>
     }))
     .filter(({ taskIds }) => taskIds.length > 0);
 
+const getNumEstimatedActivatedGeneratedTasks = (
```
khelif96, line 222: I see this function declared 3 times in this pr with slightly different implementations but they all seem to be doing the same thing. Can you make it more generic and reuse it as a util method instead of rewriting it multiple times?

apps/spruce/src/pages/configurePatch/configurePatchCore/utils.ts
```diff
@@ -0,0 +1,33 @@
+import { VariantTask } from "../../../gql/generated/types";
+import { VariantTasksState } from "../../../hooks/useConfigurePatch";
```
khelif96, line 2: abs imports

khelif96 (changes requested):

apps/spruce/src/utils/tasks/estimatedActivatedTasks.ts
```diff
@@ -0,0 +1,106 @@
...
+        numEstimatedActivatedGeneratedTasks += 1;
+        if (generatedTaskCounts[`${variant}-${task}`]) {
+          numEstimatedActivatedGeneratedTasks +=
+            generatedTaskCounts[`${variant}-${task}`];
+        }
+      }
+    });
+  });
+  return numEstimatedActivatedGeneratedTasks;
+};
```
khelif96, line 70: Same comment as above this can be written as (with a jsdoc comment)
```ts
const sumSelectedEstimatedTasks = (
  selectedBuildVariantTasks: VariantTasksState,
  variantsTasks: VariantTask[],
  generatedTaskCounts: GeneratedTaskCountResults[],
): number => {
  // Create a set of existing tasks from variantsTasks
  const existingTasks = new Set<string>(
    variantsTasks.flatMap((variantTask) =>
      variantTask.tasks.map((task) => `${variantTask.name}-${task}`),
    ),
  );

  // Sum the estimated tasks for selected tasks that are not in existingTasks
  return Object.entries(selectedBuildVariantTasks).reduce(
    (total, [variant, tasks]) =>
      Object.entries(tasks).reduce((innerTotal, [taskName, isSelected]) => {
        if (isSelected && !existingTasks.has(`${variant}-${taskName}`)) {
          const generatedTaskCount =
            generatedTaskCounts.find(
              (count) =>
                count.buildVariantName === variant &&
                count.taskName === taskName,
            )?.estimatedTasks || 0;
          return innerTotal + 1 + generatedTaskCount;
        }
        return innerTotal;
      }, total),
    0,
  );
};
```

khelif96 (changes requested):

apps/spruce/src/utils/tasks/estimatedActivatedTasks.ts
```diff
@@ -0,0 +1,106 @@
...
+import {
+  selectedStrings,
+  versionSelectedTasks,
+} from "hooks/useVersionTaskStatusSelect";
+
+type ValidInputTypes = Set<string> | versionSelectedTasks | VariantTasksState;
+
+export const getNumEstimatedActivatedTasks = (
+  inputType: ValidInputTypes,
+  map: { [key: string]: number },
```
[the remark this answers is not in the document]
hadjri: The discrepancies in the multiple implementations of counting the number of estimated generated tasks stems from the fact that the `ConfigurePatchCore`, `VersionRestartModal`, and `ScheduleTasksModal` all storing the state of selected tasks in slightly different data structures. For example:

```
// VersionRestartModal. True indicates the task has been selected
{
  "example_version_id": {
    "task_id_1": true,
    "task_id_2": false,
  }
}

// ScheduleTasksModal. Existence in the set indicates the task has been selected
Set([
  "task_id_1",
  "task_id_2",
])

// ConfigurePatchCore. True indicates the task has been selected
{
  "example_bv": {
    "task1": true,
    "task2": false,
  }
}

Since the goal of this work is to include the number of generated tasks when estimating how many tasks will get scheduled in the UI, the new generatedTaskCounts resolver returns (currently a map) of tasks to their expected number of generated tasks they will create. For tasks in versions, this is relatively straightforward because we can just map the task ID to the task document's [EstimatedNumActivatedGeneratedTasks](https://github.com/evergreen-ci/evergreen/blob/8f3341c54f8430a9be6e5ae817c81a1a65f0e9b2/model/task/task.go#L302) field. Ex: 
```
{
  "task_id_1": 10,
  "task_id_2": 5,
}
```
There is slightly more complexity in the "Configure Tasks" component because we are potentially creating tasks that do not exist until we hit "Configure", so we cannot directly create a map with task IDs since there are no IDs yet. This is why the patch resolver for this field returns a map that looks like this:
```
{
  "taskName1-variantName": 10,
  "taskName2-variantName": 5
}
```
In either case, though, the data model is a map of String-> Int. It may be beneficial to use something stricter like a [StringMap](https://github.com/evergreen-ci/evergreen/pull/8169#discussion_r1714084091) but for integers instead.

File: apps/spruce/src/utils/tasks/estimatedActivatedTasks.test.ts [hunk 11]
```diff
@@ -0,0 +1,44 @@
...
+import { VariantTasksState } from "hooks/useConfigurePatch";
+import { versionSelectedTasks } from "hooks/useVersionTaskStatusSelect";
+import { getNumEstimatedActivatedTasks } from "./estimatedActivatedTasks";
+
+describe("getNumEstimatedActivatedTasks", () => {
+  const generatedTaskCounts: { [key: string]: number } = {
+    "variant1-task1": 5,
+    "variant1-task2": 10,
+  };
+  it("should compute correctly when using VariantTasksState", () => {
```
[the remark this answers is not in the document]
hadjri: Sure, I added more test cases and tried to make the descriptions better.

apps/spruce/src/utils/tasks/estimatedActivatedTasks.ts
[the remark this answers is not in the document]
khelif96: Sorry for complicating this and for the slow response. My primary concern is improving readability and ensuring we maintain type safety across the codebase. My suggestion at Introducing the `GeneratedTaskCountResults` type was aimed at reducing the need for `any` types and providing a more structured and consistent way to handle the generated task counts. My goal with introducing `GeneratedTaskCountResults` as a stricter type was to limit the ambiguity and potential errors that arise from using more generic structures like `Map` or `StringMap`.
This would help simplify the logic, making it easier to follow and reducing the cognitive load when working with these counts. It would also make the code more self-documenting, as the fields we reference would be explicitly defined rather than inferred from the map's keys. 
I do understand why this function needs to be implemented this way due to all of the different underlying implementations but I think we could still strike a balance between having a flexible implementation and a well documented type-safe one.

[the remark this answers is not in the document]
hadjri: Apologies for late reply as well due to outage work. I understand the concern and I've opened https://github.com/evergreen-ci/evergreen/pull/8268 on the Evergreen side and will update this PR accordingly after it is merged.

hadjri:
Attempted to document better, use the new strong type, and make it generally more readable

apps/spruce/src/utils/tasks/estimatedActivatedTasks.ts
```diff
@@ -0,0 +1,139 @@
...
+      generatedTaskCounts,
+      existingVariantsTasks,
+    );
+  }
+  // Compute for "Configure Patch"
+  if (isVersionSelectedTasks(inputType)) {
+    return sumActivatedTasksInSelectedTasks(inputType, generatedTaskCounts);
+  }
+  throw new Error("Unrecognized input type for counting generated tasks");
+};
```
khelif96, line 39: Sorry, I take this back the logic for diffing the different types seems error-prone. I think we should use the underlying methods directly where they are needed.

khelif96 (approved):
Thank you for making these changes! I had a few comments but you don't need a reapproval after addressing them.

PR #311, opened 2024-08-14
DEVPROD-928: Parallelize Spruce e2e tasks

sophstad:
DEVPROD-928

### Description

- Parallelize Spruce e2e tests. This brings the wait time down from ~25 to ~10-12 minutes 😱 
  - Programmatically bucket specs into N groups; I went with 4 groups for now but this is easy to play with. The buckets should have the same-ish size so that we don't have hugely different runtimes. To do this, the spec directories are sorted by size and then distributed into buckets. We could implement a smarter algorithm if this became a pain point.
  - One consequence of programmatic bucketing is that the base status for each execution task is not necessarily meaningful. However, I don't anticipate that being a huge problem since the parent e2e_parallel base status is still correct and it's quite easy to see which test is failing. I think the reduced maintenance of not having to manually bucket tests is a good tradeoff.
- Don't parallelize Parsley: the comparatively short runtime combined with the significant setup time for e2e tests meant that there was basically no time difference (see a run [here](https://spruce.mongodb.com/task/evergreen_ui_parsley_display_e2e_parallel_patch_2f696a40d7b1fea4a86fc8dee7ecc2b94c995cbf_66bb6d59896b850007fd183d_24_08_13_14_27_39?execution=0&sortBy=STATUS&sortDir=ASC)). We can always revisit this if Parsley gets significantly longer, or try it with a different parallelization number.
- Opened a separate ticket for parallelizing Evergreen's downstream Spruce task (DEVPROD-9875). This will probably require making a new generator so I opted to split up the work.

### DIFF
--- a/.evergreen/generate.yml
+++ b/.evergreen/generate.yml
@@ -23,8 +23,8 @@ tasks:
   - name: generate-tasks
     commands:
       - func: run-generate-tasks
+      - func: attach-generated
       - command: generate.tasks
         params:
           files:
             - ui/.evergreen/generate-tasks.json
-      - func: attach-generated
--- a/.evergreen/scripts/generate-tasks.js
+++ b/.evergreen/scripts/generate-tasks.js
@@ -116,25 +116,29 @@ const targetsFromChangedFiles = (files) => {
   return Array.from(targets);
 };
 
-const generateE2E = (bv) => {
+/**
+ * generateParallelE2ETasks creates the top-level tasks required for a build variant's parallelized Cypress testing.
+ * Note that the task steps included here should likely be in sync with those included in the base e2e task.
+ * @param bv - build variant name
+ * @returns - array of tasks
+ */
+const generateParallelE2ETasks = (bv) => {
   const specs = getSpecs(`./apps/${bv}/cypress/integration`);
-  return specs.map((spec, i) => {
-    return {
-      name: `e2e_${bv}_${i}`,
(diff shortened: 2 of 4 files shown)


SupaJoon (commented):
I'm unable to run the parallelized suite from a patch ([example](https://spruce.mongodb.com/version/66bfc3807ae0d40008282a21/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)). Is that intentional? Also, I imagine since the bucketizing algorithm is deterministic, we should see the base status column match up for versions that don't introduce an e2e test diff.

minnakt (approved):

sophstad:
> I'm unable to run the parallelized suite from a patch ([example](https://spruce.mongodb.com/version/66bfc3807ae0d40008282a21/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)). Is that intentional? Also, I imagine since the bucketizing algorithm is deterministic, we should see the base status column match up for versions that don't introduce an e2e test diff.

@SupaJoon since `e2e_parallel` is a generated task, we need to split out the task into its own generator to make it runnable via patch; DEVPROD-9875 captures that! The bucketizing algorithm is deterministic within a run, but changes to files between commits could introduce changes that make the execution task statuses inconsistent, so I suppose that's more of an edge case.

Sorry for the auto-merge closing this before I could respond!

PR #313, opened 2024-08-15
DEVPROD-9904 Switch GitHub app private key from input to textarea

ZackarySantana:
DEVPROD-9904

### Description
GitHub app private keys need the new line characters to be valid. Instead of having users manually add new line characters, and then escaping them ourselves, it's better to change it to a textarea so the new line characters are preserved.

### Screenshots

Screenshot 2024-08-15 at 12 27 15 PM

### Testing
Ran locally and then saved it with the textarea, the keys are successfully saved with the new lines. With the input, they were not.

### DIFF
--- a/apps/spruce/cypress/integration/projectSettings/github_app_settings.ts
+++ b/apps/spruce/cypress/integration/projectSettings/github_app_settings.ts
@@ -58,7 +58,7 @@ describe("GitHub app settings", () => {
     cy.get("@appId").should("have.value", "12345");
     cy.get("@privateKey").should("have.value", "{REDACTED}");
     cy.get("@appId").should("have.attr", "aria-disabled", "true");
-    cy.get("@privateKey").should("have.attr", "disabled", "");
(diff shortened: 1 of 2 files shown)


apps/spruce/cypress/integration/projectSettings/github_app_settings.ts
```diff
@@ -58,7 +58,7 @@ describe("GitHub app settings", () => {
     cy.get("@appId").should("have.value", "12345");
     cy.get("@privateKey").should("have.value", "{REDACTED}");
     cy.get("@appId").should("have.attr", "aria-disabled", "true");
-    cy.get("@privateKey").should("have.attr", "aria-disabled", "true");
+    cy.get("@privateKey").should("have.attr", "disabled", "");
```
minnakt, line 61: I think maybe the test is failing because you want to call it like this:
```suggestion
    cy.get("@privateKey").should("have.attr", "disabled");
```

minnakt (approved):

PR #316, opened 2024-08-16
DEVPROD-8369 Remove Newrelic and setup Honeycomb SDK to start sending traces to Honeycomb

khelif96:
DEVPROD-8369

### Description
* Removes usage of NewRelic in most places
* Sets up honeycomb SDK
* Updates attribute names to adhere to OTEL naming conventions (This is split off into #315 so once that is merged this diff should not include those changes)
* Update deploy process to allow deploying to prod and staging with correct credentials (I created an ingest only key for both Spruce and Parsley as well as a collector endpoint)
* Created util methods for sending analytics events to honeycomb as traces
* Created methods to extract graphql query names from network request traces and propagate them.
* Create an `OpenTelemetryAttributeStore` name WIP store that injects a value into the browser window and ensures that we can globally share span attributes across the entire app even those outside of the react context
* Use collector endpoint to send traces.
* Enable B3 propagation to connect frontend to backend traces https://github.com/evergreen-ci/evergreen/pull/8285 

### Testing
Parsley dashboard
https://ui.honeycomb.io/mongodb-4b/environments/staging/datasets/parsley/home
Spruce dashboard
https://ui.honeycomb.io/mongodb-4b/environments/staging/datasets/spruce/home


.evergreen/deploy.yml
```diff
@@ -16,6 +16,8 @@ functions:
           SPRUCE_NEW_RELIC_AGENT_ID: ${SPRUCE_NEW_RELIC_AGENT_ID}
           SPRUCE_NEW_RELIC_APPLICATION_ID: ${SPRUCE_NEW_RELIC_APPLICATION_ID}
           SPRUCE_SENTRY_AUTH_TOKEN: ${SPRUCE_SENTRY_AUTH_TOKEN}
+          HONEYCOMB_INGEST_KEY_STAGING: ${HONEYCOMB_INGEST_KEY_STAGING}
+          HONEYCOMB_INGEST_KEY_PROD: ${HONEYCOMB_INGEST_KEY_PROD}
```
minnakt, line 20: Are these temporary? I feel like we should be sending to the collector

apps/spruce/src/index.tsx
```diff
@@ -1,9 +1,17 @@
...
+import { initializeHoneycomb } from "@evg-ui/lib/utils/observability";
 import { initializeErrorHandling } from "components/ErrorHandling";
+import { isDevelopmentBuild } from "utils/environmentVariables";
 import App from "./App";
 
 initializeErrorHandling();
+initializeHoneycomb({
+  serviceName: "spruce",
+  debug: isDevelopmentBuild(),
+  endpoint: process.env.REACT_APP_HONEYCOMB_ENDPOINT || "",
```
minnakt, line 12: is there a reason that in Parsley this variable is equal to `process.env.HONEYCOMB_ENDPOINT` but here it's equal to `process.env.REACT_APP_HONEYCOMB_ENDPOINT`?

Also should it be `REACT_APP_HONEYCOMB_BASE_URL`?

.evergreen/deploy.yml
```diff
@@ -16,6 +16,8 @@ functions:
           SPRUCE_NEW_RELIC_AGENT_ID: ${SPRUCE_NEW_RELIC_AGENT_ID}
           SPRUCE_NEW_RELIC_APPLICATION_ID: ${SPRUCE_NEW_RELIC_APPLICATION_ID}
```
minnakt, line 17: should the New Relic keys be removed from here?

packages/deploy-utils/src/setup-credentials/index.ts
```diff
@@ -42,6 +42,9 @@ const base: Record<TargetEnvironment, Record<string, string | undefined>> = {
     REACT_APP_GQL_URL:
```
minnakt, line 42: should the NewRrelic keys be removed from this file?

packages/lib/src/analytics/utils.ts
```diff
@@ -1,6 +1,29 @@
...
+ * @param properties - The properties to send with the event
+ */
+export const sendEventTrace = <A extends ActionType>(
+  { name, ...actionProps }: A,
+  properties: AnalyticsProperties,
+) => {
+  const tracer = trace.getTracer("analytics");
+
+  tracer.startActiveSpan(name, (span) => {
+    // use span.setAttribute to set any relevant attributes
```
minnakt, line 18: you can remove this comment haha I think I included it for the implementer

packages/lib/src/analytics/utils.ts
```diff
@@ -1,6 +1,29 @@
...
+      ...properties,
+      ...actionProps,
+    });
+    span.end();
+  });
+};
+
 /**
  * `addNewRelicPageAction` is a function that sends an event to our analytics provider
  * @param action - The action to send to our analytics provider
```
minnakt, line 29: Is it possible to remove this function now?

packages/lib/src/utils/observability/honeycomb.ts
```diff
@@ -0,0 +1,66 @@
...
+      resourceAttributes: {
+        "user.id": userId,
+      },
+      localVisualizations: debug,
+      serviceName,
+      apiKey,
+    });
+    honeycombSdk.start();
+  } catch (e) {
+    console.error(`Could not start honeycomb SDK ${e}`);
```
minnakt, line 62: nit. 
```suggestion
    console.error(`Could not start Honeycomb SDK: ${e}`);
```

minnakt (changes requested):
(i ignored the files with attribute changes)

apps/parsley/src/main.tsx
```diff
@@ -1,9 +1,21 @@
...
+  injectOpenTelemetryAttributeStoreIntoWindow,
+} from "@evg-ui/lib/utils/observability";
 import { initializeErrorHandling } from "components/ErrorHandling";
+import { isDevelopmentBuild } from "utils/environmentVariables";
 import App from "./App";
 
 initializeErrorHandling();
+initializeHoneycomb({
+  debug: isDevelopmentBuild(),
+  endpoint: process.env.REACT_APP_HONEYCOMB_ENDPOINT || "",
```
minnakt, line 14: where does this variable `REACT_APP_HONEYCOMB_ENDPOINT` come from? I don't see any references to it elsewhere

apps/parsley/src/analytics/useAnalyticAttributes.ts
```diff
@@ -4,25 +4,25 @@ import { useLogContext } from "context/LogContext";
...
     }
-
-    const { newrelic } = window;
     if (logType !== undefined) {
-      newrelic.setCustomAttribute("logType", logType);
+      AttributeStore.setGlobalAttribute("log.type", logType);
     }
     if (userId !== null) {
-      newrelic.setCustomAttribute("userId", userId);
+      AttributeStore.setGlobalAttribute("user.id", userId);
```
minnakt, line 20: how would this interact with the `"user.id"` being set in the `initializeHoneycomb` ? is one of them superfluous?

packages/lib/src/utils/observability/AttributeStore/utils.ts
```diff
@@ -0,0 +1,16 @@
+import { OpenTelemetryAttributeStore } from ".";
+
+/**
+ * Injects the OpenTelemetry Attribute store into the window object.
+ *
```
minnakt, line 5: nit. extra line
```suggestion
```

packages/lib/src/utils/observability/honeycomb.ts
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Configuration object for the Honeycomb SDK.
+ */
+interface HoneycombConfig {
+  /** The service name for the app we are running e.g. spruce or parsley */
+  serviceName: string;
+  /** The endpoint for the Honeycomb SDK to send traces to if we are not using the default */
+  endpoint: string;
+  /** The url for our honeycomb instrumented server to connect frontend and backend traces together */
```
minnakt, line 13: nit. 
```suggestion
  /** The url for our Honeycomb instrumented server to connect frontend and backend traces together */
```

packages/lib/src/utils/observability/honeycomb.ts
```diff
@@ -0,0 +1,97 @@
...
+  backendURL?: string;
+  /** Whether to enable debug mode in the honeycomb sdk this will enable additional logging and print links to traces */
```
minnakt, line 15: nit. 
```suggestion
  /** Whether to enable debug mode in the Honeycomb SDK this will enable additional logging and print links to traces */
```

packages/lib/src/utils/observability/honeycomb.ts
```diff
@@ -0,0 +1,97 @@
...
+                    span.setAttribute(
+                      "graphql.query_type",
+                      graphqlQuery.queryType,
+                    );
+                  }
+                }
+              },
+
+              // Allow connecting frontend & backend traces.
+              propagateTraceHeaderCorsUrls: [backendURL || ""],
```
minnakt, line 75: should this be wrapped in `new RegExp`? otherwise I'm not sure that it will work

packages/lib/src/utils/observability/honeycomb.ts
```diff
@@ -0,0 +1,97 @@
...
+                      graphqlQuery.operationName,
+                    );
+                    span.setAttribute(
+                      "graphql.query_type",
+                      graphqlQuery.queryType,
+                    );
+                  }
+                }
+              },
+
```
minnakt, line 73: nit. extra line
```suggestion
```

packages/lib/src/utils/observability/honeycomb.ts
```diff
@@ -0,0 +1,97 @@
...
+            },
+          }),
+        ],
+        // Add user.id as an attribute to all traces.
+        resourceAttributes: {
+          "user.id": userId,
+        },
+        localVisualizations: debug,
+        serviceName,
+        apiKey: ingestKey,
```
minnakt, line 88: could we add a comment about wanting to move away from the ingest key at some point, maybe with a ticket number?

khelif96: I don't think we do want to move away from the ingestKey its the recommended way to authenticate requests with honeycomb. https://www.honeycomb.io/blog/ingest-only-api-keys Is there something else we should be using?

apps/parsley/src/main.tsx
khelif96: This is intended to connect Honeycomb to refinery. Right now there are several cors issues with communicating with Refinery that we are still trying to work out. But once they are resolved this should be plug-in-play

apps/parsley/src/analytics/useAnalyticAttributes.ts
khelif96: Yes it would just overwrite it. This case handles the attribute not being set when the app initializes.

packages/lib/src/utils/observability/honeycomb.ts
minnakt: Honeycomb states that ingest key [is not as secure](https://docs.honeycomb.io/send-data/javascript-browser/honeycomb-distribution/#configuration-options-for-browser-telemetry) as sending to a collector, which doesn't require an ingest key. That's what I assumed your conversations with Jonathan were about, having the collector handle the auth.

khelif96: I talked to Jonathan and we managed to get the collector working if a user is on the vpn

packages/lib/src/utils/string/string.test.ts
```diff
@@ -1,8 +1,21 @@
...
+  it("escapes special characters in a string", () => {
+    // eslint-disable-next-line no-useless-escape -- This is a regex escape character test
+    expect(toEscapedRegex("hello-world")).toBe("hello\\-world");
+    // eslint-disable-next-line no-useless-escape -- This is a regex escape character test
+    expect(toEscapedRegex("hello/world")).toBe("hello\\/world");
+    expect(toEscapedRegex("https://example.com")).toBe(
+      // eslint-disable-next-line no-useless-escape -- This is a regex escape character test
+      `https:\\/\\/example\\\.com`,
+    );
+  });
```
minnakt, line 20: I feel like this will pass the tests and the linting:
```suggestion
  it("escapes special characters in a string", () => {
    expect(toEscapedRegex("hello-world")).toBe("hello\\-world");
    expect(toEscapedRegex("hello/world")).toBe("hello\\/world");
    expect(toEscapedRegex("https://example.com")).toBe(
      "https:\\/\\/example\\.com",
    );
  });
```

packages/lib/src/utils/string/index.ts
```diff
@@ -12,3 +12,12 @@ export const toSentenceCase = (string: string) => {
...
 };
+
+/**
+ * `toEscapedRegex` takes a string and returns an escaped version of it that can be used in a regex.
+ * @param str - The string to escape.
+ * @returns - The escaped string.
+ */
+export const toEscapedRegex = (str: string): string =>
+  // eslint-disable-next-line no-useless-escape -- This is a function to escape regex characters
+  str.replace(/[-\/\\^$*+?.()|[\]{}]/g, "\\$&");
```
minnakt, line 23: It seems like maybe this lint error is caused by malformatted regex? I think this will pass
```suggestion
export const toEscapedRegex = (str: string): string =>
    str.replace(/[-\\/\\^$*+?.()|[\]{}]/g, "\\$&");
```

packages/lib/src/utils/observability/honeycomb.ts
```diff
@@ -0,0 +1,96 @@
...
+  if (!ingestKey || !endpoint) {
+    console.error(
+      "Honeycomb INGEST API key or a collector endpoint is required to start the SDK in production mode.",
+    );
+  } else {
+    try {
+      const userId = localStorage.getItem("userId") ?? undefined;
+      const honeycombSdk = new HoneycombWebSDK({
+        debug,
+        endpoint,
```
minnakt, line 52: how exactly does this work if both the `endpoint` and `apiKey` are defined? Don't we by default want to send to the `endpoint` and ignore the `apiKey`? If a collector `endpoint` is provided, it seems like we shouldn't need to provide an API key

Screenshot 2024-09-10 at 12 32 03 PM

packages/lib/src/utils/observability/honeycomb.ts
minnakt: that sounds fine given that we have WARP which should automatically turn on after a period of time even if disabled by the user

packages/lib/src/utils/observability/honeycomb.ts
khelif96: It will always send to the collector if the collector endpoint is supplied. The apiKey will just be ignored.

packages/lib/package.json
```diff
@@ -14,9 +14,13 @@
     "test": "vitest --typecheck=false"
   },
   "dependencies": {
-    "@emotion/react": "11.13.3",
-    "@emotion/styled": "11.13.0",
+    "@emotion/react": "11.11.0",
+    "@emotion/styled": "11.11.0",
```
minnakt, line 18: was this downgrade intentional?

packages/lib/src/utils/string/string.test.ts
```diff
@@ -1,8 +1,18 @@
...
     expect(toSentenceCase("myString")).toBe("Mystring");
   });
 });
+
+describe("toEscapedRegex", () => {
+  it("escapes special characters in a string", () => {
+    expect(toEscapedRegex("hello-world")).toBe("hello\\-world");
+    expect(toEscapedRegex("hello/world")).toBe("hello\\/world");
+    expect(toEscapedRegex("https://example.com")).toBe(
+      `https:\\/\\/example\\.com`,
```
minnakt, line 15: nit. use regular quotes

packages/lib/src/utils/observability/honeycomb.ts
minnakt: in that case yippee!

minnakt (approved):

packages/lib/package.json
khelif96: Nope! Good catch!

PR #317, opened 2024-08-16
DEVPROD-7655: Create General table on Build Information page

sophiabarness:
DEVPROD-7655

### Description
Create the General table on the Build Information page by querying relevant fields from the GraphQL API. 

### Screenshots
Screenshot 2024-08-19 at 11 42 16 AM

### Testing

Added Jest test.


apps/spruce/src/pages/image/GeneralTable/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+      {
+        property: "Amazon Machine Image (AMI)",
+        value: image.ami ?? "N/A",
+      },
+      {
+        property: "Latest task",
+        value: image.latestTask?.id ? (
+          <StyledRouterLink to={getTaskRoute(image.latestTask?.id)}>
+            <WordBreak>{image.latestTask?.id}</WordBreak>
+          </StyledRouterLink>
```
minnakt, line 58: I don't seem to be able to click this link (this would be a good thing to add to the Cypress tests)

minnakt (changes requested):

apps/spruce/src/pages/image/GeneralTable/index.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+  const columns: LGColumnDef<PropertyValue>[] = [
+    {
+      header: "Property",
+      accessorKey: "property",
+    },
+    {
+      header: "Value",
+      accessorKey: "value",
+      cell: ({ getValue }) => getValue(),
```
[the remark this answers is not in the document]
sophiabarness: Without this line, the link does not render properly.

apps/spruce/src/pages/image/GeneralTable/index.tsx
sophiabarness: I can click the link, but I don't think it's in the database, so it doesn't render the task page.

minnakt (changes requested):
I'm not really sure why, but on this PR, whenever I try to select a different distro from the dropdown the whole page freezes and then crashes. It doesn't seem to happen on the main branch, so it makes me think the problem is related to this PR. Could you investigate what's going on?

apps/spruce/src/pages/image/GeneralTable/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+        ),
+      },
+      {
+        property: "Latest task time",
+        value: image?.latestTask?.finishTime
+          ? getDateCopy(image.latestTask.finishTime)
+          : "N/A",
+      },
+    ],
+    [image],
```
minnakt, line 67: The problem is that `useDateFormat` returns a new reference to the function which triggers the `useMemo` to run infinitely.

The dependencies indicate when to rerun the code within `useMemo`. You can probably just leave the function out of the dependencies and use `// eslint-disable-next-line react-hooks/exhaustive-deps` (you can search the codebase for examples of this)

apps/spruce/src/pages/image/GeneralTable/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+  );
+
+  const tableContainerRef = useRef<HTMLDivElement>(null);
+  const table = useLeafyGreenTable<GeneralInfo>({
+    columns: useMemo(() => columns, []),
+    data: useMemo(() => tableData, [tableData]),
```
minnakt, line 73: memoization seems unnecessary here
```suggestion
    columns,
    data: tableData,
```

apps/spruce/src/pages/image/GeneralTable/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+    onError(err) {
+      dispatchToast.error(
+        `There was an error loading the image general table: ${err.message}`,
+      );
+    },
+  });
+
+  const image = useMemo(() => imageData?.image, [imageData?.image]);
+
+  const tableData = useMemo(
```
minnakt, line 40: nit. Could we just call this `data`?
```suggestion
  const data = useMemo(
```

apps/spruce/src/pages/image/GeneralTable/GeneralTable.test.tsx
```diff
@@ -0,0 +1,88 @@
+import { MockedProvider } from "@apollo/client/testing";
```
minnakt, line 1: Can we also add Cypress tests?

minnakt (changes requested):

PR #318, opened 2024-08-16
DEVPROD-4946: Make Host Event table task link responsive

SupaJoon:
DEVPROD-4946

### Description
Update `ShortenedRouterLink` to base width from viewport width when given a width breakpoint. 

### Screenshots

https://github.com/user-attachments/assets/74c5a0fa-4827-4f6a-aff8-1c300566871e


apps/spruce/src/pages/host/TaskLink.tsx
```diff
@@ -0,0 +1,36 @@
...
+  return (
+    <ShortenedRouterLink
+      data-cy={dataCy}
+      title={taskId}
+      to={getTaskRoute(taskId)}
+      width={linkWidth}
+    >
+      {taskId + taskId}
+    </ShortenedRouterLink>
+  );
```
khelif96, line 35: This  can be done entirely in CSS by updating the styling of ShortenedRouterLink
```css
    @media (max-width: 1200px) {
      max-width: 200px;
    }
    max-width: calc(100%-1000px);
```

khelif96 (changes requested):

apps/spruce/src/pages/host/TaskLink.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const linkWidth = screenWidth > 1200 ? screenWidth - 1000 : 200;
+
+  return (
+    <ShortenedRouterLink
+      data-cy={dataCy}
+      title={taskId}
+      to={getTaskRoute(taskId)}
+      width={linkWidth}
+    >
+      {taskId + taskId}
```
khelif96, line 33: Also not sure if this was intentional.

SupaJoon: No it's not, this is from an older commit

apps/spruce/src/components/styles/Link.tsx
```diff
@@ -19,14 +19,31 @@ export const StyledLink = (props) => (
...
     vertical-align: bottom;
-    max-width: ${({ width }) => `${width ?? 200}px`};
+    ${({ baseWidth, responsiveBreakpoint }) =>
+      responsiveBreakpoint
+        ? `@media (max-width: ${responsiveBreakpoint}px) { max-width: ${baseWidth}px; }`
+        : null};
+    max-width: ${({ baseWidth, responsiveBreakpoint }) =>
+      responsiveBreakpoint
+        ? `calc(100vw - ${responsiveBreakpoint - (baseWidth ?? 200)}px)`
+        : `${baseWidth ?? 200}px`};
```
khelif96, line 46: Sorry for the slow review, but I was wondering why would this approach be advantageous as opposed to just giving it a percentage. `max-width: 50%` or something like that would accomplish the same effect of allowing the link to scale in width in relation to the screen size while avoiding this breakpoint based logic.

SupaJoon: The current solution tries to set the max-width as close as it can the length of the red line. This is fr efficient use space and preventing the truncated text from taking up an entire line of it's own. The black line is the breakpoint value. The value is configurable because of differing page layouts.
Screenshot 2024-08-22 at 10 57 38 AM

Here is a video of the current solution: 

https://github.com/user-attachments/assets/09ee5792-d967-42db-8904-28d44a0ff1d1

Here is a video of setting max-width to 50vw:

https://github.com/user-attachments/assets/9011daeb-40fc-4b3d-9408-5c5a7c039b78

Here is a video of setting max-width to 75vh:

https://github.com/user-attachments/assets/9692b9b4-4ab0-4a9f-a117-29655aa8cc25

I think the breakpoint solution looks best and allows for the most configuration at the cost of some code complexity.

khelif96 (approved):

PR #324, opened 2024-08-20
DEVPROD-7657: Create toolchains table on Build Information page

minnakt:
DEVPROD-7657

### Description
Adds the Toolchains table.

### Screenshots

Screenshot 2024-08-20 at 10 23 47 AM

### Testing
- Cypress and Jest tests


apps/spruce/src/pages/image/ToolchainsTable/ToolchainsTable.test.tsx
```diff
@@ -0,0 +1,276 @@
...
+
+    const rows = screen.getAllByDataCy("toolchains-table-row");
+    expectedVersions.forEach((expectedVersion, i) => {
+      expect(within(rows[i]).getAllByRole("cell")[2]).toHaveTextContent(
+        expectedVersion,
+      );
+    });
+  });
+
+  it("supports name filter", async () => {
```
SupaJoon, line 91: Just a question and not requesting a change: why did you make this test in unit tests instead of cypress?

apps/spruce/src/pages/image/ToolchainsTable/index.tsx
```diff
@@ -0,0 +1,118 @@
...
+      },
+    },
+    onError(err) {
+      dispatchToast.error(
+        `There was an error loading image toolchains: ${err.message}`,
+      );
+    },
+  });
+
+  const toolchains = useMemo(
```
SupaJoon, line 58: Not requesting a change: Hopefully this is isn't necessary and LG wouldn't double render on an empty data prop

SupaJoon (approved):

apps/spruce/src/pages/image/ToolchainsTable/ToolchainsTable.test.tsx
minnakt: there is a test in Cypress, but since we can't predict how the image will change (e.g. the toolchain we try to search for in the Cypress tests could get completely deleted one day) we have to keep the Cypress test really general and search for a fake item. 

The unit test just makes sure that the parameters get passed properly to the GraphQL query

PR #325, opened 2024-08-20
DEVPROD-9626: Allow rendering command sections outside of function sections

SupaJoon:
DEVPROD-9626

### Description
These code changes allow rendering a command outside of a function. During section parsing, function entries will be generated for top-level commands and they will be tagged with `containsTopLevelCommand`. This way, log searching and section toggling can make calculations based on the property that all functions have at least 1 command and during render time, function headers that contain a top-level command will be skipped.
Reach out to me for a log URL to test in beta.


apps/parsley/src/utils/filterLogs/index.ts
```diff
@@ -80,18 +80,27 @@ const filterLogs = (options: FilterLogsParams): ProcessedLogLines => {
       for (let idx = 0; idx < logLines.length; idx++) {
         const func = sectionData.functions[funcIndex];
         const isFuncStart = func && idx === func.range.start;
-        const isFuncOpen = sectionState[func?.functionID]?.isOpen ?? false;
+        const isFuncOpen =
+          (sectionState[func?.functionID]?.isOpen ||
+            func?.containsTopLevelCommand) ??
+          false;
```
minnakt, line 86: if there a reason we can't just automatically set the section to be open during processing if the function contains top level command?

apps/parsley/src/components/LogRow/SubsectionHeader/index.tsx
```diff
@@ -1,31 +1,44 @@
...
 const { gray } = palette;
 
-interface SectionHeaderProps extends Row {
+interface SubsectionHeaderProps extends Row {
   commandName: string;
   functionID: string;
   commandID: string;
   open: boolean;
   step: string;
+  status: SectionStatus | undefined;
```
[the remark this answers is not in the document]
SupaJoon: The main difference between `?` and `undefined` is that `?` allows omitting the the prop completely whereas `undefined` requires the prop to be set. In implementation, the component is always assigned a value for `status` and switches the value between undefined and defined so I prefer to type it that way.

apps/parsley/src/utils/filterLogs/index.ts
SupaJoon: I rather handle this during processedLogLine construction (the render prop) so I can treat functions that contain top level commands the same as any other function when it comes to toggling all sections or any other open/close state logic we might add in the future.

apps/parsley/src/types/logs.ts
```diff
@@ -22,7 +22,7 @@ interface SkippedLinesRow {
 
 interface SectionHeaderRow {
   functionID: string;
-  functionName: string;
+  functionName: string | undefined;
```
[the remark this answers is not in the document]
SupaJoon: This should technically always be defined since functions containing top level commands are not pushed to processedLogLines. The ts compiler can't infer that from [here](https://github.com/evergreen-ci/ui/pull/325/files#diff-86d2f3595501c7254caabc17d058ba3afcbf84faa7d6b9835120ef4e59fb0d36R95) and [here](https://github.com/evergreen-ci/ui/pull/325/files#diff-572ddb65042aa207cbd06f85b85d004a6a52b820af6421cd1f4c07fbed4e42f7R94).

apps/parsley/src/utils/filterLogs/index.ts
```diff
@@ -80,18 +80,27 @@ const filterLogs = (options: FilterLogsParams): ProcessedLogLines => {
...
+        // A function start is detected.
         if (isFuncStart) {
-          // A function start is detected.
-          filteredLines.push({
-            ...func,
-            isOpen: isFuncOpen,
-            rowType: RowType.SectionHeader,
-          });
+          // Don't show the function header if contains a top-level command.
+          if (!func.containsTopLevelCommand) {
```
minnakt, line 94: tbh the more I think about it, the more I'm confused about how we have an invisible function that doesn't get rendered. is there not a way to just render a standalone command independent of a function?

apps/parsley/src/components/styles/sectionHeader.ts
```diff
@@ -0,0 +1,23 @@
+import { css } from "@leafygreen-ui/emotion";
+import { palette } from "@leafygreen-ui/palette";
+import { size } from "constants/tokens";
+
+const { gray } = palette;
+
```
minnakt, line 6: it looks like the sections are not rendering properly:

Screenshot 2024-08-23 at 10 19 56 AM

apps/parsley/src/components/LogRow/SubsectionHeader/SubsectionHeader.test.tsx
```diff
@@ -70,32 +75,49 @@ describe("SubsectionHeader", () => {
...
+        status={SectionStatus.Pass}
+      />,
+      { wrapper },
+    );
+    expect(screen.getByLabelText("Checkmark With Circle Icon")).toBeVisible();
+    rerender(<SubsectionHeader {...subsectionHeaderProps} />);
+    expectError(
+      () => screen.getByLabelText("Checkmark With Circle Icon"),
+      "Unable to find a label with the text of: Checkmark With Circle Icon",
+    );
```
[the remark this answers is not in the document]
SupaJoon: The screen query throws an error in this assertion and will automatically fail:
`expect(screen.getByLabelText("Checkmark with Circle Icon")).not.toBeInTheDocument();`
I tried this also and and the same error blocks the assertion from completing:
`expect(screen.getByLabelText("Checkmark with Circle Icon")).toThrow()`
That's how I decided to check the error itself with a try/catch.

apps/parsley/src/components/LogRow/RowRenderer/index.tsx
```diff
@@ -90,13 +90,22 @@ const ParsleyRow: RowRendererFunction = ({ processedLogLines }) => {
     }
 
     if (isSubsectionHeaderRow(processedLogLine)) {
+      let status;
+      // Only show status icon for top-level commands
+      if (processedLogLine.isTopLevelCommand) {
+        status = includesLineNumber(processedLogLine, failingLine)
+          ? SectionStatus.Fail
+          : SectionStatus.Pass;
+      }
```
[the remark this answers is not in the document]
SupaJoon: We can do this earlier on achieve the refactor you're mentioning but I don't think it will necessarily be cleaner at this point in time. Currently, the status calculation only happens in this file and is only important for render purposes. 
To get the value to propagate through, `useSections` would have to accept a separate `failingLineNumber` prop, and `FunctionEntry`, `CommandEntry`, `SubsectionHeader` and `SectionHeader` would all need to push along the value.

apps/parsley/src/utils/filterLogs/index.ts
SupaJoon: There are alternative ways to do this but I found them more confusing and would require more refactoring (I can go over them if you like). You can think of the model as so: Instead of promoting a command up to the level of a function, see through the function and access the command directly. The neat thing about this solution is that the casing really only happens at render time (and searching but only for performance reasons).

minnakt (commented):
not sure why it didn't register my re-review so leaving another comment 👀 (I left comments in the threads above)

apps/parsley/src/utils/filterLogs/index.ts
SupaJoon: I refactored the code so the open state that's pushed to ProcessedLogLines is not related to `containsTopLevelCommand`

apps/parsley/src/components/LogRow/RowRenderer/index.tsx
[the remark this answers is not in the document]
SupaJoon: @minnakt I think we should make this ticket when we need the SectionStatus to do operations on ProcessedLogLines. I don't think the refactor will have a great benefit until then since `includesLineNumber` is a constant time operation.

apps/parsley/src/utils/filterLogs/index.ts
```diff
@@ -80,20 +80,26 @@ const filterLogs = (options: FilterLogsParams): ProcessedLogLines => {
...
+              functionName,
+              isOpen: isFuncOpen,
+              range,
+              rowType: RowType.SectionHeader,
+            });
+          }
           funcIndex += 1;
-          if (!isFuncOpen) {
+          if (!renderCommands) {
             // The function is closed. Skip all log lines until the end of the function.
```
minnakt, line 103: I'm not sure if this comment is clear after the updates (i.e. why does `!renderCommands` mean the function is closed?)

apps/parsley/src/utils/filterLogs/index.ts
minnakt: I still think functions should be open if the command is visible and that the state should reflect that, but if the refactor will address that, then I'm fine with this for now

apps/parsley/src/components/LogRow/RowRenderer/index.tsx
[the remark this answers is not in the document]
minnakt: it's less about the performance benefits (I assume there are none) and more that I think it makes sense to have all of this information in one place. I can make a ticket since I feel pretty strongly about it

minnakt (approved):

apps/parsley/src/utils/filterLogs/index.ts
SupaJoon: Good point, I updated the comment.

apps/parsley/src/utils/filterLogs/index.ts
SupaJoon: Yup, that PR will address how top-level commands are represented in SectionState and SectionData by removing any function placeholders.

PR #328, opened 2024-08-21
DEVPROD-5303: Track sectioning analytics

SupaJoon:
DEVPROD-5303

### Description
These code changes are based off of https://github.com/evergreen-ci/ui/pull/325
Allow tracking section toggling with insights on command vs function, nested vs unnested and section pass/fail status when applicable.
Also allow tracking if the log loaded with jump to failing line and sectioning.

Have you have updated the analytics documentation if necessary?  
The analytics doc has been updated.


apps/parsley/vite.config.ts.timestamp-1724274878501-90fb533d23468.mjs
minnakt, line 1: seems like this file got added accidentally (maybe add to `.gitignore`)?

apps/parsley/src/components/LogRow/SubsectionHeader/index.tsx
```diff
@@ -1,40 +1,55 @@
...
       <CaretToggle
         onClick={() => {
           sendEvent({
-            name: "Toggled section",
+            name: "Toggled section caret",
+            nested: !isTopLevelCommand,
             open: !open,
             "section.name": commandName,
             "section.type": "command",
+            status,
```
minnakt, line 52: I think the attributes should reference their parent object (otherwise we could ask, is it a task status? is it a build status? some other evergreen status?)
```suggestion
        "section.nested": !isTopLevelCommand,
        "section.open": !open,
        "section.status": status,
```

apps/parsley/src/components/LogRow/SectionHeader/index.tsx
```diff
@@ -35,10 +36,12 @@ const SectionHeader: React.FC<SectionHeaderProps> = ({
       <CaretToggle
         onClick={() => {
           sendEvent({
-            name: "Toggled section",
+            name: "Toggled section caret",
+            nested: false,
             open: !open,
             "section.name": functionName,
             "section.type": "function",
+            status,
```
minnakt, line 44: same comment

apps/parsley/src/components/LogRow/SectionHeader/SubsectionControls.tsx
```diff
@@ -23,7 +28,15 @@ const SubsectionControls: React.FC<{ functionID: string }> = ({
       {showExpandButton && (
         <Button
           data-cy="open-subsections-btn"
-          onClick={() => toggleAllCommandsInFunction(functionID, true)}
+          onClick={() => {
+            sendEvent({
+              "function.name": functionName,
+              name: "Clicked open subsections button",
+              status,
```
minnakt, line 35:
```suggestion
              "function.status": status,
```

apps/parsley/src/components/LogRow/SectionHeader/SubsectionControls.tsx
```diff
@@ -32,7 +45,14 @@ const SubsectionControls: React.FC<{ functionID: string }> = ({
       {showCloseButton && (
         <Button
           data-cy="close-subsections-btn"
-          onClick={() => toggleAllCommandsInFunction(functionID, false)}
+          onClick={() => {
+            toggleAllCommandsInFunction(functionID, false);
+            sendEvent({
+              "function.name": functionName,
+              name: "Clicked close subsections button",
+              status,
```
minnakt, line 53:
```suggestion
              "function.status": status,
```

apps/parsley/src/components/LogPane/index.tsx
```diff
@@ -56,6 +58,11 @@ const LogPane: React.FC<LogPaneProps> = ({ rowCount, rowRenderer }) => {
           setPrettyPrint(true);
         }
         performedScroll.current = true;
+        sendEvent({
+          "jump.to.failing.line.enabled": settings.jumpToFailingLineEnabled,
+          name: "Viewed log with sections and jump to failing line",
+          "sections.enabled": settings.sectionsEnabled,
```
minnakt, line 64: possibly want to reference settings as a parent object:
```suggestion
          "settings.jump.to.failing.line.enabled": settings.jumpToFailingLineEnabled,
          name: "Viewed log with sections and jump to failing line",
          "settings.sections.enabled": settings.sectionsEnabled,
```

minnakt (approved):
analytics lgtm

apps/parsley/src/components/LogRow/SubsectionHeader/index.tsx
SupaJoon: I used "section.name" instead of "name" because "name" was already taken. The [docs](https://opentelemetry.io/docs/specs/semconv/general/attribute-naming/) mention that namespaces should be primiarily be used to avoid name collisions and not indicate hierarchies. The docs aren't really clear what is considered a naming collision though. Your example does kind of highlight a naming collision between status types.

apps/parsley/src/components/LogPane/index.tsx
SupaJoon: I think it's okay to not indicate hierarchy here because the names are unique

khelif96: I actually agree with Minna here. I'm planning on reviewing all of our span attributes in our projects and adding this hierarchy. So that it better falls in line with otel standards and avoids excessive dot notation.

`settings.jump_to_failing_line.enabled`
`settings.sections.enabled`

I'm planning on doing this in a follow up pr but I just wanted to shed some light on my intentions.

minnakt: ah that's a good callout, `jump.to.failing.line` probably creates excessive namespaces that we would never use otherwise (`jump`, `jump.to`, `jump.to.failing`)

apps/parsley/.gitignore
```diff
@@ -6,6 +6,7 @@ cypress/screenshots/
 /src/**/*.css
 build.txt
 build.txt-e
+vite.config.ts.timestamp-*
```
khelif96, line 9: What is this?

khelif96 (approved):

apps/parsley/.gitignore
SupaJoon: it's from a vite bug: https://github.com/vitejs/vite/issues/13267

PR #330, opened 2024-08-22
DEVPROD-7659: Support global filter on Event Log page

sophiabarness:
DEVPROD-7659

### Description

Add a search bar at the top of the Event Log page as a global filter. The global filter can be applied on the client side (LeafyGreen table should have built-in fields to support it).

### Screenshots

GlobalSearch

### Testing

Added Jest and Cypress tests.

### DIFF
--- a/apps/spruce/cypress/integration/image/event_log.ts
+++ b/apps/spruce/cypress/integration/image/event_log.ts
@@ -10,7 +10,7 @@ describe("event log page", () => {
     );
   });
 
-  it("should show no events when filtering for a nonexistent item", () => {
+  it("should show no events when filtering by name for a nonexistent item", () => {
     cy.visit("/image/ubuntu2204/event-log");
     cy.dataCy("image-event-log-card").should("have.length", IMAGE_EVENT_LIMIT);
     cy.dataCy("image-event-log-name-filter").first().click();
@@ -20,5 +20,12 @@ describe("event log page", () => {
       .within(() => {
         cy.dataCy("image-event-log-table-row").should("have.length", 0);
       });
+  });
+
+  it("should show no events when global filtering for a nonexistent item", () => {
+    cy.visit("/image/ubuntu2204/event-log");
+    cy.dataCy("image-event-log-card").should("have.length", IMAGE_EVENT_LIMIT);
+    cy.get('input[placeholder="Global search by name"]').type("bogus{enter}");
+    cy.dataCy("image-event-log-table-row").should("have.length", 0);
   });
 });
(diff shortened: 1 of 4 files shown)


apps/spruce/cypress/integration/image/event_log.ts
```diff
@@ -21,4 +21,11 @@ describe("event log page", () => {
         cy.dataCy("image-event-log-table-row").should("have.length", 0);
       });
   });
+
+  it("should show no events when global filtering for a nonexistent item", () => {
+    cy.visit("/image/ubuntu2204/event-log");
+    cy.dataCy("image-event-log-card").should("have.length", IMAGE_EVENT_LIMIT);
+    cy.get('input[placeholder="Global search by name"]').type("bogus{enter}");
```
minnakt, line 28: is it possible to use `dataCy` here? (`dataCy` is generally preferred)

apps/spruce/src/pages/image/tabs/EventLogTab/EventLogTab.test.tsx
```diff
@@ -380,6 +380,48 @@ describe("image event log page", async () => {
       expect(rows).toHaveLength(3);
     });
   });
+
+  it("supports global filter for existing entries", async () => {
```
minnakt, line 384: could we add a test for the functionality of the global filter versus the table filter? I think the global filter probably takes precedence, could we test for that?

minnakt (commented):
looks good overall 🙂

apps/spruce/src/pages/image/ImageEventLog/ImageEventLog.tsx
```diff
@@ -20,11 +22,26 @@ export const ImageEventLog: React.FC<ImageEventLogProps> = ({
...
+      <SearchContainer>
+        <SearchInput
+          aria-labelledby="event-log-global-search"
+          value={globalSearch}
+          onChange={handleGlobalSearchChange}
+          placeholder="Global search by name"
+          data-cy="event-log-global-search"
+        />
+      </SearchContainer>
+
```
minnakt, line 44: nit. remove extra line
```suggestion
```

apps/spruce/src/pages/image/tabs/EventLogTab/EventLogTab.test.tsx
minnakt: bump on this: could we write an additional test `global filter takes precedence over table filters`

and what would happen is that you could apply a name search within the table to show results, then apply a global filter on top of it which would change the displayed results

minnakt (approved):
great work!

PR #348, opened 2024-08-29
DEVPROD-8894: Use caret for LeafyGreen packages

sophstad:
DEVPROD-8894

### Description

- Unpin LG dependencies so they can more accurately resolve their internal dependencies and reduce the amount of manual upgrades required.
- Also bump toast to `^6.1.27`, which allowed removing try/catch in toast tests

### Testing

Updated snapshots


minnakt (approved):
yappee

PR #352, opened 2024-08-29
DEVPROD-5906 Consolidate useConfigurePatch into single folder

khelif96:
DEVPROD-5906

### Description
We had 2 instances of useConfigurePatch that we were making changes to and updating but only one was used in production. 

### Testing
Under the assumption that the one we have on production is the source of truth, I am relying on the existing tests for this functionality.


apps/spruce/src/hooks/useConfigurePatch.ts
minnakt, line 1: I think you might have to merge main and delete this again since the file path changed. So if you merged this PR like it is now, I believe the file won't actually get deleted

khelif96: Great catch!

PR #353, opened 2024-08-29
DEVPROD-8660: Move types to lib

sophstad:
DEVPROD-8660

### Description

- Move `ApolloMock`, `TestStatus`, `TaskStatus`, and util types to shared lib
- Remove `All` and `TimeOut` TestStatus; timeout is [not a status](https://github.com/evergreen-ci/evergreen/blob/e304bc656fcb9ace10a78d45ee7df7d1e6b3f46f/globals.go#L156-L159) and `All` is more of a filter state than an actual status.


apps/parsley/src/components/Badge/TestStatusBadge/TestStatusBadge.stories.tsx
```diff
@@ -1,7 +1,7 @@
 import styled from "@emotion/styled";
+import { TestStatus } from "@evg-ui/lib/types/test";
```
minnakt, line 2: can the `types/test.ts` file be deleted now from Parsley?

packages/lib/src/types/utils.ts
minnakt, line 1: can the `types/utils.ts` file be deleted now from Parsley?

packages/lib/package.json
minnakt, line 1: (random comment placement) I think we could be able to delete the `types/leafygreen.ts` in Spruce because when I remove their usage (it's only used like twice) there aren't any TS errors 👀

minnakt (approved):

PR #135, opened 2024-05-22
DEVPROD-7502: Allow "evg-db-ops --reseed-and-dump" to run from app directories

SupaJoon:
DEVPROD-7502

### Description
The yarn cwd flag was preventing the script from navigating to the `evergreen-ci/evergreen` project directory.

PR #355, opened 2024-08-29
DEVPROD-6289: Consolidate and rename bootstrap-logkeeper script

SupaJoon:
DEVPROD-6289


khelif96 (approved):

apps/spruce/package.json
```diff
@@ -4,7 +4,7 @@
   "private": true,
   "type": "module",
   "scripts": {
-    "bootstrap-logkeeper": "./scripts/bootstrap-logkeeper.sh",
+    "bootstrap-s3-logs": "../../scripts/bootstrap-s3-logs.sh",
```
sophstad, line 7: Yarn commands can be called from anywhere within the Spruce directory and it would be nice to support this rather than failing due to a relative path; would using the `--cwd` path allow that to work?

SupaJoon: --cwd will also adjust the cwd referenced in the script which I think is harder to debug. I ran into that during this [PR](https://github.com/evergreen-ci/ui/pull/135)

PR #356, opened 2024-08-30
DEVPROD-7854: Remove complex version redirect logic

minnakt:
DEVPROD-7854

### Description
Sometimes the redirect logic in the `Version` component doesn't work correctly. When you visit an unconfigured patch, it will occasionally throw this error, even though the logic in the component states that we should fetch the patch and redirect to the `ConfigurePatch` component:

Screenshot 2024-08-30 at 11 09 00 AM

I think we should just remove the redirect logic because it is really hard to understand, and because we can link to the correct routes from the parent component. This is analogous to how the legacy UI handles it ([code](https://github.com/evergreen-ci/evergreen/blob/ea920d63b25e7c1ba814632fe2f3a69fd3624e5a/service/templates/patches.html#L53-L67)).

### Testing
- Updated Cypress tests
- Tested that patch/version links returned from the CLI still work on Spruce

### DIFF
--- a/apps/spruce/src/components/Redirects/PatchRedirect.tsx
+++ b/apps/spruce/src/components/Redirects/PatchRedirect.tsx
@@ -1,7 +1,12 @@
 import { useParams, Navigate } from "react-router-dom";
-import { getVersionRoute, slugs } from "constants/routes";
+import { getPatchRoute, slugs } from "constants/routes";
 
 export const PatchRedirect: React.FC = () => {
   const { [slugs.versionId]: versionId } = useParams();
-  return <Navigate replace to={getVersionRoute(versionId ?? "")} />;
+  return (
+    <Navigate
+      replace
+      to={getPatchRoute(versionId ?? "", { configure: true })}
+    />
+  );
 };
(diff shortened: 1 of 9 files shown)


apps/spruce/src/components/Redirects/PatchRedirect.tsx
```diff
@@ -1,7 +1,12 @@
...
 
 export const PatchRedirect: React.FC = () => {
   const { [slugs.versionId]: versionId } = useParams();
-  return <Navigate replace to={getVersionRoute(versionId ?? "")} />;
+  return (
+    <Navigate
+      replace
+      to={getPatchRoute(versionId ?? "", { configure: true })}
+    />
+  );
```
sophstad, line 11: In DEVPROD-5907 I updated this component so that navigating to `/patch/<id>` for an activated task would redirect to `/version/<id>`. Now that directs to `/patch/<id>/configure` which seems less desirable; could we perhaps still direct to the version page if the patch is activated?

sophstad (changes requested):

apps/spruce/src/components/Redirects/PatchRedirect.tsx
minnakt: what do you think about the logic now 👀

sophstad (approved):
Works great! Seems like a nice lighter weight version of what we were trying to do before

PR #362, opened 2024-09-03
DEVPROD-9478 Update host expiration test to select the first of the next month

khelif96:
DEVPROD-9478

### Description
The host expiration test will now always select the first of the next month which should prevent the test from failing.


sophstad (changes requested):
I think this would fail on the 1st of a 31-day month since the exemption period can only be 30 days long 😅 
https://github.com/evergreen-ci/ui/blob/a0aebdcd8d0740c159db9c2facdbdd10096ba572/apps/spruce/src/components/Spawn/editVolumeModal/getFormSchema.tsx#L76

apps/spruce/cypress/integration/spawn/host.ts
```diff
@@ -242,7 +242,25 @@ describe("Navigating to Spawn Host page", () => {
...
-    cy.get("td[aria-current=true]").next().click();
+    cy.get("td[aria-current=true]").as("currentDateCell");
+    // Select a date in the future either this month or next month
+    // if the current date is the last day of the month select the first day of the next month
+    // if it is not select the next day
+    // We can determine if the current date is the last day of the month if the next cell is disabled
+    cy.get("@currentDateCell")
+      .next()
+      .then(($el) => {
+        if ($el.attr("aria-disabled") === "false") {
```
sophstad, line 253: I _think_ this would fail if the last day of the month fell on a Saturday; there would be no next `td` to grab. We can probably do this
```suggestion
        if (!$el || $el.attr("aria-disabled") === "false") {
```

sophstad (approved):
Nice work, this was deceptively tricky 😆

PR #374, opened 2024-09-10
DEVPROD-8657 Migrate test_utils to lib

khelif96:
DEVPROD-8567

### Description
Migrates test_utils to the lib package and reduces duplication. Sorry for the large PR most of it is just file moves and import statement changes.
I also updated the tsconfig in the lib folder to allow for absolute imports.


apps/spruce/tsconfig.json
```diff
@@ -13,8 +13,6 @@
   "include": [
     "src",
     "scripts",
-    "test_utils/index.tsx",
-    "test_utils/custom-queries.ts",
```
khelif96, line 17: Removed these from this file

khelif96 (commented):
Here are the files that actually had changes. The rest are file moves and import statement updates

packages/lib/src/test_utils/utils.ts
```diff
@@ -1,18 +1,39 @@
...
+   * @param value - The value to set the environment variable to.
+   */
+  mockEnv: (variable: string, value: string) => void;
+};
+
+/**
+ * Creates mock environment variables for testing purposes.
+ * @returns An object with `cleanup` and `mockEnv` functions.
+ */
+export const mockEnvironmentVariables = (): MockEnvironmentVariables => {
```
sophstad, line 18: I think we could switch to Vitest's [`stubEnv`](https://vitest.dev/api/vi.html#vi-stubenv)

Maybe you could add a `@deprecated` tag here?

packages/lib/src/test_utils/hooks.ts
```diff
@@ -64,4 +64,30 @@ const useEffectDebugger = (
...
+/**
+ * `renderComponentWithHook` is a utility function that renders a component with a given hook for use in testing
+ * @param useHook - The hook to use
+ * @param Comp - The component to render
+ * @returns - The component and the hook
+ */
+const renderComponentWithHook = <
+  T extends () => any,
+  U extends JSX.Element | null,
+>(
```
minnakt, line 76: Does this belong in the `hooks` folder? The other functions in this file are true hooks, but this is a render function to help unit testing hooks. Maybe it belongs in `index`?

apps/spruce/src/test_utils/index.tsx
minnakt, line 1: Can we also delete `test_utils/graphql.tsx`? I'm pretty sure it's only used in `BuildBaron.test.tsx` and that you could just replace it with the one from the apollo library and it will pass

minnakt, line 1: Any reason not to move up `test_utils/toast-decorator.tsx` as well? it exists in both Spruce and Parsley

apps/spruce/src/test_utils/utils.ts
```diff
@@ -33,4 +11,4 @@ const selectLGOption = async (dataCy: string, option: string) => {
   expect(screen.queryByDataCy(dataCy)).toHaveTextContent(option);
 };
 
-export { mockEnvironmentVariables, selectLGOption };
+export { selectLGOption };
```
minnakt, line 14: [request] Could we fix the errors in `selectLGOption`, it's really easy just replace 
`screen.queryByDataCy` => `screen.getByDataCy`
`screen.queryByText` => `screen.getByText`

packages/lib/src/test_utils/utils.ts
khelif96: Removed it and replaced it all with stubEnv!

packages/lib/src/test_utils/hooks.ts
khelif96: Good call out!

apps/spruce/src/test_utils/index.tsx
khelif96: It relies on `context/toast` and I felt it was out of scope to also move that over. That work will be covered by https://jira.mongodb.org/browse/DEVPROD-10167

minnakt: thanks for updating the ticket

minnakt (approved):

PR #375, opened 2024-09-11
DEVPROD-5790: Make parseSortString more generic

SupaJoon:
DEVPROD-5790

### Description
The tests table name sort was failing because the sort categories were hardcoded into the `parseSortString` function. I updated the typing allow defining the categories as well as the output keys.
The new code is a little more verbose but now the `parseSortString` output typing can be linked directly to GQL input types. It's also a lot harder to pass in incorrect parameters and this code will be reused for the task duration table sort improvements (DEVPROD-828).


SupaJoon:
evergreen retry

SupaJoon:
evergreen retry

sophstad (changes requested):

PR #381, opened 2024-09-12
DEVPROD-11316 Update deploy script to include packages directory when generating change log

khelif96:
DEVPROD-11316

### Description
The deploy script was ignoring changes in the packages directory when generating a change log.

### Testing
image


packages/deploy-utils/src/utils/git/index.ts
```diff
@@ -41,13 +41,14 @@ const getCommitMessages = (
...
   const appDir = resolve(gitRoot, "apps", app);
+  const packageDir = resolve(gitRoot, "packages");
   const excludeDir = resolve(
     gitRoot,
     "apps",
     app === "spruce" ? "parsley" : "spruce",
   );
   const commitMessages = execTrim(
-    `git log ${fromCommit}..${toCommit} --oneline -- ${appDir} '!${excludeDir}'`,
+    `git log ${fromCommit}..${toCommit} --oneline -- ${appDir} ${packageDir} '!${excludeDir}'`,
```
sophstad, line 51: Does this have the intented effect? i.e. everything except the excludeDir?
```suggestion
    `git log ${fromCommit}..${toCommit} --oneline -- '!${excludeDir}'`,
```

khelif96: Unfortunately not
image
I believe you need to provide a path to include when trying to limit output in git operations otherwise git will only perform the exclusion but not actually include anything.

 In [this example](https://stackoverflow.com/questions/5685007/making-git-log-ignore-changes-for-certain-paths) they use the `.` to include all files and then exclude ones using the `!` operator.

sophstad: Any reason we shouldn't also use `. !${excludeDir}`?

khelif96: That would include all files in the repo including those that aren't actually used in the apps. For example a change to the README would be included as a commit for the deploy.

sophstad (approved):

PR #383, opened 2024-09-16
DEVPROD-9738: Match on user defined failing commands

SupaJoon:
DEVPROD-9738

### Description
It's possible for the user to force a failed status on a task by calling the task_status API within a command. In this case, there may not be an explicit failure logged but the command section should still be matched on with the jump to failing line feature.

### DIFF
--- a/apps/parsley/src/utils/string/index.ts
+++ b/apps/parsley/src/utils/string/index.ts
@@ -135,7 +135,14 @@ export const trimSeverity = (line: string) => {
  * @returns true if line is the failing log line, false otherwise
  */
 export const isFailingLine = (line: string, failingCommand: string) => {
+  console.log(failingCommand);
   const failedExpression = `${failingCommand} failed`;
   const timeoutExpression = `${failingCommand} stopped early`;
-  return line.includes(failedExpression) || line.includes(timeoutExpression);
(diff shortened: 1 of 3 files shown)


apps/parsley/src/utils/string/index.ts
```diff
@@ -135,7 +135,14 @@ export const trimSeverity = (line: string) => {
  * @returns true if line is the failing log line, false otherwise
  */
 export const isFailingLine = (line: string, failingCommand: string) => {
+  console.log(failingCommand);
```
minnakt, line 138:
```suggestion
```

apps/parsley/src/utils/string/index.ts
```diff
@@ -135,7 +135,14 @@ export const trimSeverity = (line: string) => {
...
   const failedExpression = `${failingCommand} failed`;
   const timeoutExpression = `${failingCommand} stopped early`;
-  return line.includes(failedExpression) || line.includes(timeoutExpression);
+  // The "Finished" command is matched when the task status is set to fail by the user.
+  const finished = `Finished command ${failingCommand}`;
```
minnakt, line 142: nit. could we call it `finishedExpression` (or something similar) to match the other variables? `finished` sounds like it might be a boolean

PR #390, opened 2024-09-16
docs: Mark status as deprecated in version page logic decision doc

minnakt:
### Description
Update the Version Page Logic decision docs since the logic described in this doc was removed.


apps/spruce/docs/decisions/2023-12-13_version_page_logic.md
```diff
@@ -4,6 +4,15 @@
 - date: 2023-12-13
```
khelif96, line 4: The status should be deprecated 
https://github.com/joelparkerhenderson/architecture-decision-record/tree/main/locales/en/templates/decision-record-template-by-michael-nygard

khelif96 (changes requested):

khelif96 (approved):

PR #391, opened 2024-09-16
DEVPROD-9268: Omit secrets from local deploys

minnakt:
DEVPROD-9268

### Description
Publishes `.env-cmdrc.json` to the repo, sensitive secrets removed. This is to overwrite everyone's locally stored copy of `.env-cmdrc.json` and wipe secrets on their machine.

### Testing
- Deployed to staging with `--force` flag and verified that the app still works


sophstad (commented):
I discussed this with @khelif96 previously and I believe we came to the conclusion that `BUCKET` is safe to publish—or let me know if you disagree Mohamed!

packages/deploy-utils/src/setup-credentials/index.ts
sophstad, line 1: A lot of variables in this file are now defined in `.env-cmdrc.json`, do you think we could delete them here?

minnakt:
confirmed with Infrastructure team that publishing bucket names is safe

packages/deploy-utils/src/setup-credentials/index.ts
minnakt: merged them!

.evergreen/deploy.yml
```diff
@@ -4,7 +4,6 @@ functions:
       params:
         working_dir: ui/${app_dir}
```
sophstad, line 5: I believe we could use `add_expansions_to_env` instead of copying all of these 👀

minnakt: done and checked that sentry reporting was working on staging

packages/deploy-utils/src/setup-credentials/index.ts
```diff
@@ -26,60 +28,16 @@ const sharedKeys = {
 const base: Record<TargetEnvironment, Record<string, string | undefined>> = {
```
sophstad, line 28: Could we delete the Sentry values above?

packages/deploy-utils/src/setup-credentials/index.ts
```diff
@@ -26,60 +28,16 @@ const sharedKeys = {
...
-    // TODO: Deduplicate Spruce and Parsley vars
-    REACT_APP_GRAPHQL_URL:
-      "https://evergreen-staging.corp.mongodb.com/graphql/query",
-    REACT_APP_GQL_URL:
-      "https://evergreen-staging.corp.mongodb.com/graphql/query",
-
-    REACT_APP_HONEYCOMB_BASE_URL:
-      "https://ui.honeycomb.io/mongodb-4b/environments/staging",
     REACT_APP_HONEYCOMB_ENDPOINT: process.env.HONEYCOMB_ENDPOINT_STAGING,
     REACT_APP_HONEYCOMB_INGEST_KEY: process.env.HONEYCOMB_INGEST_KEY_STAGING,
```
sophstad, line 32: What if these variables were added to a task's environment via a function variable passed to the `env` param? Then we could specify the value for staging vs beta vs prod tasks, and potentially delete this file altogether 👀

sophstad (commented):
Thanks for all the iterating!

packages/deploy-utils/src/setup-credentials/index.ts
minnakt: I don't think so, `add_expansions_to_env` adds the expansions as environment variables but I think they still need to be added to the JSON manually. I tested by deploying to staging via the deploy task, and couldn't receive any reports to Sentry

sophstad: Interesting, I think that has to do with env-cmd overriding the variables set in the environment with the file variables. We could potentially use the `--no-override` [flag](https://www.npmjs.com/package/env-cmd#-help). Any thoughts?

apps/spruce/.env-cmdrc.json
```diff
@@ -0,0 +1,72 @@
+{
+  "local": {
+    "NODE_ENV": "production",
```
sophstad, line 3: Could you file a ticket about the `NODE_ENV` bug if there isn't one already? Is very weird

apps/spruce/.env-cmdrc.json
sophstad, line 1: Wahoo, doing DEVPROD-812 would be a lot easier now thanks to this!

apps/spruce/.env-cmdrc.json
sophstad: I wonder now if there's a chance `env-cmd` was clobbering Vite's `NODE_ENV` and perhaps `--no-override` would fix that. I'll leave a followup comment in the ticket :)

apps/spruce/README.md
```diff
@@ -230,19 +225,8 @@ following:
...
-
 ### How to Deploy:
 
-Run one of the following commands to deploy to the appropriate environment
-
-1. `yarn deploy:prod` = deploy to https://spruce.mongodb.com
-2. `yarn deploy:staging` = deploy to https://spruce-staging.corp.mongodb.com
-3. `yarn deploy:beta` = deploy to https://spruce-beta.corp.mongodb.com (Beta
-   connects to the production backend)
+For production, use `yarn deploy:prod` to push a git tag and trigger a new build. In case of emergency (i.e. Evergreen, GitHub, or other systems are down), a production build can be pushed directly to S3 with `yarn deploy:prod --force`.
```
sophstad, line 230: Optional improvement -- could add these details to a README in deploy-utils and link there from each app's README to centralize identical info

sophstad (approved):
Great work!! 💯

apps/spruce/.env-cmdrc.json
minnakt: created: [DEVPROD-11638](https://jira.mongodb.org/browse/DEVPROD-11638)

PR #395, opened 2024-09-18
DEVPROD-828: Add multisort to TaskDurationTable

SupaJoon:
DEVPROD-828

### Description
These code changes add multisort to the Task Duration Table by utilizing the `useTableSort` hook. 
Tickets produced during this work: 
https://jira.mongodb.org/browse/DEVPROD-11421
https://jira.mongodb.org/browse/DEVPROD-11539
https://jira.mongodb.org/browse/DEVPROD-11537
https://jira.mongodb.org/browse/DEVPROD-11422

### DIFF
--- a/apps/spruce/src/hooks/useTableSort/index.ts
+++ b/apps/spruce/src/hooks/useTableSort/index.ts
@@ -8,6 +8,7 @@ const { getSortString } = queryString;
 
 interface Props {
   sendAnalyticsEvents?: (sorter: SortingState) => void;
+  // TODO: DEVPROD-11539 - Remove this prop and make the default behavior to use a single query param.
   singleQueryParam?: boolean;
 }
 
--- a/apps/spruce/src/hooks/useTableSort/index.ts
+++ b/apps/spruce/src/hooks/useTableSort/index.ts
@@ -7,7 +7,7 @@ import { queryString } from "utils";
 const { getSortString } = queryString;
 
 interface Props {
-  sendAnalyticsEvents?: (sorter?: SortingState) => void;
+  sendAnalyticsEvents?: (sorter: SortingState) => void;
 }
 
 type CallbackType = (sorter: SortingState) => void;
@@ -31,7 +31,6 @@ export const useTableSort = (props?: Props): CallbackType => {
       [TableQueryParams.SortDir]: undefined,
       [TableQueryParams.SortBy]: undefined,
     };
-
     if (sorter.length === 1) {
       const { desc, id } = sorter[0];
       // @ts-expect-error: FIXME. This comment was added by an automated script.
--- a/apps/spruce/src/hooks/useTableSort/index.ts
+++ b/apps/spruce/src/hooks/useTableSort/index.ts
@@ -31,6 +31,7 @@ export const useTableSort = (props?: Props): CallbackType => {
       [TableQueryParams.SortDir]: undefined,
       [TableQueryParams.SortBy]: undefined,
     };
+
     if (sorter.length === 1) {
       const { desc, id } = sorter[0];
       // @ts-expect-error: FIXME. This comment was added by an automated script.
--- a/apps/spruce/src/hooks/useTableSort/index.ts
+++ b/apps/spruce/src/hooks/useTableSort/index.ts
@@ -8,6 +8,7 @@ const { getSortString } = queryString;
 
 interface Props {
   sendAnalyticsEvents?: (sorter: SortingState) => void;
+  singleQueryParam?: boolean;
 }
 
 type CallbackType = (sorter: SortingState) => void;
@@ -32,7 +33,7 @@ export const useTableSort = (props?: Props): CallbackType => {
       [TableQueryParams.SortBy]: undefined,
     };
 
-    if (sorter.length === 1) {
+    if (sorter.length === 1 && !props?.singleQueryParam) {
       const { desc, id } = sorter[0];
       // @ts-expect-error: FIXME. This comment was added by an automated script.
       nextQueryParams[TableQueryParams.SortDir] = desc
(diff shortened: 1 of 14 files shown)


apps/spruce/src/pages/version/useQueryVariables/index.ts
```diff
@@ -17,37 +19,39 @@ export const useQueryVariables = (
...
+  const taskSortCategories: string[] = Object.values(TaskSortCategory);
+  const parsedSortBy = getString(queryParams[TableQueryParams.SortBy]);
+  const sortBy = taskSortCategories.includes(parsedSortBy)
+    ? (parsedSortBy as TaskSortCategory)
+    : undefined;
+  const parsedSortDir = getString(queryParams[TableQueryParams.SortDir]);
+  const sortDir =
+    parsedSortDir === SortDirection.Desc
+      ? SortDirection.Desc
+      : SortDirection.Asc;
```
sophstad, line 39: I think you can type assert instead
```suggestion
  const sortDir = getString(queryParams[TableQueryParams.SortDir]) as SortDirection;
```

apps/spruce/src/pages/version/taskDuration/TaskDurationTable.tsx
```diff
@@ -257,20 +274,26 @@ const getInitialParams = (queryParams: {
...
 
+  let initialSort: SortingState = [];
+  if (sortBy && sortDir) {
+    initialSort = [
+      {
+        id: sortCategoryToColumnId[sortBy],
+        desc: sortDir === SortDirection.Desc,
+      },
+    ];
+  } else if (sorts) {
```
sophstad, line 285: Is it possible to just use `sorts` and remove the `sortBy`/`sortDir` parameters? Using just a single item in `sorts` would be a lot cleaner than having two separate cases for single vs multisort. We do this in the tests table for example

apps/spruce/src/pages/version/useQueryVariables/index.ts
```diff
@@ -17,37 +19,39 @@ export const useQueryVariables = (
...
+  // 'sortBy', 'sortDir' and 'sorts' are set by useTableSort in <TaskDurationTable />
+  // 'sorts' is set by <PatchTasksTable />
```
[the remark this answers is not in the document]
SupaJoon: https://jira.mongodb.org/browse/DEVPROD-11422

apps/spruce/src/pages/version/taskDuration/TaskDurationTable.tsx
SupaJoon: I realized it would be a poor experience to temporarily utilize `sortDir`/`sortBy` so I added a casing in useTableSort to tell it when to only use 1 query parameter as an intermediate step before `sortDir` and `sortBy` is removed for in the rest of the application.

apps/spruce/src/pages/version/useQueryVariables/index.ts
SupaJoon: I removed this code because I adjusted the table back to using `sorts` only

apps/spruce/src/hooks/useTableSort/index.ts
```diff
@@ -7,7 +7,8 @@ import { queryString } from "utils";
 const { getSortString } = queryString;
 
 interface Props {
-  sendAnalyticsEvents?: (sorter?: SortingState) => void;
+  sendAnalyticsEvents?: (sorter: SortingState) => void;
+  singleQueryParam?: boolean;
```
sophstad, line 11: Could you add a comment referencing the ticket that will implement singleQueryParam as the default for all tables? That way it's clear it should be removed

apps/spruce/src/pages/version/TaskDuration.tsx
```diff
@@ -101,12 +100,18 @@ const TaskDuration: React.FC<Props> = ({ taskCount }) => {
...
-      <TaskDurationTable
-        loading={loading}
-        // @ts-expect-error: FIXME. This comment was added by an automated script.
-        numLoadingRows={limit}
-        tasks={tasksData}
-      />
+      {
+        // Ensures that the TaskDurationTable initial sort
+        // button states intialize with the correct default values.
+        hasInitialized && (
```
sophstad, line 106: Would it be possible to check if there is a sort defined instead of adding this new variable?

apps/spruce/src/pages/version/taskDuration/TaskDurationTable.tsx
```diff
@@ -145,6 +145,7 @@ export const TaskDurationTable: React.FC<Props> = ({
         header: "Build Variant",
         size: 150,
         enableColumnFilter: true,
+        enableSorting: true,
```
sophstad, line 148: What's the reason for adding `enableSorting` to these columns? Other tables like the TestsTable just set `enableMultisort` at the top level of the table since all columns are usable in multisort

apps/spruce/src/pages/version/TaskDuration.tsx
SupaJoon: It's not possible. The `sorts` query param may be defined in the URL from the task table so it's not sufficient to check whether the variable is defined. This worked without this before because the code was setting the default value twice (once in this useEffect and once in getInitialParams) and also because the TaskDuration and Tasks table utilized different query params for the initial value (sortBy/sortDir vs sorts)

apps/spruce/src/pages/version/taskDuration/TaskDurationTable.tsx
SupaJoon: I think this is required. I tried removing the `enableSorting` from each column and the sorts don't show up. TestsTable applies this value to the columns. Check it out [here](https://github.com/evergreen-ci/ui/blob/7a69ecadec70775e06d1dd8b01793117ad0824ff/apps/spruce/src/pages/task/taskTabs/testsTable/getColumnsTemplate.tsx#L36)

sophstad (approved):

SupaJoon:
evergreen retry

SupaJoon:
All tests in the `test` task are passing even though the status is failed.

PR #396, opened 2024-09-18
DEVPROD-10319: Fix time conversion logic in msToDuration

minnakt:
DEVPROD-10319

### Description
A user reported that the task durations were out of order. It seems like the values in the task duration table are sorted correctly, but the value that we report is not correct.

### Screenshots

(Properly renders `1d` now)

Screenshot 2024-09-18 at 12 40 12 PM

### Testing
- Added unit test


SupaJoon (approved):
Good one!

PR #398, opened 2024-09-19
DEVPROD-7340 Move ConditionalWrapper to lib and setup vitest for jsdom testing in lib

khelif96:
DEVPROD-7340

### Description
This is the first part in a series of PRs in order to help breakup the setup work required to move the LG table component over to the lib package.

Move the ConditionalWrapper component to lib as well as initialize the testing infrastructure in the lib package to support jsdom based matchers.


packages/lib/config/vitest/setupTests.ts
sophstad, line 1: This needs to be referenced in `vitest.config.ts` to take effect

packages/lib/config/vitest/global-setup.ts
sophstad, line 1: Same as above

sophstad (changes requested):

packages/lib/config/vitest/setupTests.ts
khelif96: Thanks!

packages/lib/config/vitest/setupTests.ts
```diff
@@ -0,0 +1,21 @@
...
+import "vitest-canvas-mock";
+
+// @ts-ignore
+window.crypto.randomUUID = (() => {
+  let value = 0;
+  return () => {
+    value += 1;
+    return value.toString();
+  };
+})();
```
minnakt, line 12: is this block necessary? If not, could we delete it, and if it is, can we leave a comment about why it's necessary?

PR #399, opened 2024-09-19
DEVPROD-10186: Add waterfall active version labels

sophstad:
DEVPROD-10186

### Description

- Add labels with links and analytics for active versions.
- Implement useSuspenseQuery to leverage suspense while loading waterfall
- Add base CSS grid, tbd if this will stick around pending Safari performance once tasks are introduced
- Updated analytics spreadsheet to note that waterfall page is included in ProjectHealthPage event object

### Screenshots

image

### Testing

Added stories for versions with upstream projects and git tags

### DIFF
--- a/apps/spruce/src/pages/waterfall/VersionLabel.tsx
+++ b/apps/spruce/src/pages/waterfall/VersionLabel.tsx
@@ -2,7 +2,7 @@ import styled from "@emotion/styled";
 import { Body, InlineCode } from "@leafygreen-ui/typography";
 import { Link } from "react-router-dom";
 import { Unpacked } from "@evg-ui/lib/types/utils";
-import { useProjectHealthAnalytics } from "analytics/projectHealth/useProjectHealthAnalytics";
+import { useWaterfallAnalytics } from "analytics";
 import { StyledRouterLink } from "components/styles";
 import { getVersionRoute, getTriggerRoute } from "constants/routes";
 import { WaterfallQuery } from "gql/generated/types";
@@ -29,9 +29,7 @@ export const VersionLabel: React.FC<VersionFields> = ({
   const spruceConfig = useSpruceConfig();
   const jiraHost = spruceConfig?.jira?.host ?? "";
 
-  const { sendEvent } = useProjectHealthAnalytics({
-    page: "Waterfall",
-  });
+  const { sendEvent } = useWaterfallAnalytics();
 
   return (
     <VersionContainer>
--- a/apps/spruce/src/analytics/index.ts
+++ b/apps/spruce/src/analytics/index.ts
@@ -15,3 +15,4 @@ export { useTaskQueueAnalytics } from "./taskQueue/useTaskQueueAnalytics";
 export { useUserPatchesAnalytics } from "./patches/useUserPatchesAnalytics";
 export { useProjectPatchesAnalytics } from "./patches/useProjectPatchesAnalytics";
 export { useVersionAnalytics } from "./version/useVersionAnalytics";
+export { useWaterfallAnalytics } from "./waterfall/useWaterfallAnalytics";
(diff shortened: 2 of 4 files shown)


apps/spruce/src/pages/waterfall/index.tsx
```diff
@@ -1,17 +1,13 @@
...
-
-  return (
-    <PageWrapper data-cy="waterfall-page">
-      Waterfall Page for {projectIdentifier}
-    </PageWrapper>
-  );
-};
+const Waterfall: React.FC = () => (
+  <PageWrapper data-cy="waterfall-page">
+    <Suspense fallback={<h3>Loading waterfall...</h3>}>
```
khelif96, line 7: Sweet!

apps/spruce/src/analytics/projectHealth/useProjectHealthAnalytics.ts
```diff
@@ -3,7 +3,11 @@ import { AnalyticsIdentifier } from "analytics/types";
 import { ProjectHealthView } from "gql/generated/types";
 
 // The comments below are used to indicate which pageType the action is relevant to (e.g. "Commit chart")
-type pageType = "Commit chart" | "Task history" | "Variant history";
+type pageType =
+  | "Commit chart"
+  | "Task history"
+  | "Variant history"
+  | "Waterfall";
```
khelif96, line 10: How do you feel about just creating an entirely new analytics object/hook for this? 
It's a new page and we are moving away from the project health verbiage it could be nice to start over with these analytics attributes.

apps/spruce/src/pages/waterfall/VersionLabel.tsx
```diff
@@ -0,0 +1,106 @@
...
+const VersionContainer = styled.div`
+  > * {
+    font-size: 12px;
+    line-height: 1.3;
+  }
+`;
+
+const CommitMessage = styled(Body)`
+  display: -webkit-box;
+  -webkit-box-orient: vertical;
```
khelif96, line 103: [The weird webkit flexbox way](https://css-tricks.com/line-clampin/#aa-the-standardized-way) 😭

apps/spruce/src/pages/waterfall/VersionLabel.tsx
```diff
@@ -0,0 +1,106 @@
...
+              upstreamOwner: upstreamProject.owner,
+              upstreamRepo: upstreamProject.repo,
+            })}
+          >
+            {upstreamProject.project}
+          </StyledRouterLink>
+        </Body>
+      )}
+      {/* @ts-expect-error */}
+      <CommitMessage title={message}>
```
khelif96, line 79: I tried this out in prod and noticed some weird formatting I tried this out with prod data and noticed some weird formatting 
image

khelif96 (commented):
Great work so far!

apps/spruce/src/pages/waterfall/VersionLabel.tsx
sophstad: I think this is expected for now as we don't yet have logic to determine the number of columns to render on the page. I opted not to set a fixed width on each column as I don't think that benefits the layout. So the extra width seems ok at the moment! Unless there's something else you thought was weird

apps/spruce/src/analytics/projectHealth/useProjectHealthAnalytics.ts
sophstad: That does make sense, I thought it could be nice to avoid redundancy but I think a separate object will ultimately be easier to manage. Will do!

khelif96 (approved):

apps/spruce/src/pages/waterfall/VersionLabel.tsx
khelif96: Just wanted to make sure that the lines not wrapping for the 1st,3rd, and 5th commit were expected. But if its only temporary that sgtm!

sophstad: Ah, that's because the other commits have short messages and Git Tag is rendered on its own line so as not to be cut off 👍

PR #407, opened 2024-09-24
DEVPROD-610 Create Test Analysis Tab

khelif96:
DEVPROD-610

### Description
Creates new test analysis tab


apps/spruce/src/pages/version/TestAnalysis/index.tsx
```diff
@@ -0,0 +1,215 @@
...
+    searchValue.length > 0;
+
+  const hasMatchingResults = totalFilteredTestCount > 0;
+  const hasResults = data && totalTestCount > 0;
+  return (
+    <div>
+      {loading ? (
+        <ListSkeleton />
+      ) : (
+        <div>
```
sophstad, line 89: [nit] Can use a fragment here since there's already a parent div

```suggestion
        <>
```

apps/spruce/src/pages/version/TestAnalysis/index.tsx
```diff
@@ -0,0 +1,215 @@
...
+          <Title>
+            {numberOfTestsThatFailedOnMoreThanOneTask}{" "}
+            {pluralize("test", numberOfTestsThatFailedOnMoreThanOneTask)} failed
+            across more than one task
+          </Title>
+          <Body>
+            This page shows tests that failed across more than one task. If a
```
sophstad, line 96: > This page shows tests that failed across more than one task

This is technically true, but the tab is actually showing all failed tests. Could we reword this so as not to imply that the page _only_ shows tests that failed across more than one task?

apps/spruce/src/pages/version/TestAnalysis/index.tsx
```diff
@@ -0,0 +1,215 @@
...
+            <div>
+              <LabelWrapper>
+                <Label
+                  htmlFor="test-failure-search"
+                  id="test-failure-search-label"
+                >
+                  Search Test Failures
+                </Label>
+              </LabelWrapper>
+              <SearchInput
```
sophstad, line 110: [nit] It would be nice if the focus overflow wasn't cut off
image

apps/spruce/src/pages/version/TestAnalysis/index.tsx
```diff
@@ -0,0 +1,215 @@
...
+            >
+              {taskStatuses.map((status) => (
+                <ComboboxOption
+                  key={status}
+                  displayName={taskStatusToCopy[status]}
+                  value={status}
+                />
+              ))}
+            </Combobox>
+            <Combobox
```
sophstad, line 135: Same as above w.r.t. overflow
image

apps/spruce/src/pages/version/TestAnalysis/index.tsx
```diff
@@ -0,0 +1,215 @@
...
+}
+const TestAnalysis: React.FC<TestAnalysisProps> = ({ versionId }) => {
+  const [selectedTaskStatuses, setSelectedTaskStatuses] = useState<string[]>(
+    [],
+  );
+  const [selectedBuildVariants, setSelectedBuildVariants] = useState<string[]>(
+    [],
+  );
+  const [searchValue, setSearchValue] = useState<string>("");
+  const { data, loading } = useQuery<
```
sophstad, line 40: Loading this tab on prod with lots of failed tests chokes the page. Not sure exactly if the data is inaccurate or where the error originates. We should either fix the error, or implement pagination or some other sort of limit to avoid this situation. I produced the performance slowdown this on the tab for version `66fc06e0c3df2c0007b55c22`. `66fbc5745743660007ed2bd8` is another version that is able to be rendered but has so many tests present that it suffers from a subpar UX imo.

apps/spruce/src/pages/version/TestAnalysis/index.tsx
```diff
@@ -0,0 +1,215 @@
...
+    groupedTestsMap,
+    searchValue,
+    selectedTaskStatuses,
+    selectedBuildVariants,
+  );
+  const groupedTestsMapEntries = Array.from(
+    filteredGroupedTestsMap.entries(),
+  ).sort((a, b) => b[1].length - a[1].length);
+
+  const numberOfTestsThatFailedOnMoreThanOneTask =
```
sophstad, line 72: Preference for consistency with variables below
```suggestion
  const testsThatFailedOnMoreThanOneTaskCount =
```

sophstad (changes requested):

apps/spruce/src/pages/version/TestAnalysis/index.tsx
khelif96: 😮 Great catch this was actually an error in grouping the tasks which caused a test to repeat itself thousands of times if it occurred in multiple tasks. I fixed that logic and made sure it was covered in a test.

apps/spruce/src/components/TaskStatusBadgeWithLink/__snapshots__/TaskStatusBadge_WithCounts.storyshot
```diff
@@ -0,0 +1,208 @@
...
+      </div>
+    </div>
+    <div
+      class="css-1tojx22"
+    >
+      <div
+        class="css-11x3fkx-StyledBadge-badgeWidthMaxContent-badgeWidthMaxContent e19oz5u0 leafygreen-ui-1ciuhbt"
+        data-cy="task-status-badge"
+      >
+        5 Succeededs
```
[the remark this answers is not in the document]
khelif96: Good catch this was from an earlier iteration I've cleaned it up

apps/spruce/src/pages/version/TestAnalysis/GroupedTestMapList/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+const GroupedTestMapList: React.FC<GroupedTestMapListProps> = ({
+  groupedTestsMapEntries,
+}) => (
+  <>
+    {groupedTestsMapEntries.map(([testName, tasks]) => (
+      <SpacedDiv>
+        <FailedTestGroup key={testName} tasks={tasks} testName={testName} />
+      </SpacedDiv>
+    ))}
+  </>
```
minnakt, line 18: Could we use a wrapper container with `gap` (instead of putting `margin`s on each element)?
```suggestion
  <Container>
    {groupedTestsMapEntries.map(([testName, tasks]) => (
      <FailedTestGroup key={testName} tasks={tasks} testName={testName} />
    ))}
  </Container>
);

const Container = styled.div`
  display: flex;
  flex-direction: column;
  gap: ${size.s};
`;
```

apps/spruce/src/pages/version/TestAnalysis/GroupedTestMapList/FailedTestGroup/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+  tasks,
+  testName,
+}) => {
+  const statusMap = countStatuses(tasks);
+  const sortedStatuses = Object.entries(statusMap).sort((a, b) =>
+    a[0].localeCompare(b[0]),
+  );
+  return (
+    <Accordion
+      shouldRenderChildIfHidden={false}
```
minnakt, line 26: Could you explain about what this prop does? I couldn't really tell after toggling it off

apps/spruce/src/components/Accordion/index.tsx
```diff
@@ -108,6 +117,8 @@ const AnimatedAccordion = styled.div<{
         ? "max-height 0.3s cubic-bezier(0, 1, 0, 1)"
         : "max-height 0.6s ease-in-out"
     }`};
+
+  ${({ hide }) => hide && `display: none;`}
```
minnakt, line 121: I think `display: none` disables all animations on the `Accordion` component

apps/spruce/src/pages/version/TestAnalysis/index.tsx
```diff
@@ -0,0 +1,219 @@
...
+  );
+
+  const buildVariants = useMemo(
+    () => getAllBuildVariants(groupedTestsMap),
+    [data],
+  );
+  const taskStatuses = useMemo(
+    () => getAllTaskStatuses(groupedTestsMap),
+    [data],
+  );
```
minnakt, line 67: nit. eslint warnings on these `useMemo`s

apps/spruce/src/pages/version/TestAnalysis/GroupedTestMapList/FailedTestGroup/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+      title={
+        <TitleContainer>
+          <Body weight="medium">
+            {testName} failed on {tasks.length}{" "}
+            {pluralize("task", tasks.length)}
```
minnakt, line 31: is the "failed on" part superfluous given the badges have numbers? Imo it might look cleaner without this part

minnakt (changes requested):

apps/spruce/src/pages/version/TestAnalysis/TestAnalysisTable.tsx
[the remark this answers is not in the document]
khelif96: Its not I renamed and moved it and it must have gotten lost in the git history good catch!

apps/spruce/src/pages/version/TestAnalysis/GroupedTestMapList/FailedTestGroup/index.tsx
khelif96: I thought it would be useful if there are several different statuses in the badges. Reduces the need for some mental math. :) 
image

apps/spruce/src/pages/version/TestAnalysis/index.tsx
```diff
@@ -0,0 +1,242 @@
...
+          </Title>
+          <Body>
+            This page provides an overview of all test failures in the current
+            version. If a test fails across multiple tasks, it could suggest a
+            flaky test or a broader issue.
+          </Body>
+          <FilterContainer>
+            <div>
+              <LabelWrapper>
+                <Label
```
sophstad, line 128: Could you add disabled styling to this? I also wouldn't be opposed to using the regular TextInput since it has more features built in and we're not using the search results components.
image

apps/spruce/src/pages/version/TestAnalysis/index.tsx
```diff
@@ -0,0 +1,242 @@
...
+              ))}
+            </Combobox>
+          </FilterContainer>
+          {hasResults && (
+            <FilterSubheaderContainer>
+              <Body weight="medium">
+                {totalFilteredTestCount}/{totalTestCount} Filtered Failed{" "}
+                {pluralize("Test", totalFilteredTestCount)}
+              </Body>
+              <Button
```
sophstad, line 186: Would it be possible to disable this button if no filters are applied?

apps/spruce/src/pages/version/TestAnalysis/index.tsx
```diff
@@ -0,0 +1,242 @@
...
+                  displayName={bv.buildVariantDisplayName}
+                  value={bv.buildVariant}
+                />
+              ))}
+            </Combobox>
+          </FilterContainer>
+          {hasResults && (
+            <FilterSubheaderContainer>
+              <Body weight="medium">
+                {totalFilteredTestCount}/{totalTestCount} Filtered Failed{" "}
```
sophstad, line 183: Would suggest omitting "filtered" because using a fraction implies filtering in a UI imo. It's also slightly confusing if no filters are applied
```suggestion
                {totalFilteredTestCount}/{totalTestCount} Failed{" "}
```

apps/spruce/src/pages/version/TestAnalysis/utils.ts
```diff
@@ -0,0 +1,161 @@
...
+ * @returns - a map of test names to an array of tasks
+ */
+const groupTestsByName = (
+  tasks: TestAnalysisQueryTasks,
+): Map<string, TaskBuildVariantField[]> => {
+  const testMap = new Map<string, TaskBuildVariantField[]>();
+  const processedTasks = new Set<string>();
+
+  tasks.forEach((task) => {
+    const tests = getTestsInTask(task.tests);
```
sophstad, line 21: `getTestsInTask` seems like it just returns a subset of the fields in `task.tests.testResults`; should we consider skipping this operation and iterating on `task.tests.testResults` directly?

apps/spruce/src/pages/version/TestAnalysis/index.tsx
```diff
@@ -0,0 +1,242 @@
...
+                <ComboboxOption
+                  key={bv.buildVariant}
+                  displayName={bv.buildVariantDisplayName}
+                  value={bv.buildVariant}
+                />
+              ))}
+            </Combobox>
+          </FilterContainer>
+          {hasResults && (
+            <FilterSubheaderContainer>
```
sophstad, line 181: Does this look reasonable on the same line as the filters or is it super cramped? I think it could make sense there since it's part of a logical group

apps/spruce/src/pages/version/TestAnalysis/index.tsx
khelif96: TIL labels have disabled styling. This does use the regular TextInput under the hood with the addition of some validation logic. I switched it to using the default label prop instead of the random extra label. 🤦

apps/spruce/src/pages/version/TestAnalysis/index.tsx
khelif96: image
It does feel pretty cramped

.evergreen/deploy.yml
```diff
@@ -10,6 +10,9 @@ functions:
           elif [ "${target}" = "production" ] || [ "${target}" = "beta" ]; then
             echo "REACT_APP_HONEYCOMB_ENDPOINT: ${HONEYCOMB_ENDPOINT_PROD}" >> honeycomb.yml
             echo "REACT_APP_HONEYCOMB_INGEST_KEY: ${HONEYCOMB_INGEST_KEY_PROD}" >> honeycomb.yml
+          else
+            echo "REACT_APP_HONEYCOMB_ENDPOINT: " >> honeycomb.yml
+            echo "REACT_APP_HONEYCOMB_INGEST_KEY: " >> honeycomb.yml
```
sophstad, line 15: [nit] I think you can delete this change thanks to #437

apps/spruce/src/pages/version/TestAnalysis/index.tsx
```diff
@@ -0,0 +1,235 @@
...
+  gap: ${size.xs};
+`;
+
+const FilterContainer = styled.div`
+  display: flex;
+  justify-content: space-between;
+  margin-bottom: ${size.m};
+  margin-top: ${size.xs};
+  > * {
+    width: 30%;
```
sophstad, line 226: `flex-basis` should probably be preferred to set an item's base width
```suggestion
    flex-basis: 30%;
```

apps/spruce/src/pages/version/TestAnalysis/index.tsx
```diff
@@ -0,0 +1,235 @@
...
+            {pluralize("test", numberOfTestsThatFailedOnMoreThanOneTask)} failed
+            across more than one task
+          </Title>
+          <Body>
+            This page provides an overview of all test failures in the current
+            version. If a test fails across multiple tasks, it could suggest a
+            flaky test or a broader issue.
+          </Body>
+          <FilterContainer>
+            <div>
```
sophstad, line 130: Using `flex-basis` allows you to remove this div

apps/spruce/src/pages/version/TestAnalysis/index.tsx
```diff
@@ -0,0 +1,235 @@
...
+              >
+                Clear Filters
+              </Button>
+            </FilterSubheaderContainer>
+          )}
+
+          <GroupedTestMapList groupedTestsMapEntries={groupedTestsMapEntries} />
+          {!hasMatchingResults && (
+            <BasicEmptyState
+              description="For additional analytics on tests please visit Honeycomb"
```
sophstad, line 200: Is there any partially prefilled query link we can provide here? I imagine an average user would have no idea where to begin on Honeycomb.

apps/spruce/src/pages/version/TestAnalysis/index.tsx
```diff
@@ -180,10 +177,11 @@ const TestAnalysis: React.FC<TestAnalysisProps> = ({ versionId }) => {
           {hasResults && (
             <FilterSubheaderContainer>
               <Body weight="medium">
-                {totalFilteredTestCount}/{totalTestCount} Filtered Failed{" "}
+                {totalFilteredTestCount}/{totalTestCount} Failed{" "}
                 {pluralize("Test", totalFilteredTestCount)}
               </Body>
               <Button
+                disabled={hasFiltersApplied}
```
sophstad, line 184: Shouldn't disable if there is a filter!
```suggestion
                disabled={!hasFiltersApplied}
```

apps/spruce/src/pages/version/TestAnalysis/index.tsx
```diff
@@ -124,18 +128,11 @@ const TestAnalysis: React.FC<TestAnalysisProps> = ({ versionId }) => {
...
             <div>
-              <LabelWrapper>
-                <Label
-                  htmlFor="test-failure-search"
-                  id="test-failure-search-label"
-                >
-                  Search Test Failures
-                </Label>
-              </LabelWrapper>
               <TextInputWithValidation
```
sophstad, line 131: Sadly performance on this input is super laggy 🙁 I observed this on version `mongodb_mongo_master_ea79e18886a3237740b9a16be5e400dd12edf277` on prod data

apps/spruce/src/pages/version/TestAnalysis/index.tsx
```diff
@@ -124,18 +128,11 @@ const TestAnalysis: React.FC<TestAnalysisProps> = ({ versionId }) => {
...
                 aria-labelledby="test-failure-search-label"
```
sophstad, line 132: Does the plus button on this input do anything? It seems like it should not be present since we only support a single live search

sophstad (changes requested):

apps/spruce/src/pages/version/TestAnalysis/index.tsx
khelif96: hmm I've tried several optimizations. While this performs well on prod when Strict mode is turned off; the poor performance on the local dev server does indicate there is probably an underlying performance problem using live search. I've updated this to now require manual submission like the rest of the inputs on the page. (The user will need to press enter to submit the search) Thanks for calling this out!

apps/spruce/src/pages/version/TestAnalysis/index.tsx
khelif96: I left this back in to fit into the new pattern from above!

apps/spruce/src/pages/version/TestAnalysis/index.tsx
sophstad: Bump

apps/spruce/src/pages/version/TestAnalysis/index.tsx
sophstad: Bump

apps/spruce/src/pages/version/TestAnalysis/index.tsx
sophstad: Bump

sophstad (changes requested):

apps/spruce/src/pages/version/TestAnalysis/index.tsx
khelif96: I totally made this change https://github.com/evergreen-ci/ui/pull/407/commits/3e32d37b2240c805af970f90539c378d71ef7628#diff-4f8210ab3951ea08d400d2c6b4fcfc866f3b71b3d660a72bfe49a40085a2c311R225 but I was burned by my own revert 😞 Thanks for the follow up!

apps/spruce/src/pages/version/TestAnalysis/index.tsx
khelif96: I think the goal behind this on the designs was to try to shift some patterns to honeycomb but taking a look there isn't much I could find for test specific results so I've omitted this description.

apps/spruce/src/pages/version/VersionTabs.tsx
```diff
@@ -123,7 +127,12 @@ const tabMap = ({
       key="test-analysis-tab"
       data-cy="test-analysis-tab"
       id="test-analysis-tab"
-      name="Test Analysis"
+      name={
+        <span ref={refEl as React.Ref<HTMLDivElement>}>
+          Test Analysis
+          <Badge variant="green">New</Badge>
```
sophstad, line 133: Could you add some spacing here? Could also optionally make the badge dependent on `SEEN_TEST_ANALYSIS_TAB_GUIDE_CUE`
image

apps/spruce/src/pages/version/TestAnalysis/index.tsx
```diff
@@ -129,24 +131,38 @@ const TestAnalysis: React.FC<TestAnalysisProps> = ({ versionId }) => {
           <FilterContainer>
             <TextInputWithValidation
               aria-labelledby="test-failure-search-label"
+              defaultValue={testName}
               disabled={!hasResults}
               id="test-failure-search-input"
               label="Search Test Failures"
-              onSubmit={setTestName}
+              onSubmit={(value) => {
```
sophstad, line 138: Would it be possible to make it more clear what search is applied? Something like this could work
image

apps/spruce/src/pages/version/TestAnalysis/index.tsx
```diff
@@ -129,24 +131,38 @@ const TestAnalysis: React.FC<TestAnalysisProps> = ({ versionId }) => {
           <FilterContainer>
             <TextInputWithValidation
               aria-labelledby="test-failure-search-label"
+              defaultValue={testName}
```
sophstad, line 134: This works great 👍

sophstad (changes requested):
So sorry just some tiny thoughts! And e2e tests still appear broken

apps/spruce/src/pages/version/VersionTabs.tsx
khelif96: Good idea!

apps/spruce/src/pages/version/TestAnalysis/index.tsx
khelif96: It should be clear from the filter inputs which is what I tried to do with the defaultValue from your below comment.
image

I think repeating it twice might be too much. Do you think its necessary?

khelif96:
> So sorry just some tiny thoughts! And e2e tests still appear broken

The tests will fail until I merge the backend pr with the new data. But since I want to avoid breaking one of the e2e tests by modifying the task data I am holding off until this is ready to merge. 
You can view the patch where I tested these changes together using set-module:
https://spruce.mongodb.com/version/67092901c14991000769dfc6/tasks

apps/spruce/src/pages/version/VersionTabs.tsx
```diff
@@ -235,18 +229,13 @@ export const VersionTabs: React.FC<VersionTabProps> = ({ version }) => {
...
-      <StyledTabs
-        aria-label="Patch Tabs"
-        selected={selectedTab}
-        setSelected={selectNewTab}
-      >
-        {/* @ts-expect-error: FIXME. This comment was added by an automated script. */}
-        {activeTabs.map((t: string) => allTabs[t])}
-      </StyledTabs>
-    </>
+    <StyledTabs
```
sophstad, line 232: Not sure exactly why your changes are introducing this but I'm seeing an empty tab rendered that isn't on prod (version `67096bb4f3a665000769a76d` with prod data)
image

sophstad (changes requested):

apps/spruce/src/components/TextInputWithValidation/index.tsx
```diff
@@ -45,7 +47,7 @@ const TextInputWithValidation: React.FC<TextInputWithValidationProps> =
       ...rest
```
sophstad, line 47: [nit] I know this is an existing bug but can you reflect the `disabled` status in the `IconButton` too?

apps/spruce/src/pages/version/TestAnalysis/index.tsx
```diff
@@ -0,0 +1,254 @@
...
+                Clear Filters
+              </Button>
+            </FilterSubheaderContainer>
+          )}
+
+          <GroupedTestMapList groupedTestsMapEntries={groupedTestsMapEntries} />
+          {!hasMatchingResults && (
+            <BasicEmptyState
+              description=""
+              title={`No ${hasResults ? "Matching " : ""}Failed Tests Found`}
```
sophstad, line 220: [nit] Sentence case preferred
```suggestion
              title={`No ${hasResults ? "matching " : ""}failed tests found`}
```

sophstad (approved):

apps/spruce/src/pages/version/TestAnalysis/index.tsx
minnakt: I kind of agree, it's hard for me to tell when filters are applied. I feel like there should be filter badges here

It also seems like Clear Filters doesn't actually clear the inputs which makes it look like filters are still applied

apps/spruce/src/pages/version/TestAnalysis/TestAnalysisTabGuideCue/index.tsx
```diff
@@ -0,0 +1,53 @@
+import { useRef, useState } from "react";
+import styled from "@emotion/styled";
+import Badge from "@leafygreen-ui/badge";
+import { GuideCue, TooltipAlign } from "@leafygreen-ui/guide-cue";
+import Cookies from "js-cookie";
+import { size } from "@evg-ui/lib/constants/tokens";
+import { SEEN_TEST_ANALYSIS_TAB_GUIDE_CUE } from "constants/cookies";
+
+export const TestAnalysisTabGuideCue: React.FC = () => {
```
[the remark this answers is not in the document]
khelif96: https://jira.mongodb.org/browse/DEVPROD-12094

apps/spruce/src/pages/version/TestAnalysis/index.tsx
khelif96: I don't think filter badges are necessarily the best choice here because we already use the badges in the comboboxes so having to manage badges in 2 places is confusing. And the test filter input only allows one input while the badge pattern is more commonly used when we allow users to apply multiple filters like we do on the project health page or task history pages.

apps/spruce/src/components/Accordion/index.tsx
```diff
@@ -13,10 +13,13 @@ interface AccordionProps {
...
   titleTag?: React.FC;
+  /** `toggledTitle` replaces the title element when the accordion is open */
   toggledTitle?: React.ReactNode;
   toggleFromBottom?: boolean;
   useIndent?: boolean;
   subtitle?: React.ReactNode;
+  /** This prop prevents the accordion from rendering the child component if the accordion is collapsed. It is useful if the child is expensive to render. */
+  shouldRenderChildIfHidden?: boolean;
 }
 export const Accordion: React.FC<AccordionProps> = ({
```
minnakt, line 24: Could we align the caret at the top?
Screenshot 2024-10-18 at 2 04 53 PM

The badge should probably be aligned too, or in-line with the text

apps/spruce/src/pages/version/TestAnalysis/index.tsx
```diff
@@ -0,0 +1,260 @@
...
+const TestAnalysis: React.FC<TestAnalysisProps> = ({ versionId }) => {
+  const [selectedTaskStatuses, setSelectedTaskStatuses] = useQueryParam<
+    string[]
+  >(TestAnalysisQueryParams.Statuses, []);
+
+  const [selectedBuildVariants, setSelectedBuildVariants] = useQueryParam<
+    string[]
+  >(TestAnalysisQueryParams.Variants, []);
+
+  const [testName, setTestName] = useQueryParam(
```
minnakt, line 48: nit. when submitting with an empty string, it ends up in the URL with `testName=`. I think it should maybe be omitted

apps/spruce/src/pages/version/TestAnalysis/GroupedTestMapList/FailedTestGroup/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+          {sortedStatuses.map(([status, count]) => (
+            <TaskStatusBadge
+              key={status}
+              status={status as TaskStatus}
+              taskCount={count}
+            />
+          ))}
+        </TitleContainer>
+      }
+      useIndent={false}
```
minnakt, line 51: We should maybe use indent here, it makes it clear which accordion is open
```suggestion
      useIndent
```

apps/spruce/src/pages/version/TestAnalysis/index.tsx
```diff
@@ -0,0 +1,260 @@
...
+
+const FilterSubheaderContainer = styled.div`
+  display: flex;
+  margin-bottom: ${size.m};
+  gap: ${size.xs};
+  align-items: center;
+`;
+
+const FilterContainer = styled.div`
+  display: flex;
```
minnakt, line 246: would maybe recommend more `margin-top` on this component

apps/spruce/src/pages/version/TestAnalysis/index.tsx
```diff
@@ -0,0 +1,260 @@
...
+                  name: "Filtered test analysis tab",
+                  "filter.by": "test name",
+                });
+              }}
+              placeholder="Search failed tests (regex)"
+              validator={validateRegexp}
+            />
+            <Combobox
+              disabled={!hasResults}
+              label="Failure type"
```
minnakt, line 151: nit. 
```suggestion
              label="Failure Type"
```

apps/spruce/src/pages/version/TestAnalysis/index.tsx
```diff
@@ -0,0 +1,260 @@
...
+          )}
+        </>
+      )}
+    </Container>
+  );
+};
+
+const Container = styled.div`
+  margin: 0 ${size.xs};
+`;
```
minnakt, line 236: what do you think about increasing the side margin / applying a percentage width on this container? it feels a little bit like things are going far too right into the page

apps/spruce/src/components/Accordion/index.tsx
```diff
@@ -13,10 +13,13 @@ interface AccordionProps {
   showCaret?: boolean;
   title: React.ReactNode;
   titleTag?: React.FC;
+  /** `toggledTitle` replaces the title element when the accordion is open */
   toggledTitle?: React.ReactNode;
   toggleFromBottom?: boolean;
   useIndent?: boolean;
   subtitle?: React.ReactNode;
+  /** This prop prevents the accordion from rendering the child component if the accordion is collapsed. It is useful if the child is expensive to render. */
+  shouldRenderChildIfHidden?: boolean;
```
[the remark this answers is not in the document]
khelif96: I wouldn't want to do it in this pr since that also has the consequence of unmounting the component so it could result in unexpected behaviors for other components that use the Accordion component.

apps/spruce/src/components/TextInputWithValidation/index.tsx
```diff
@@ -45,7 +48,11 @@ const TextInputWithValidation: React.FC<TextInputWithValidationProps> =
...
     } = props;
 
-    const [input, setInput] = useState("");
+    const [input, setInput] = useState(defaultValue);
+    useEffect(() => {
+      setInput(defaultValue);
+    }, [defaultValue]);
+
```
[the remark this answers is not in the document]
khelif96: The pattern I'm using derives the component's state from its props. The `useEffect` ensures that the text input reacts to changes in the `defaultValue` prop. Specifically, I want the internal state to reset, clearing the input field when filters are cleared. This ensures that the component stays in sync with changes from the parent, such as resetting the input when the filter state changes externally.

apps/spruce/src/components/Accordion/index.tsx
khelif96: image
The badges are aligned with the test name. Aligning the caret to the top however seems more difficult. I can't come up with a way to modify the caret styling since that is controlled by the accordion without having it affect the styling for the rest of the accordions in which we want it to be center aligned. Open to suggestions!

minnakt: ooh.... I guess we could introduce a prop like `caretAlign`? kind of like how LeafyGreen has tooltip aligns and justify

apps/spruce/src/pages/version/TestAnalysis/GroupedTestMapList/FailedTestGroup/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+              status={status as TaskStatus}
+              taskCount={count}
+            />
+          ))}
+        </TitleContainer>
+      }
+      toggledTitle={
+        <TitleContainer>
+          <Title>{testName}</Title>
+          {sortedStatuses.map(([status, count]) => (
```
minnakt, line 42: one more thought here; it's pretty jarring that expanding the toggle makes the title take up more horizontal space. Is there a way to make the horizontal space fixed when the accordion gets toggled? i.e., expanding the toggle only expands the title text vertically

minnakt (commented):
last one i promise 🙏

apps/spruce/src/pages/version/TestAnalysis/GroupedTestMapList/FailedTestGroup/index.tsx
khelif96: I tried to make some changes but its pretty difficult to get this effect right. lmk what you think or if you have any suggestions on tweaking it!

apps/spruce/src/pages/version/TestAnalysis/GroupedTestMapList/FailedTestGroup/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+    },
+    {} as Record<string, number>,
+  );
+  return statuses;
+};
+
+const TitleContainer = styled.div`
+  display: flex;
+  gap: ${size.xs};
+  align-items: center;
```
minnakt, line 86: nit. this should be `flex-start `imo
```suggestion
  align-items: flex-start;
```

apps/spruce/src/pages/version/TestAnalysis/GroupedTestMapList/FailedTestGroup/index.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const sortedStatuses = Object.entries(statusMap).sort((a, b) =>
+    a[0].localeCompare(b[0]),
+  );
+  return (
+    <Accordion
+      caretAlignSelf="baseline"
+      shouldRenderChildIfHidden={false}
+      title={
+        <TitleContainer>
+          <Title allowWrap={false} title={testName}>
```
minnakt, line 30: I think the approach before was better (I don't really like this `allowWrap` pattern). Let's just leave it for now because I can't figure out the best way to display this. I suspect something is wrong with the design pattern in the first place and that accordion titles should never be that long

minnakt (approved):
yay!

PR #409, opened 2024-09-24
DEVPROD-6249: Automatically assign release version to deploy

SupaJoon:
DEVPROD-6249

### Description
These code changes will detect [major] or [minor] (case insensitive) in commit titles to determine how the deploy should be versioned when running the prod deploy script.


sophstad (changes requested):
This is really cool, could you update the pull request template with usage instructions?

.github/pull_request_template.md
```diff
@@ -1,4 +1,5 @@
 DEVPROD-NNNNN
+<!-- Does this PR have a minor or major release change? Include [minor] or [major] in the title ☝️ -->
```
khelif96, line 2: Can we write down and link to some guidance on what exactly constitutes a major or minor change?

SupaJoon: Since we are using semver, there is already official guidance [documented](https://semver.org/). Since this is new and low stakes, I think we should see how the team uses the tool and go through any discussion that arises from that instead of making a decision framework right away.

SupaJoon: I created this ticket to come up with some ideas: https://jira.mongodb.org/browse/DEVPROD-11627

sophstad: Agreed Arjun, maybe we should just mention/link to Semver here mostly for the sake of the occasional app engineer opening a PR

sophstad (approved):

PR #412, opened 2024-09-26
DEVPROD-10188: Render waterfall

sophstad:
DEVPROD-10188

### Description

- Render tasks for builds
- Some performance techniques
  - Memoize `Square`: this makes a big difference when polling; you can try setting a `pollInterval` to observe how hover is noticeably interrupted if the square is unmemoized
  - Use CSS tooltip to avoid delay of LG tooltip
  - Along with CSS grid, Safari was greatly slowed down by `PageWrapper`'s approach of setting a div to the viewport height and leveraging `overflow-y: scroll`. To overcome this I configured the waterfall page to use a content-height div and set the navbar to use `position: fixed`
- Time to first paint is a bit slow right now but we'll focus on optimizing this in the future!
- Task status will be correctly rendered in DEVPROD-11368
- Upgrade DB from 7.0.2 to 7.0.14 to fix a crazy database bug that caused the Waterfall query to return incorrect data 😵 I opted not to upgrade to 8.0 because that caused failures; instead opened DEVPROD-11699

### Screenshots

image

### Testing

- Add Cypress tests


apps/spruce/src/pages/waterfall/BuildRow.tsx
```diff
@@ -0,0 +1,161 @@
...
+    <Row>
+      <BuildVariantTitle>
+        <StyledLink
+          href={getVariantHistoryRoute(projectIdentifier, build.id)}
+          onClick={handleVariantClick}
+        >
+          {displayName}
+        </StyledLink>
+      </BuildVariantTitle>
+      <BuildGroup data-cy="build-group">
```
minnakt, line 55: looks like there are some missing key errors:

Screenshot 2024-09-27 at 4 24 01 PM

Screenshot 2024-09-27 at 4 25 00 PM

apps/spruce/src/pages/waterfall/BuildRow.tsx
```diff
@@ -0,0 +1,161 @@
...
+        {versions.map(({ inactiveVersions, version }) => {
+          if (inactiveVersions) {
+            return <InactiveVersion data-cy="inactive-column" />;
+          }
+          // The list of builds returned does not include a placeholder for inactive builds, so we need to check whether the build matches the version in the current column
+          if (version && version.id === builds?.[buildIndex]?.version) {
```
minnakt, line 61: is the reason this works because the `builds` is guaranteed to be the same order as the `versions`? It could be good to leave a comment about that

apps/spruce/src/pages/waterfall/BuildRow.tsx
```diff
@@ -0,0 +1,161 @@
...
+  handleTaskClick: (s: string) => () => void;
+}> = ({ build, handleTaskClick }) => (
+  <Build>
+    {build.tasks.map(({ displayName, id, status }) => (
+      <SquareMemo
+        key={id}
+        data-tooltip={displayName}
+        onClick={handleTaskClick(status)}
+        status={status}
+        to={getTaskRoute(id)}
```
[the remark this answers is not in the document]
sophstad: Good callout, created DEVPROD-11734.

apps/spruce/src/pages/waterfall/index.tsx
```diff
@@ -1,13 +1,39 @@
...
+        }
+        nav {
+          position: fixed !important;
+          width: 100%;
+          z-index: 1;
+        }
+      `}
+    />
+    <PageContainer data-cy="waterfall-page">
+      <Suspense fallback={<h3>Loading waterfall...</h3>}>
```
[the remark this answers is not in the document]
sophstad: 😎 
image

apps/spruce/src/pages/waterfall/WaterfallGrid.tsx
```diff
@@ -28,23 +35,36 @@ export const WaterfallGrid: React.FC = () => {
...
-          version ? <VersionLabel key={version.id} {...version} /> : null,
-        )}
+        <BuildVariantTitle />
+        <Versions data-cy="version-labels">
+          {data.waterfall.versions.map(({ inactiveVersions, version }, i) =>
+            version ? (
+              <VersionLabel key={version.id} {...version} />
+            ) : (
+              <InactiveVersion
+                key={`${inactiveVersions?.[0]?.id}-${i}`} // eslint-disable-line react/no-array-index-key
```
minnakt, line 45: would just the ID be enough? (assuming inactive versions don't contain duplicate versions)
```suggestion
                key={inactiveVersions?.[0]?.id}
```

apps/spruce/src/pages/waterfall/BuildRow.tsx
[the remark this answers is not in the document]
minnakt: yay, can we add a TODO here with the ticket number 😁

apps/spruce/src/pages/waterfall/VersionLabel.tsx
```diff
@@ -90,10 +90,16 @@ export const VersionLabel: React.FC<VersionFields> = ({
 };
 
 const VersionContainer = styled.div`
+  flex-basis: 20%;
```
minnakt, line 93: would we want to save this `flex-basis` in a shared variable?

apps/spruce/src/pages/waterfall/BuildRow.tsx
```diff
@@ -0,0 +1,167 @@
...
+const BuildGroup = styled.div`
+  ${gridGroupCss}
+  border: 1px solid ${gray.light2};
+  border-radius: ${size.xs};
+  padding-bottom: ${size.xs};
+  padding-top: ${size.xs};
+`;
+
+const Build = styled.div`
+  flex-basis: 20%;
```
minnakt, line 111: (to share with here)

minnakt (approved):
the tooltip makes me sooo happy ☺️

apps/spruce/src/pages/waterfall/WaterfallGrid.tsx
sophstad: I'll use it just as a fallback!

apps/spruce/src/pages/waterfall/BuildRow.tsx
```diff
@@ -0,0 +1,168 @@
...
+const BuildGrid: React.FC<{
+  build: WaterfallBuild;
+  handleTaskClick: (s: string) => () => void;
+}> = ({ build, handleTaskClick }) => (
+  <Build>
+    {build.tasks.map(({ displayName, id, status }) => (
+      <SquareMemo
+        key={id}
+        data-tooltip={displayName}
+        onClick={handleTaskClick(status)}
```
SupaJoon, line 95: You can decrease the number of event handlers by adding the onClick to `Build`. The callback would look like this:
```
(event) => {
    handleTaskClick(event.target?.getAttribute('status'))
}
```
I can't really notice a difference in render speed but this technically would reduce memory usage.

SupaJoon (approved):

PR #413, opened 2024-09-26
DEVPROD-11365: Create OS table on Build Information page

minnakt:
DEVPROD-11365

### Description
Adds the OS table and fixes a couple visual and functional bugs here and there. The page will become easier to navigate after the completion of [DEVPROD-11363](https://jira.mongodb.org/browse/DEVPROD-11363).

### Screenshots

Screenshot 2024-09-26 at 3 36 00 PM

### Testing
- Unit tests & Cypress tests


apps/spruce/src/pages/image/OperatingSystemTable/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  const numTotalItems = useMemo(
+    () =>
+      osData?.image?.operatingSystem.filteredCount ??
+      osData?.image?.operatingSystem.totalCount,
+    [
+      osData?.image?.operatingSystem.filteredCount,
+      osData?.image?.operatingSystem.totalCount,
+    ],
+  );
```
sophstad, line 66: If I'm not mistaken it seems like these `useMemo`s have no impact. Basically they're saying "if the data value changes, update the variable value" which feels... tautological

sophstad (changes requested):

apps/spruce/src/pages/image/OperatingSystemTable/index.tsx
minnakt: oh yeah I'm not sure why we started doing that... I removed them all around here

apps/spruce/src/pages/image/OperatingSystemTable/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+      search: {
+        "data-cy": "os-name-filter",
+        placeholder: "Name regex",
+      },
+    },
+  },
+  {
+    header: "Version",
+    accessorKey: "version",
+    cell: ({ getValue }) => (getValue() as string).replace(/"/g, ""),
```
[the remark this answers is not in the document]
minnakt: the linux builds will probably have 1 or 2 links, I haven't seen any for mac/windows builds though

apps/spruce/src/pages/image/OperatingSystemTable/index.tsx
sophstad: Ok I kinda feel like I had Friday brain when I made this comment since the useMemo would still do *something*, but I think removing them is fine too for simplicity's sake since I don't anticipate this page having performance issues.

apps/spruce/src/utils/string/Linkify.tsx
```diff
@@ -0,0 +1,40 @@
...
+      key={`${match}${i}`}
+      data-cy="jira-link"
+      href={getJiraTicketUrl(jiraHost, match)}
+      onClick={onClick}
+    >
+      {match}
+    </StyledLink>
+  ));
+
+const linkifyString = (unlinkified: string) =>
```
sophstad, line 33: It looks like we have a [dependency `linkify-html`](https://github.com/evergreen-ci/ui/blob/7b76bf5af3a83e09c1a77fead48dc945e7abcc58/apps/spruce/src/pages/task/taskTabs/logs/logTypes/LogMessageLine.tsx#L30) that does this.

I wonder if the dependency has more security handling, or if we should just drop it?

minnakt: I think we should just drop it tbh, you have to call the `parse` function on top of `linkify-html` to parse it into real html and I think it's not worth it

I'll keep the other changes for testing the preexisting linkify functions though since we probably should have had tests for those!

sophstad (approved):

PR #417, opened 2024-09-30
DEVPROD-11710: Update Evergreen logo for fall

minnakt:
DEVPROD-11710

### Description
Updates Evergreen logo. The logo is animated on hover. We could also leave the icon animated all the time, but I'm not sure if that's too distracting.

### Screenshots

https://github.com/user-attachments/assets/21c6f40f-ef87-4b42-90e5-2a008d981183


apps/spruce/src/components/Icon/icons/FallTree.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+  const onMouseEnter = () => {
+    const fallTree = document.getElementsByTagName("svg");
+    fallTree[0].unpauseAnimations();
+  };
+
+  const onMouseLeave = () => {
+    const fallTree = document.getElementsByTagName("svg");
+    fallTree[0].pauseAnimations();
+  };
```
khelif96, line 18: I think animating all the time could be cool but up to you!

apps/spruce/src/components/Icon/icons/FallTree.tsx
```diff
@@ -0,0 +1,127 @@
+import { useEffect } from "react";
+
+export const FallTree: React.FC = () => {
+  // Animations should be paused by default on page load.
+  useEffect(() => {
+    const fallTree = document.getElementsByTagName("svg");
+    fallTree[0].pauseAnimations();
```
khelif96, line 7: If we keep this approach can we target the svg using a ref or an id? This would match all svgs on the page.

khelif96 (approved):

PR #426, opened 2024-10-02
DEVPROD-10196: Render Inactive Commits Button and Modal in Waterfall

SupaJoon:
DEVPROD-10196
### Description
These code changes also contain the result of `yarn run eslint:fix` which ran on files that did not originally contain diffs for this ticket.
Screenshot 2024-10-02 at 2 31 03 PM
Screenshot 2024-10-02 at 2 31 08 PM

### DIFF
--- a/apps/spruce/src/pages/waterfall/InactiveVersionsButton/index.tsx
+++ b/apps/spruce/src/pages/waterfall/InactiveVersionsButton/index.tsx
@@ -1,0 +1,65 @@
+import { useState } from "react";
+import styled from "@emotion/styled";
+import Button from "@leafygreen-ui/button";
+import { palette } from "@leafygreen-ui/palette";
+import pluralize from "pluralize";
+import { DisplayModal } from "components/DisplayModal";
+import Icon from "components/Icon";
+import { size } from "constants/tokens";
+import { WaterfallQuery } from "gql/generated/types";
+import { VersionLabel } from "../VersionLabel";
+
+const { gray } = palette;
+interface Props {
+  versions: WaterfallQuery["waterfall"]["versions"][0]["inactiveVersions"];
+  containerHeight: number | undefined;
+}
+export const InactiveVersionsButton: React.FC<Props> = ({
+  containerHeight,
+  versions,
+}) => {
+  const [modalOpen, setModalOpen] = useState(false);
+  return (
+    <>
+      <DisplayModal
+        data-cy="inactive-versions-modal"
+        open={modalOpen}
+        setOpen={setModalOpen}
+        title={`${versions?.length} Inactive ${pluralize("Version", versions?.length)}`}
+      >
+        {versions?.map((version) => (
+          <StyledVersionLabel
+            key={version.id}
+            commitType="inactive"
+            {...version}
+          />
+        ))}
+      </DisplayModal>
+      <Button
+        aria-label="Open inactive versions modal"
+        data-cy="inactive-versions-button"
+        onClick={() => {
+          setModalOpen(true);
+        }}
+        size="xsmall"
+      >
+        <Icon fill={gray.base} glyph="List" />
+        <span>{versions?.length}</span>
+        <InactiveVersionLine containerHeight={containerHeight ?? 0} />
+      </Button>
+    </>
+  );
+};
+
+const InactiveVersionLine = styled.div<{ containerHeight: number }>`
+  border-left: 2px dashed ${gray.base};
+  height: ${({ containerHeight }) => `${containerHeight}px`};
+  position: absolute;
+  margin-left: 50%;
+  top: 30px;
+  z-index: 1;
+`;
+
+const StyledVersionLabel = styled(VersionLabel)`
+  padding-top: ${size.xs};
+`;
--- a/apps/spruce/src/pages/waterfall/InactiveVersionsButton/index.tsx
+++ b/apps/spruce/src/pages/waterfall/InactiveVersionsButton/index.tsx
@@ -30,7 +30,7 @@ export const InactiveVersionsButton: React.FC<Props> = ({
         {versions?.map((version) => (
           <StyledVersionLabel
             key={version.id}
-            commitType="inactive"
+            trimMessage={false}
             {...version}
           />
         ))}
--- a/apps/spruce/src/pages/waterfall/VersionLabel/InactiveBadge.tsx
+++ b/apps/spruce/src/pages/waterfall/VersionLabel/InactiveBadge.tsx
@@ -1,0 +1,11 @@
+import styled from "@emotion/styled";
+import Badge, { Variant } from "@leafygreen-ui/badge";
+import { size } from "constants/tokens";
+
+export const InactiveBadge = () => (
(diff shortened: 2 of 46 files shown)


apps/spruce/src/pages/waterfall/InactiveVersionsButton/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+  versions,
+}) => {
+  const [modalOpen, setModalOpen] = useState(false);
+  return (
+    <>
+      <DisplayModal
+        data-cy="inactive-versions-modal"
+        open={modalOpen}
+        setOpen={setModalOpen}
+        title={`${versions?.length} Inactive ${pluralize("Version", versions?.length)}`}
```
SupaJoon, line 28: This text may change when DEVPROD-11706 is implemented. "inactive-versions" may change to "inactive-and-filtered" depending on how that ticket is implemented.

sophstad (changes requested):
I think the ESLint config you're using locally is out of date, since lint is passing on main and these changes are undone when I run it locally.

Could you pull in main and remove these changes to simplify reviewing? 🫶

apps/spruce/src/pages/waterfall/InactiveVersionsButton/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+      >
+        {versions?.map((version) => (
+          <StyledVersionLabel
```
sophstad, line 31: Cool to reuse the VersionLabel component 😎 for this instance, I think we should remove the text trim present in that component in case of long commits. The `title` attribute is also redundant.

We could also opt to use our regular font size instead of the small version, whatever you think.

apps/spruce/src/pages/waterfall/InactiveVersionsButton/index.tsx
sophstad, line 1: Could you add a story for this?

apps/spruce/src/pages/waterfall/InactiveVersionsButton/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+        <span>{versions?.length}</span>
+        <InactiveVersionLine containerHeight={containerHeight ?? 0} />
+      </Button>
+    </>
+  );
+};
+
+const InactiveVersionLine = styled.div<{ containerHeight: number }>`
+  border-left: 2px dashed ${gray.base};
+  height: ${({ containerHeight }) => `${containerHeight}px`};
```
sophstad, line 56: This looks perfect 😍

sophstad (changes requested):

apps/spruce/src/pages/waterfall/InactiveVersionsButton/index.tsx
SupaJoon: I removed the title property when the message is full length

apps/spruce/src/pages/waterfall/InactiveVersionsButton/index.tsx
SupaJoon: I rendered the badge in this VersionLabel [story](http://localhost:6006/?path=/docs/pages-waterfall-versionlabel--docs#inactive%20untrimmed%20message) ([code](https://github.com/evergreen-ci/ui/pull/426/files#diff-8b193d898596d3782191d22721dc6f973b60492615d66fb749998cf0d71a411eR107))

apps/spruce/src/pages/waterfall/InactiveVersionsButton/index.tsx
SupaJoon: also made a size prop

apps/spruce/src/pages/waterfall/VersionLabel/InactiveBadge.tsx
```diff
@@ -0,0 +1,11 @@
+import styled from "@emotion/styled";
+import Badge, { Variant } from "@leafygreen-ui/badge";
+import { size } from "constants/tokens";
+
+export const InactiveBadge = () => (
+  <StyledBadge variant={Variant.LightGray}>Inactive</StyledBadge>
```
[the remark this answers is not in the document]
SupaJoon: I left it separate

apps/spruce/src/pages/waterfall/VersionLabel/index.tsx
```diff
@@ -89,22 +106,28 @@ export const VersionLabel: React.FC<VersionFields> = ({
...
   }
 `;
 
-const CommitMessage = styled(Body)`
+const CommitMessage = styled(Body)<{ trimMessage: boolean }>`
   display: -webkit-box;
   -webkit-box-orient: vertical;
   -webkit-line-clamp: 3;
-  overflow: hidden;
+  ${(props) => props.trimMessage && "overflow: hidden;"}
```
sophstad, line 132: I think you can potentially make all of these properties conditional on `trimMessage`; right now the overflow is incorrect
image

apps/spruce/src/pages/waterfall/VersionLabel/index.tsx
```diff
@@ -31,15 +42,17 @@ export const VersionLabel: React.FC<VersionFields> = ({
 
   const { sendEvent } = useWaterfallAnalytics();
 
+  const commitType = activated ? "active" : "inactive";
```
sophstad, line 45: It would probably be cleaner to case directly on `activated` rather than introduce a magic string

apps/spruce/src/pages/waterfall/WaterfallGrid.tsx
```diff
@@ -30,21 +33,25 @@ export const WaterfallGrid: React.FC = () => {
...
-              <VersionLabel key={version.id} {...version} />
+              <VersionLabel key={version.id} size="small" {...version} />
             ) : (
-              <InactiveVersion
-                key={inactiveVersions?.[0]?.id ?? i} // eslint-disable-line react/no-array-index-key
-                data-cy="inactive-label"
-              >
-                inactive
+              <InactiveVersion>
+                <InactiveVersionsButton
```
sophstad, line 50: Would it be possible to center the line and button in the column?

sophstad (changes requested):

apps/spruce/src/pages/waterfall/VersionLabel/index.tsx
```diff
@@ -93,13 +93,24 @@ export const VersionLabel: React.FC<Props> = ({
...
-        })}
+        {jiraLinkify(
+          message +
+            message +
+            message +
+            message +
+            message +
+            message +
+            message +
+            message,
```
sophstad, line 104: Hmm I don't think this is right
image

apps/spruce/src/pages/waterfall/VersionLabel/index.tsx
sophstad: Hmm good point, I'm okay with either approach. Your call!

sophstad (changes requested):

SupaJoon:
evergreen retry

sophstad (approved):

PR #428, opened 2024-10-04
DEVPROD-11145: Upgrade @leafygreen-ui/combobox to v9.1.4

minnakt:
DEVPROD-11145

### Description
The ComboBox is broken due to weird interactions with `@leafygreen-ui/hooks`.

You can use this [CodeSandbox](https://g7rmv7.csb.app/) to verify that it doesn't work for any version after `8.1.3`.


package.json
```diff
@@ -20,7 +20,8 @@
     "@testing-library/dom": "10.1.0",
     "@types/react": "18.2.0",
     "@types/react-dom": "18.2.0",
-    "jackspeak": "2.1.1"
+    "jackspeak": "2.1.1",
+    "@leafygreen-ui/hooks": "8.1.3"
```
sophstad, line 24: Did you try only forcing this version for combobox? I believe this would be the syntax:
```json
"@leafygreen-ui/combobox/@leafygreen-ui/hooks"
```

It would be preferable to not apply this to all LG components I think if it works

khelif96, line 24: I would prefer if we get LG to fix this on their end instead of pinning the dependency since we would likely forget about this or run into unpredictable behavior if LG updates this in the future.

khelif96 (changes requested):

package.json
minnakt: Oh yes, I could only get this working by moving the content to Spruce `package.json`, but I assume that's fine

minnakt: I think we can just ticket this and track it on our side. It's been broken since August 27th, I think we should patch it

khelif96 (approved):

sophstad (approved):

khelif96:
Commenting here just in case you missed the thread while you are out. The [fix](https://github.com/mongodb/leafygreen-ui/pull/2498) for this should be released as of today or tomorrow.  So we should be able to just update Combobox.

PR #429, opened 2024-10-04
DEVPROD-11368: Render task status on waterfall

sophstad:
DEVPROD-11368

### Description

- Render correct status color
  - For builds where activated = false, use the light gray inactive color like the legacy waterfall does
- Add status icons using [Data URIs](https://css-tricks.com/data-uris/)
- Add status text to tooltip

### Screenshots

image

### Testing
Storybook entry will be added with the legend in DEVPROD-10198


apps/spruce/src/pages/waterfall/BuildRow.tsx
```diff
@@ -128,12 +135,12 @@ const Square = styled(Link)<{ status: string }>`
...
   position: relative;
 
-  /* TODO DEVPROD-11368: Render colors for all statuses. Could use background-image property to render icons. */
-  ${({ status }) =>
-    status === TaskStatus.Succeeded
-      ? `background-color: ${green.dark1};`
-      : `background-color: ${gray.light2};
-  `}
+  ${({ status }) => {
+    const icon = statusIconMap?.[status];
```
SupaJoon, line 139:
```suggestion
    const icon = statusIconMap[status];
```

apps/spruce/src/pages/waterfall/styles.ts
```diff
@@ -38,3 +49,36 @@ export const InactiveVersion = styled.div`
   width: ${INACTIVE_WIDTH}px;
   text-align: center;
 `;
+
+export const statusColorMap: Record<string, string> = {
```
SupaJoon, line 53: super nit: statusColorMap and statusIconMap aren't actually styles. Maybe they should go in icons.ts?

SupaJoon (approved):

apps/spruce/src/pages/waterfall/BuildRow.tsx
sophstad: These are functionally the same since accessing a property that doesn't exist in an object yields `undefined` (no error is thrown)

SupaJoon: That's right. `icon` may evaluate to undefined but since `statusIconMap` is never undefined, the `?.` is unnecessary. I suggested to make usage with statusColorMap since they are similar.

PR #430, opened 2024-10-07
DEVPROD-1770: Use modern query param hooks with tables

SupaJoon:
DEVPROD-1770

### Description
Replace `useUpdateURLQueryParams` with `useQueryParams` or `useQueryParam` in table components. I also updated ExectionTaskTable so rely on sorts instead of sortBy and sortDir because it made these code changes simpler.

### DIFF
--- a/apps/spruce/src/pages/task/taskTabs/ExecutionTasksTable.tsx
+++ b/apps/spruce/src/pages/task/taskTabs/ExecutionTasksTable.tsx
@@ -18,8 +18,7 @@ import {
   SortDirection,
 } from "gql/generated/types";
 import { useTableSort } from "hooks";
-import { useQueryParams } from "hooks/useQueryParam";
-import { useUpdateURLQueryParams } from "hooks/useUpdateURLQueryParams";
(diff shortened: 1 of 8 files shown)


apps/spruce/src/pages/task/taskTabs/ExecutionTasksTable.tsx
```diff
@@ -117,42 +108,27 @@ export const ExecutionTasksTable: React.FC<Props> = ({
...
+      }).map(({ direction, id }) => ({
+        id,
+        desc: direction === SortDirection.Desc,
+      }))
+    : [{ id: TaskSortCategory.Status, desc: false }];
 
   return initialSorting;
 };
+
+const defaultSortQueryParam = `${TaskSortCategory.Status}:${SortDirection.Asc}`;
```
khelif96, line 134: Can we construct this with toSortString to ensure its consistent with that logic?

khelif96 (approved):

SupaJoon:
The test marked as failed actually succeeded.

PR #436, opened 2024-10-07
DEVPROD-11900: Add alpha info banner to beta deploys

sophstad:
DEVPROD-11900

### Description

Add banner directing users to give feedback on the waterfall

I made the banner non-dismissible because I think having it visible will help reinforce that this page is a work in progress

### Screenshots

image

### DIFF
--- a/apps/spruce/src/pages/waterfall/index.tsx
+++ b/apps/spruce/src/pages/waterfall/index.tsx
@@ -1,38 +1,35 @@
 import { Suspense } from "react";
 import { Global, css } from "@emotion/react";
 import styled from "@emotion/styled";
+import Banner from "@leafygreen-ui/banner";
 import { TableSkeleton } from "@leafygreen-ui/skeleton-loader";
 import { useParams } from "react-router-dom";
 import { navBarHeight } from "components/Header/Navbar";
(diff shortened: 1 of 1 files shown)


apps/spruce/src/pages/waterfall/index.tsx
```diff
@@ -52,4 +49,20 @@ const PageContainer = styled.div`
   padding: ${size.m} ${size.l};
 `;
 
+/* Safari performance of the waterfall chokes if using overflow-y: scroll, so we need the page to scroll instead.
```
khelif96, line 52: This is such a wild quirk

khelif96: I can't tell by the below comment but is this override going to be permanent or is the plan to eventually update the Navbar component?

khelif96 (approved):

apps/spruce/src/pages/waterfall/index.tsx
sophstad: Permanent, unless a Safari fix is introduced 🙃 pages with the LeafyGreen sidebar in particular need to utilize overflow-y

PR #438, opened 2024-10-07
DEVPROD-10197: Implement broken versions warning on waterfall

SupaJoon:
DEVPROD-10197

### Description
Shows a badge on the waterfall and inactive/unmatching versions modal.

### Screenshots
Screenshot 2024-10-07 at 5 03 18 PM
Screenshot 2024-10-07 at 5 03 27 PM

### Evergreen PR
https://github.com/evergreen-ci/evergreen/pull/8374


apps/spruce/src/pages/waterfall/WaterfallGrid.tsx
```diff
@@ -46,7 +46,12 @@ export const WaterfallGrid: React.FC<WaterfallGridProps> = ({
             version ? (
               <VersionLabel key={version.id} size="small" {...version} />
             ) : (
```
sophstad, line 48: If you calculate the inactive version count here, then you can pass the value into each prop (either the count or `hasError`) instead of having to both count and use the `.some()` function in two separate places

sophstad (changes requested):

SupaJoon:
evergreen retry

apps/spruce/src/pages/waterfall/styles.ts
```diff
@@ -34,7 +35,8 @@ export const Row = styled.div`
   margin-bottom: ${size.s};
 `;
 
-export const InactiveVersion = styled.div`
-  width: ${INACTIVE_WIDTH}px;
+export const InactiveVersion = styled.div<{ hasError: boolean }>`
+  width: ${({ hasError }) =>
+    hasError ? INACTIVE_WITH_ERROR_WIDTH : INACTIVE_WITHOUT_ERROR_WIDTH}px;
```
sophstad, line 40: As-is on my laptop this element shrinks to 132px
image


I believe we can prevent shrinkage like so. And even 150px is enough vs 160.
```suggestion
  flex-shrink: 0;
  flex-basis: ${({ hasError }) =>
    hasError ? INACTIVE_WITH_ERROR_WIDTH : INACTIVE_WITHOUT_ERROR_WIDTH}px;
```

apps/spruce/src/pages/waterfall/styles.ts
```diff
@@ -4,7 +4,8 @@ import { wordBreakCss } from "components/styles";
 import { size } from "constants/tokens";
 
 const BUILD_VARIANT_WIDTH = 200;
-const INACTIVE_WIDTH = 80;
+const INACTIVE_WITHOUT_ERROR_WIDTH = 80;
+const INACTIVE_WITH_ERROR_WIDTH = 160;
```
sophstad, line 8: As mentioned above
```suggestion
const INACTIVE_WITH_ERROR_WIDTH = 150;
```

apps/spruce/src/pages/waterfall/InactiveVersionsButton/index.tsx
```diff
@@ -63,3 +75,8 @@ const InactiveVersionLine = styled.div<{ containerHeight: number }>`
 const StyledVersionLabel = styled(VersionLabel)`
   padding-top: ${size.xs};
 `;
+
+const StyledBadge = styled(Badge)`
+  top: -${size.xs};
```
sophstad, line 80: [nit] I know the designs have this slightly above the rest of the elements but I think positioning all the filters will be easier if it's inline with the top of VersionLabel with some padding between it and the button

sophstad (changes requested):

apps/spruce/src/pages/waterfall/InactiveVersionsButton/index.tsx
```diff
@@ -35,6 +41,12 @@ export const InactiveVersionsButton: React.FC<Props> = ({
           />
         ))}
       </DisplayModal>
+      {brokenVersionsCount > 0 && (
+        <StyledBadge data-cy="broken-versions-badge" variant={Variant.Red}>
+          {brokenVersionsCount} broken{" "}
+          {pluralize("version", brokenVersionsCount)}
```
sophstad, line 47: Sorry for not suggesting this earlier, but what do you think of omitting "versions" entirely? I don't think it's so helpful that it warrants taking away space from the useful columns on the page. Without it we could just have one fixed width for the `InactiveVerison` div I think.

sophstad (approved):

PR #439, opened 2024-10-07
DEVPROD-9478 Refactor date selection logic in Spawn Host page test

khelif96:
DEVPROD-9478

### Description
I noticed that these tests broke again because .next() would be undefined. I refactored this test to not rely on `.next()` and to instead check if the cell has a sibling. If it does then we can select it, otherwise toggle the next month logic.

### DIFF
--- a/apps/spruce/cypress/integration/spawn/host.ts
+++ b/apps/spruce/cypress/integration/spawn/host.ts
@@ -251,23 +251,34 @@ describe("Navigating to Spawn Host page", () => {
 
     cy.getInputByLabel("Temporary Sleep Schedule Exemption").click();
     cy.get("td[aria-current=true]").as("currentDateCell");
+
     // Select a date in the future either this month or next month
     // if the current date is the last day of the month select the first day of the next month
     // if it is not select the next day
     // We can determine if the current date is the last day of the month if the next cell is disabled
-    cy.get("@currentDateCell")
-      .next()
-      .then(($el) => {
-        if (!$el || $el.attr("aria-disabled") === "false") {
-          cy.get("@currentDateCell").next().click();
-        } else {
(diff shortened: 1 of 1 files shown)


apps/spruce/cypress/integration/spawn/host.ts
```diff
@@ -251,23 +251,34 @@ describe("Navigating to Spawn Host page", () => {
...
-          cy.get("@currentDateCell").next().click();
-        } else {
-          cy.log("Current date is the last day of the month");
-          cy.get('button[aria-label="Next month"]').click();
-          cy.get('td[data-testid="lg-date_picker-calendar_cell"]')
-            .first()
-            .click();
-        }
-      });
+    cy.get("@currentDateCell").then(($currentDateCell) => {
```
SupaJoon, line 259: You might be able to simplify this logic and make the code path more deterministic by setting the browser date in cypress with [cy.clock](https://docs.cypress.io/api/commands/clock)

khelif96: I tried this out using `cy.clock()` and it had the unintended side effect of messing with how timers work on the page. Using it prevented the edit modal from even opening because it uses a timer to trigger opening the modal. I think this would introduce too much complexity to the page if we needed to handle managing all of the timers the page uses under the hood.

SupaJoon: Thanks for checking it out!

SupaJoon (approved):

apps/spruce/cypress/integration/spawn/host.ts
khelif96: Great suggestion tho! I didn't know about this API. I think it's more geared towards component testing

PR #441, opened 2024-10-08
DEVPROD-11911 Add pause back to the spruce logo animation.

khelif96:
DEVPROD-11911

### DIFF
--- a/apps/spruce/src/components/Icon/icons/FallTree.tsx
+++ b/apps/spruce/src/components/Icon/icons/FallTree.tsx
@@ -1,126 +1,105 @@
-import { useEffect, useRef } from "react";
+import { forwardRef } from "react";
 
-export const FallTree: React.FC = () => {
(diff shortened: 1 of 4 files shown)


apps/spruce/src/components/Icon/icons/FallTree.tsx
```diff
@@ -1,105 +1,126 @@
...
-            values="0.582776;0.838822;0.900502;0.900502;0.900502;0.900502;0.900502;0.900502;0.900502;0.838822;0.582776"
-          />
-        </stop>
-      </linearGradient>
-    </defs>
-  </svg>
-);
+import { useEffect, useRef } from "react";
+
+export const FallTree: React.FC = () => {
```
sophstad, line 3: Might we extract this into a reusable "pause icon" component that imports the fall tree? This seems like a pattern we ought to keep around

SupaJoon (approved):

PR #443, opened 2024-10-11
DEVPROD-10191: Add waterfall build variant filter

SupaJoon:
DEVPROD-10191

### Description
These code changes add client side filtering for build variants. I updated the analytics document with the 3 new events.

Screenshot 2024-10-11 at 5 49 51 PM


apps/spruce/src/pages/waterfall/WaterfallFilters.tsx
```diff
@@ -32,7 +36,7 @@ export const WaterfallFilters: React.FC<WaterfallFiltersProps> = ({
 
 // Temporary - update styles as more filters are added.
 const FilterItem = styled.div`
-  width: 300px;
+  ${({ width }: { width: number }) => `width: ${width}%;`}
 `;
 
 const Container = styled.div`
```
sophstad, line 42: I think using `align-items: flex-end;` will have the most consistent look here

apps/spruce/src/pages/waterfall/index.tsx
```diff
@@ -31,6 +42,19 @@ const Waterfall: React.FC = () => {
...
+          <FilterBadges
+            badges={badges}
+            onClearAll={() => {
+              sendEvent({ name: "Deleted all filter badges" });
+              handleClearAll();
+            }}
+            onRemove={(b) => {
+              sendEvent({ name: "Deleted one filter badge" });
+              handleOnRemove(b);
+            }}
```
sophstad, line 55: Since we're using Suspense on this page, these handlers represent a perfect use case for [`useTransition`](https://react.dev/reference/react/useTransition). Notice how after "clear all" is clicked on a particularly heavy page, the other page elements become unusable. Adding `startTransition` to these handlers fixes that issue so the page is responsive in the middle of a rerender.

apps/spruce/src/pages/waterfall/index.tsx
```diff
@@ -31,6 +42,19 @@ const Waterfall: React.FC = () => {
           </Banner>
         )}
         <WaterfallFilters projectIdentifier={projectIdentifier ?? ""} />
+        <BadgeWrapper>
```
sophstad, line 45: Could we maybe update the spacing of this component so that the waterfall shifts as little as possible when a filter is added? Maybe with placeholder padding or something.

apps/spruce/src/pages/waterfall/WaterfallFilters.tsx
```diff
@@ -32,7 +36,7 @@ export const WaterfallFilters: React.FC<WaterfallFiltersProps> = ({
 
 // Temporary - update styles as more filters are added.
 const FilterItem = styled.div`
-  width: 300px;
+  ${({ width }: { width: number }) => `width: ${width}%;`}
```
sophstad, line 39: I think `flex-basis` would be the preferred property here
```suggestion
  ${({ width }: { width: number }) => `flex-basis: ${width}%;`}
```

sophstad (changes requested):

sophstad:
Just wanted to flag #451, I think it may be helpful to standardize the filtering approach here to to also use the `useFilters` hook!

apps/spruce/src/pages/waterfall/index.tsx
SupaJoon: I think it's better use of space and looks cleaner without padding. 1 row is about 40px which is a noticeable gap when no filters are applied.

apps/spruce/src/pages/waterfall/NameFilter.tsx
```diff
@@ -0,0 +1,41 @@
...
+          sendEvent({ name: "Created build variant filter" });
+          break;
+        default:
+      }
+    },
+    [onSubmit, sendEvent],
+  );
+
+  return (
+    <TupleSelectWithRegexConditional
```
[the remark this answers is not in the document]
sophstad: I'm still seeing this
image

apps/spruce/src/components/FilterBadges/FilterBadge.tsx
```diff
@@ -44,47 +40,30 @@ const FilterBadge: React.FC<FilterBadgeProps> = ({ badge, onClose }) => {
...
-const ClickableIcon = styled(Icon)`
-  position: absolute;
-  right: ${size.xxs};
-  :hover {
-    cursor: pointer;
-    color: ${gray.light1};
-  }
+const ChipLabel = styled.span`
+  text-transform: uppercase;
+  font-weight: 700;
```
sophstad, line 61: Can we remove this styling from the chip? Specifically I think the text-transform has always been confusing because it seems like a different search is applied than what the user has input

apps/spruce/src/pages/waterfall/index.tsx
sophstad: Gotcha, could we try removing some of this padding at least? It's a pretty huge amount of whitespace given that users' priority for this page is information density
image

I almost feel like the default amount of whitespace is enough to fit the filters:
image

apps/spruce/src/pages/waterfall/useFilters.ts
```diff
@@ -64,14 +72,37 @@ export const useFilters = (waterfall: WaterfallQuery["waterfall"]) => {
...
 
   const buildVariants = useMemo(() => {
     if (!hasFilters) {
       return waterfall.buildVariants;
     }
 
     const bvs: WaterfallBuildVariant[] = [];
-    waterfall.buildVariants.forEach((bv) => {
+    const waterfallBuildVariants = buildVariantFilterRegex.length
+      ? waterfall.buildVariants.filter(({ displayName }) => {
```
sophstad, line 96: Could we apply this filter within the forEach? As we build out filtering I think it would be preferable to minimize the number of mapping operations

apps/spruce/src/pages/waterfall/useFilters.ts
```diff
@@ -14,7 +14,15 @@ export const useFilters = (waterfall: WaterfallQuery["waterfall"]) => {
...
 
-  const hasFilters = useMemo(() => requesters.length, [requesters]);
+  const [buildVariantFilter] = useQueryParam<string[]>(
+    WaterfallFilterOptions.BuildVariant,
+    [],
+  );
+
+  const hasFilters = useMemo(
+    () => requesters.length || buildVariantFilter.length,
+    [requesters],
```
sophstad, line 24:
```suggestion
    [buildVariantFilter, requesters],
```

apps/spruce/src/pages/waterfall/NameFilter.tsx
```diff
@@ -0,0 +1,41 @@
...
+      onSubmit={onSubmitTupleSelect}
```
sophstad, line 27: Remind me how exact vs. regex is distinguished in the query params again? I haven't been able to observe a difference in behavior

apps/spruce/src/components/FilterBadges/FilterBadge.tsx
SupaJoon: I create this ticket: https://jira.mongodb.org/browse/DEVPROD-12590

apps/spruce/src/pages/waterfall/NameFilter.tsx
SupaJoon: If exact is selected, the string will be escaped in the query param.

sophstad (approved):
😎

PR #451, opened 2024-10-16
DEVPROD-10193: Add waterfall requester filters

sophstad:
DEVPROD-10193

### Description

- Add requester filters, which then omits versions that don't match the filter as well as build variants that have no visible versions
- Add `useFilters` hook which will be extended to handle all client side filtering
- Leverage [`useTransition`](https://react.dev/reference/react/useTransition) for performant state updates
- This PR also causes changes to the GitHub app project settings page because I changed the requester titles to their more commonly-used named. Let me know if you think we shouldn't make this standardization!

### Screenshots

image

### Testing

- Add Cypress testing


apps/spruce/src/constants/requesters.ts
```diff
@@ -11,18 +11,16 @@ enum Requester {
...
+  [Requester.AdHoc]: "Periodic build",
   [Requester.GitHubMergeQueue]: "GitHub Merge Request",
   [Requester.GitHubPR]: "GitHub Pull Request",
-  [Requester.GitTag]: "Git Tag Request",
-  [Requester.Gitter]: "Gitter Request",
-  [Requester.Patch]: "Patch Request",
-  [Requester.Trigger]: "Trigger Request",
+  [Requester.GitTag]: "Git tag",
+  [Requester.Gitter]: "Commit",
+  [Requester.Patch]: "Patch",
```
minnakt, line 19: nit. Could we use a consistent case here? (choose one between sentence case and title case)

apps/spruce/src/pages/waterfall/RequesterFilter.tsx
```diff
@@ -0,0 +1,50 @@
...
+  Requester.Trigger,
+];
+
+export const RequesterFilter = () => {
+  const { sendEvent } = useWaterfallAnalytics();
+  const [, startTransition] = useTransition();
+  const [requesters, setRequesters] = useQueryParam(
+    WaterfallFilterOptions.Requesters,
+    [] as string[],
+  );
```
minnakt, line 21: nit.
```suggestion
  const [requesters, setRequesters] = useQueryParam<string[]>(
    WaterfallFilterOptions.Requesters,
    [],
  );
```

apps/spruce/src/pages/waterfall/useFilters.ts
```diff
@@ -0,0 +1,98 @@
+import { useMemo } from "react";
+import { WaterfallQuery, WaterfallVersionFragment } from "gql/generated/types";
+import { useQueryParam } from "hooks/useQueryParam";
+import { WaterfallFilterOptions } from "types/waterfall";
+
+export const useFilters = (waterfall: WaterfallQuery["waterfall"]) => {
+  const [requesters] = useQueryParam(
+    WaterfallFilterOptions.Requesters,
+    [] as string[],
+  );
```
minnakt, line 10: (same comment)

apps/spruce/src/pages/waterfall/useFilters.ts
```diff
@@ -0,0 +1,98 @@
+import { useMemo } from "react";
+import { WaterfallQuery, WaterfallVersionFragment } from "gql/generated/types";
+import { useQueryParam } from "hooks/useQueryParam";
+import { WaterfallFilterOptions } from "types/waterfall";
+
+export const useFilters = (waterfall: WaterfallQuery["waterfall"]) => {
```
minnakt, line 6: Could we add tests for this hook (since it looks like we might use it a lot!)?

apps/spruce/src/pages/waterfall/useFilters.ts
```diff
@@ -0,0 +1,98 @@
...
+  const activeVersionIds = useMemo(
+    () =>
+      new Set(
+        versions.reduce((ids, { version }) => {
+          if (version) {
+            ids.push(version.id);
+          }
+          return ids;
+        }, [] as string[]),
+      ),
```
minnakt, line 59: nit. I think you can do it without casting like this
```suggestion
        versions.reduce((ids: string[], { version }) => {
          if (version) {
            ids.push(version.id);
          }
          return ids;
        }, []),
      ),
```

or 

```suggestion
       versions.reduce<string[]>((ids, { version }) => {
          if (version) {
            ids.push(version.id);
          }
          return ids;
        }, []),
```

apps/spruce/src/pages/waterfall/useFilters.ts
```diff
@@ -0,0 +1,98 @@
...
+    [versions],
+  );
+
+  const buildVariants = useMemo(() => {
+    if (!hasFilters) {
+      return waterfall.buildVariants;
+    }
+
+    const bvs: typeof waterfall.buildVariants = [];
```
minnakt, line 68: nit. should we use the GraphQL type?
```suggestion
    const bvs: WaterfallBuildVariant[] = []
```

apps/spruce/src/pages/waterfall/useFilters.ts
```diff
@@ -0,0 +1,98 @@
...
+    waterfall.buildVariants.forEach((bv) => {
+      if (activeVersionIds.size !== bv.builds.length) {
+        const activeBuilds: typeof bv.builds = [];
```
minnakt, line 71: nit. (same question)
```suggestion
        const activeBuilds: WaterfallBuild[] = [];
```

apps/spruce/src/pages/waterfall/useFilters.ts
minnakt, line 6: should the filters clear when switching to a different project? (not sure if this is captured by another ticket 😅)

minnakt (approved):

PR #458, opened 2024-10-23
DEVPROD-11699 Bump MongoDB to 8.0

khelif96:
DEVPROD-11699

### Description
Bump MongoDB to 8.0 to match prod

### DIFF
--- a/.evergreen/evergreen.yml
+++ b/.evergreen/evergreen.yml
@@ -46,9 +46,9 @@ buildvariants:
     expansions:
       app_dir: apps/spruce
       goroot: /opt/golang/go1.20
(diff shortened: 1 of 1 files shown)


.evergreen/evergreen.yml
```diff
@@ -46,9 +46,9 @@ buildvariants:
     expansions:
```
minnakt, line 46: I think you need to update one more spot, the `lib` variant is still using the old versions

minnakt (approved):

.evergreen/evergreen.yml
khelif96: I just removed them since we don't actually use them!

PR #461, opened 2024-10-25
DEVPROD-10189: Add pagination to waterfall

minnakt:
DEVPROD-10189

### Description
Add pagination to the waterfall. Currently, this implementation does not leverage the cache in the best way, but it's an initial step to having something working for pagination.

### Screenshots

https://github.com/user-attachments/assets/42f49f00-c344-4d79-a1bb-fbfb23170bb1

### Testing
* Cypress tests

### Evergreen PR
* evergreen-ci/evergreen/pull/8431


minnakt:
it's working, but I notice that there's no loading animation now. It's just blank until it figures out what `versions` & `buildVariants` to read

minnakt:
Notes on this PR:
* The `read` function is currently set to return `undefined`, so the `read` function doesn't actually do anything at the moment. This is on purpose.

Learnings from previous commits:
* I added the `Version` field to `WaterfallBuild`, but this is actually kind of weird 😅 A `WaterfallBuild` has multiple versions so this `Version` field actually references its first `Version`. However, we need some sort of unique identifier for the `WaterfallBuild` object or else it can't leverage the cache correctly.
* The `read` and `merge` functions are actually more complicated than I thought. Earlier commits on this PR have a naive implementation, but they don't consider that we likely have to maintain a structure that stores the results for _each_ project; otherwise every project's commits will get merged into one mega array.
* `fetchMore` doesn't seem to clear the existing entries when re-entering a loading state. You can see this on the project events & distro events pages. `fetchMore` keeps existing entries on the page and loads more in. This is probably intentional by Apollo but problematic for our intended implementation.

apps/spruce/src/gql/client/cache.ts
sophstad, line 1: I noticed a previous commit included `keyFields` for the waterfall query in this file; do you know if that will address this concern mentioned in your comment?

> they don't consider that we likely have to maintain a structure that stores the results for each project

apps/spruce/src/pages/waterfall/WaterfallGrid.tsx
```diff
@@ -16,36 +22,56 @@ import {
...
+
   const refEl = useRef<HTMLDivElement>(null);
   const { height } = useDimensions(
     refEl as React.MutableRefObject<HTMLElement>,
   );
 
-  const { buildVariants, versions } = useFilters(data.waterfall);
+  const groupedVersions = groupInactiveVersions(
+    data.waterfall.flattenedVersions,
+  );
```
sophstad, line 69: If we stick with flattened versions, would it make sense to do this in `useFilters`?

apps/spruce/src/pages/waterfall/PaginationButtons/index.tsx
sophstad, line 1: Nice, it seems like caching works decently if you're navigating to a page for the second time; i.e. when going forward then back then forward again, the cached page is loaded quickly on the second visit.

But that did bring up an edge case for followup work: if you navigate back to the zeroth page but a new version has since been added to the waterfall, you can navigate backwards to it and only one version is loaded on the page! You can observe this somewhat easily in a frequently-commited project like mms. This might be something we could tangentially address in DEVPROD-11707 too.

apps/spruce/src/pages/waterfall/PaginationButtons/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+    sendEvent({
+      name: "Changed page",
+      direction: "previous",
+    });
+    setQueryParams({
+      ...queryParams,
+      [WaterfallFilterOptions.MaxOrder]: undefined,
+      [WaterfallFilterOptions.MinOrder]: prevPageOrder,
+    });
+  };
```
sophstad, line 41: Could we implement these with [useTransition](https://react.dev/reference/react/useTransition)? It would improve the ability to interact with the page while cached data is rendering as described above. (I haven't tried it myself so hopefully this doesn't break the loading animation for non-cached pages)

apps/spruce/src/pages/waterfall/RequesterFilter/index.tsx
```diff
@@ -40,6 +40,7 @@ export const RequesterFilter = () => {
     >
       {commitRequesters.map((requester) => (
         <ComboboxOption
+          key={`${requester}-option`}
```
sophstad, line 43: What's this for?

sophstad (commented):
Cool, I'm pretty happy with how this works! Can't thank you enough for sticking with it. Mostly had some questions!

apps/spruce/src/pages/waterfall/RequesterFilter/index.tsx
minnakt: Just because there was an error in the console asking to add a `key`!

sophstad: 🤦‍♀️

apps/spruce/src/gql/client/cache.ts
minnakt: You know what, I just realized that the [query](https://github.com/evergreen-ci/ui/blob/c53ae70faeca3aca6267673580eaa54378c61c76/apps/spruce/src/gql/queries/waterfall.graphql#L3) probably just has to be rewritten slightly by making a discrete argument `$projectIdentifier` in addition to `$options`. Then it should be able to recognize the `$projectIdentifier` args field and work as you described

minnakt: maybe this actually 😄:

Screenshot 2024-11-05 at 12 37 38 AM

minnakt: Confirmed that setting it to `keyArgs: ["options", ["projectIdentifier"]]` worked (didn't push but let me know if you want me to push it)

apps/spruce/src/pages/waterfall/PaginationButtons/index.tsx
minnakt: interesting I will check this tomorrow! it sounds like some ways it could be fixed would be:
* Use the cache. But I'm not sure that the `read` function can allow us to restore the loading state
* Do a better calculation for `prevPageOrder` (if num active versions on prev page < LIMIT, set `prevPageOrder` to 0. Not sure how expensive extra query would be)

apps/spruce/src/pages/waterfall/PaginationButtons/index.tsx
minnakt: Tried it (hopefully correctly) and pushed it, but it's looking like the loading animations did indeed disappear...

sophstad (changes requested):
Yay! Almost there, just a few followup items to ticket and let's add that project key.

apps/spruce/src/gql/client/cache.ts
minnakt: Added it in but I think it requires a `merge` function and a `read` function (or else the page just froze for me), so I added those back in as well. The `read` function returns `undefined` so I believe there's now 0 caching entirely 😅 I can keep it in or revert the commit

sophstad: ohhh hm in that case I think let's leave it for a separate PR. sorry for the flip flopping!

minnakt: reverted!

apps/spruce/src/pages/waterfall/PaginationButtons/index.tsx
minnakt: Created [DEVPROD-12594](https://jira.mongodb.org/browse/DEVPROD-12594)

apps/spruce/src/pages/waterfall/PaginationButtons/index.tsx
```diff
@@ -0,0 +1,71 @@
...
+}
+
+export const PaginationButtons: React.FC<PaginationButtonsProps> = ({
+  pagination,
+}) => {
+  const { hasNextPage, hasPrevPage, nextPageOrder, prevPageOrder } =
+    pagination ?? {};
+  const { sendEvent } = useWaterfallAnalytics();
+
+  const [, startTransition] = useTransition();
```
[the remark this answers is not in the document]
minnakt: Created [DEVPROD-12595](https://jira.mongodb.org/browse/DEVPROD-12595)

sophstad (approved):
🫶

PR #468, opened 2024-10-28
DEVPROD-7857: Send Sentry alerts for 422 errors in Spruce

minnakt:
DEVPROD-7857

### Description
Send immediate Sentry alerts for 422 errors.

This PR also moves certain utils up to the `lib` folder so that they can be used in the upcoming PR for Parsley, which requires a lot more changes.

### Testing
- [Sample issue](https://mongodb-org.sentry.io/issues/6025875166/?referrer=slack&notification_uuid=a77e421c-39d8-4d51-91ce-08ee813d1b3c&environment=staging&alert_rule_id=14677842&alert_type=issue)

### DIFF
--- a/apps/spruce/src/gql/client/link/logGQLErrorsLink.ts
+++ b/apps/spruce/src/gql/client/link/logGQLErrorsLink.ts
@@ -1,5 +1,6 @@
 import { Operation } from "@apollo/client";
 import { onError } from "@apollo/client/link/error";
+import { ApolloServerErrorCode } from "@apollo/server/errors";
 import { GraphQLError } from "graphql";
 import { reportError } from "utils/errorReporting";
 import { deleteNestedKey } from "utils/object";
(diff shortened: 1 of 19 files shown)


apps/spruce/src/gql/client/link/logGQLErrorsLink.ts
```diff
@@ -11,7 +12,20 @@ export const reportingFn =
...
+
+    const isValidationError =
+      gqlErr.extensions?.code ===
+      ApolloServerErrorCode.GRAPHQL_VALIDATION_FAILED;
+
+    const err = isValidationError
+      ? new GraphQLError(
+          `GraphQL validation error in '${operation.operationName}': ${gqlErr.message}`,
+        )
+      : new Error(
```
SupaJoon, line 24: Why not utilize `GraphQLError` instead of `Error`?

sophstad (approved):

SupaJoon (approved):

apps/spruce/src/gql/client/link/logGQLErrorsLink.ts
minnakt: I was thinking `GraphQLError` would be like a legitimate error in GraphQL, whereas `Error` might be an error in our actual application! So we might get an error because of wrong implementation in Evergreen, but it doesn't mean that something is wrong in GraphQL itself

minnakt:
merging because the failing test has been failing across all PRs

PR #470, opened 2024-10-29
DEVPROD-12386 Track `is_patch` and `requester` for tasks and versions

khelif96:
DEVPROD-12386

### Description
This change adds automatic tracking of requesters for version and task interactions in honeycomb. This will be useful to determine what types of actions people take on different types of version and task pages.

### DIFF
--- a/apps/spruce/src/analytics/version/useVersionAnalytics.ts
+++ b/apps/spruce/src/analytics/version/useVersionAnalytics.ts
@@ -62,10 +62,12 @@ export const useVersionAnalytics = (id: string) => {
       fetchPolicy: "cache-first",
     },
   );
-  const { status } = eventData?.version || {};
+  const { isPatch, requester, status } = eventData?.version || {};
 
   return useAnalyticsRoot<Action, AnalyticsIdentifier>("Version", {
     "version.status": status || "",
     "version.id": id,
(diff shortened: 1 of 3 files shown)


apps/spruce/src/analytics/version/useVersionAnalytics.ts
```diff
@@ -62,10 +62,12 @@ export const useVersionAnalytics = (id: string) => {
...
     },
   );
-  const { status } = eventData?.version || {};
+  const { isPatch, requester, status } = eventData?.version || {};
 
   return useAnalyticsRoot<Action, AnalyticsIdentifier>("Version", {
     "version.status": status || "",
     "version.id": id,
+    "version.is_patch": isPatch || false,
+    "version.requester": requester || "",
```
minnakt, line 71: should we track `requester` for `usePatchAnalytics` or does it not matter for them?

minnakt (approved):

apps/spruce/src/analytics/version/useVersionAnalytics.ts
khelif96: I don't think its as valuable as it is for version interactions. We also don't currently return requester for patches in our API so we could add it down the line but outside of the scope of this ticket.

PR #472, opened 2024-10-31
DEVPROD-12354: Instrument waterfall render

sophstad:
DEVPROD-12354

### Description

Add instrumentation that creates a span measuring the duration of the waterfall render. Timing does not include the length of request.

I had hoped to add this span to the trace that includes the HTTP.POST event, but I was unable to do so. I'd prefer to start tracking this measurement and potentially figure that out later.

### Testing

[Example span](https://ui.honeycomb.io/mongodb-4b/environments/staging/datasets/spruce/result/rfdvjSkAbeW/trace/C113KhhWPh5?fields[]=s_name&source=query&span=e45e5b4be0f3e987) - this render very performant since staging doesn't have any sizable projects.

### DIFF
--- a/apps/spruce/src/pages/waterfall/WaterfallGrid.tsx
+++ b/apps/spruce/src/pages/waterfall/WaterfallGrid.tsx
@@ -1,6 +1,8 @@
-import { useRef } from "react";
+import { useEffect, useRef } from "react";
 import { useSuspenseQuery } from "@apollo/client";
 import styled from "@emotion/styled";
+import { trace } from "@opentelemetry/api";
+import { TRACER_NAME } from "@evg-ui/lib/analytics/utils";
 import { DEFAULT_POLL_INTERVAL } from "constants/index";
 import { WaterfallQuery, WaterfallQueryVariables } from "gql/generated/types";
 import { WATERFALL } from "gql/queries";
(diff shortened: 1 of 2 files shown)


apps/spruce/src/pages/waterfall/WaterfallGrid.tsx
```diff
@@ -21,9 +23,26 @@ type WaterfallGridProps = {
   projectIdentifier: string;
 };
 
+const tracer = trace.getTracer(TRACER_NAME);
```
minnakt, line 26: I wonder if we want to use a different tracer name here (you can use any name you want I believe, like `performance`, `render`, etc)

Calling it with `analytics` tracer name means it shows up in our events like this:

Screenshot 2024-11-01 at 10 48 26 AM

or we can rename the event to follow our conventions as `Rendered waterfall` and add it to the analytics spreadsheet

apps/spruce/src/pages/waterfall/WaterfallGrid.tsx
```diff
@@ -21,9 +23,26 @@ type WaterfallGridProps = {
...
+
+  useEffect(() => {
+    resolveRender();
+  }, []);
+
+  tracer.startActiveSpan("render waterfall", async (span) => {
+    renderPromise.then(() => {
+      span.end();
+    });
+  });
```
minnakt, line 44: i wonder if it might make sense to extract this logic into a hook like `useWaterfallTrace` (or something like that) so that the logic is isolated elsewhere!

apps/spruce/src/pages/waterfall/WaterfallGrid.tsx
```diff
@@ -21,9 +23,26 @@ type WaterfallGridProps = {
...
   projectIdentifier,
 }) => {
+  let resolveRender: (v?: any) => void;
+  const renderPromise = new Promise((resolve) => {
+    resolveRender = resolve;
+  });
+
+  useEffect(() => {
+    resolveRender();
+  }, []);
```
minnakt, line 38: nit. (fix eslint error)

Screenshot 2024-11-01 at 10 52 50 AM

minnakt (approved):

PR #473, opened 2024-11-02
DEVPROD-7857: Send Sentry alerts for 422 errors in Parsley

minnakt:
DEVPROD-7857

### Description
Copies Spruce's configuration for logging GraphQL errors. I would move some stuff up to the `lib` level but I think it'd make more sense to move `utils/errorReporting` up first.

### Testing
* [Sample issue](https://mongodb-org.sentry.io/issues/6037168995/?alert_rule_id=14709291&alert_type=issue&environment=staging&notification_uuid=0ebd2699-08b8-4948-91c4-0f2b8dee214b&project=4505902479441920&referrer=slack)


SupaJoon (approved):
Do you think the GQL client creation code can be shared? Not suggesting for this PR but curious what you think.

minnakt:
> Do you think the GQL client creation code can be shared? Not suggesting for this PR but curious what you think.

good question... I think it would be good if we could share it. The only thing that might make it tricky is how each app has its own set of queries so we can't seem to really "share" the `SecretFields` query. ☹️ I feel like that might be the first obstacle.

PR #484, opened 2024-11-06
DEVPROD-803 Support uploading files to Parsley via clipboard [minor]

khelif96:
DEVPROD-803

### Description
Adds support to copy and paste files into Parsley from the upload page.
Went against the approach of including a text box because the UX didn't work and pasting a large file into a text box causes the browser to freeze up.

### Screenshots
image

### DIFF
--- a/apps/parsley/src/pages/LogDrop/state.ts
+++ b/apps/parsley/src/pages/LogDrop/state.ts
@@ -1,4 +1,5 @@
 import { useReducer } from "react";
+import { LogDropType } from "./constants";
 
 type CurrentState =
   | "WAITING_FOR_FILE"
@@ -10,8 +11,9 @@ type State = {
   currentState: CurrentState;
   file: File | null;
   text: string | null;
-  type: "file" | "text" | null;
+  type: LogDropType | null;
 };
+
 type Action =
   | { type: "DROPPED_FILE"; file: File }
   | { type: "PARSE_FILE" }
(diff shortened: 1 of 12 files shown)


apps/parsley/src/pages/LogDrop/state.ts
```diff
@@ -9,15 +9,20 @@ type CurrentState =
 type State = {
   currentState: CurrentState;
   file: File | null;
+  text: string | null;
+  type: "file" | "text" | null;
```
sophstad, line 13: Could you use an enum here to avoid magic strings?

sophstad (commented):
Would it be significant scope creep to accept a pasted file? When I did so it just parsed the filename instead of its contents. No worries if not, this is still a nice improvement

khelif96:
> Would it be significant scope creep to accept a pasted file? When I did so it just parsed the filename instead of its contents. No worries if not, this is still a nice improvement

I think it may be. From what I understand about the feature request, the typical use case is that users run some tests locally, copy the results from their terminal, and paste them into Parsley. Accepting a pasted file just seems like adding a third way to perform the same functionality we already support with drag-and-drop and the file picker. 
As an aside I can't really figure out how to detect the file clip in the first place it keeps appearing as text clipboard contents.

sophstad (approved):
Nice! Would this perhaps qualify for a minor release given the guidelines we discussed yesterday? If so, I'd remove the Spruce GQL file from this diffset so that Spruce doesn't also get a minor version bump.

PR #485, opened 2024-11-06
DEVPROD-12004: Send entire public key document when editing spawn host

minnakt:
DEVPROD-12004

### Description
This bug is caused by the fact that the user has a public key with an empty string name (`""`). I have no idea how they were able to do this, because we have checks to make sure this doesn't happen. I have told the user in the ticket to delete their empty key.

Regardless, the error behavior can be fixed by making sure that we send the entire public key document when editing the host.

Separately, fixed nearby TypeScript errors and updated deprecated `ConfirmationModal` props.

### Steps to reproduce original bug
1. Have a public key with an empty string name (`""`). I had to add via the database since you cannot do this from the UI.
2. Have a public key with a non-empty name. Use the **same key** as the public key above.
3. Try to edit host with your public key. It will fail. This is because it diffs the keys (which are the same) and results in an empty string. So we end up trying to add an empty key.

### DIFF
--- a/apps/spruce/src/pages/spawn/spawnHost/spawnHostButton/SpawnHostModal.tsx
+++ b/apps/spruce/src/pages/spawn/spawnHost/spawnHostButton/SpawnHostModal.tsx
@@ -61,12 +61,12 @@ export const SpawnHostModal: React.FC<SpawnHostModalProps> = ({
     SpawnHostMutation,
     SpawnHostMutationVariables
   >(SPAWN_HOST, {
-    onCompleted(hostMutation) {
+    onCompleted: (hostMutation) => {
       const { id } = hostMutation?.spawnHost ?? {};
       dispatchToast.success(`Successfully spawned host: ${id}`);
       setOpen(false);
     },
-    onError(err) {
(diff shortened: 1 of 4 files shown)


apps/spruce/src/pages/spawn/spawnHost/spawnHostButton/SpawnHostModal.tsx
```diff
@@ -151,14 +145,16 @@ export const SpawnHostModal: React.FC<SpawnHostModalProps> = ({
...
-      data-cy="spawn-host-modal"
-      onCancel={() => {
-        setOpen(false);
+      cancelButtonProps={{
+        onClick: () => setOpen(false),
+      }}
+      confirmButtonProps={{
+        children: loadingSpawnHost ? "Spawning" : "Spawn a host",
+        onClick: spawnHost,
+        disabled: hasError || loadingSpawnHost,
```
khelif96, line 154: Thanks for doing this!

khelif96 (approved):
Great job with the investigation!

PR #488, opened 2024-11-07
DEVPROD-811: Require copy assertion in `validateToast`

minnakt:
DEVPROD-811

### Description
Add copy with `validateToast` where missing. Removed any copy that comes from the backend because it changes too frequently to maintain.

### DIFF
--- a/apps/parsley/cypress/support/commands.ts
+++ b/apps/parsley/cypress/support/commands.ts
@@ -191,7 +191,7 @@ Cypress.Commands.add("toggleDrawer", () => {
 
 Cypress.Commands.add(
   "validateToast",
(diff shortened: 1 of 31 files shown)


sophstad (commented):
> Removed any copy that comes from the backend because it changes too frequently to maintain.

By this, do you just mean that you used partial strings in the validate toast method and tried to manually omit the backend details?

minnakt:
> > Removed any copy that comes from the backend because it changes too frequently to maintain.
> 
> By this, do you just mean that you used partial strings in the validate toast method and tried to manually omit the backend details?

I removed any part of the string that is not present in the frontend!

sophstad (approved):

PR #489, opened 2024-11-07
DEVPROD-11709: Add tasks filter to waterfall

sophstad:
DEVPROD-11709

### Description

- Apply tasks filter and remove task squares that don't match. Also omit build variants with no matching tasks
- Update filter badges to use Chip's truncation implementation. This required updating some tests that tested for the presence of a string in a badge.

### Screenshots

image

### Testing

- Add unit and e2e tests
- Update unit test data to be more consistent


apps/spruce/src/pages/waterfall/TaskStatsTooltip.tsx
```diff
@@ -27,7 +27,7 @@ export const TaskStatsTooltip: React.FC<
     0;
 
   return (
-    <>
+    <div>
```
sophstad, line 30: Adding this div fixes above where the rightmost task stats popover was not positioned correctly

minnakt (approved):

PR #506, opened 2024-11-19
DEVPROD-10198: Show updated task status legend on waterfall page

minnakt:
DEVPROD-10198

### Description
Adds new task status legend while keeping support for old task status legend.

In order for floating content to be positioned properly, I set `overflow-x: hidden` on the waterfall page. I'm not really sure why that works but it fixes the position of the portalled elements during scroll (tooltip & popover) 🤷 I checked Safari and didn't notice any performance issue... Another option is to just set `usePortal` to false, although this has some weird side effects with opacity.

### Screenshots

Order and layout should match Figma designs

Screenshot 2024-11-19 at 3 31 15 PM


apps/spruce/src/components/TaskStatusIconLegend/index.tsx
```diff
@@ -1,109 +1,152 @@
...
+export const LegendContent: React.FC<LegendContentProps> = ({
+  useWaterfall,
+}) => {
+  const Container = useWaterfall
+    ? WaterfallContainer
+    : MainlineCommitsContainer;
+
+  const groupedStatuses = useWaterfall
+    ? waterfallGroupedStatuses
+    : groupedIconStatuses;
```
sophstad, line 32: Hm this is supposed to show the old icon set on mainline commits right? That isn't working for some reason (and I do think showing separate icon sets is the desired behavior)
image

minnakt: added more boolean logic to make this behave better !

apps/spruce/src/components/Content/Layout.tsx
```diff
@@ -55,7 +56,8 @@ export const Layout: React.FC = () => {
         />
       )}
       <FloatingContent>
-        <TaskStatusIconLegend />
+        {/* TODO: Use user's beta feature setting instead. */}
```
sophstad, line 59: Let's include the ticket number here
```suggestion
        {/* TODO DEVPROD-10203: Use user's beta feature setting instead. */}
```

apps/spruce/src/components/TaskStatusIconLegend/index.tsx
```diff
@@ -1,109 +1,158 @@
...
-                setIsActive(false);
+                setOpen(false);
               }}
             >
               <Icon glyph="X" />
             </IconButton>
           </TitleContainer>
-          <LegendContent />
+          <LegendContent
+            useWaterfall={isWaterfall || (useWaterfall && !isMainlineCommits)}
```
sophstad, line 140: If I'm not mistaken the build variant and task history pages use the old icon set. Maybe for now we should only show the new icon set on the waterfall page and default to the old set. When the task history page is redone we can update the legend there too.

sophstad (changes requested):

apps/spruce/src/components/TaskStatusIconLegend/index.tsx
minnakt: ok sounds good!

sophstad (approved):

PR #507, opened 2024-11-19
DEVPROD-9386 Add disclaimer to missing project event variable entries

khelif96:
DEVPROD-9386

### Description
Display a disclaimer on Evergreen Event log entries to indicate project variables are unavailable.

### Screenshots
image

### DIFF
--- a/apps/spruce/src/components/Settings/EventLog/EventDiffTable.tsx
+++ b/apps/spruce/src/components/Settings/EventLog/EventDiffTable.tsx
@@ -5,14 +5,25 @@ import { useLeafyGreenTable, LGColumnDef } from "@leafygreen-ui/table";
 import { fontFamilies } from "@leafygreen-ui/tokens";
 import { BaseTable } from "components/Table/BaseTable";
 import { getEventDiffLines } from "./eventLogDiffs";
-import { Event, EventDiffLine, EventValue } from "./types";
+import {
+  CustomKeyRenderConfig,
+  Event,
+  EventDiffLine,
+  EventValue,
+} from "./types";
+import { applyCustomKeyRender } from "./utils";
 
 type TableProps = {
   after: Event["after"];
   before: Event["before"];
+  customKeyRenderConfig?: CustomKeyRenderConfig;
 };
 
-export const EventDiffTable: React.FC<TableProps> = ({ after, before }) => {
+export const EventDiffTable: React.FC<TableProps> = ({
+  after,
+  before,
+  customKeyRenderConfig,
+}) => {
   const eventLogEntries = useMemo(
     () => getEventDiffLines(before, after) ?? [],
     [after, before],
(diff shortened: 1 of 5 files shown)


apps/spruce/src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -25,6 +28,13 @@ export const EventLogTab: React.FC<TabProps> = ({ limit, projectType }) => {
   return (
     <EventLog
       allEventsFetched={allEventsFetched}
+      customKeyRenderConfig={{
+        "vars.vars": (key) => (
+          <StyledInlineDefinition definition="Evergreen does not display project variable values.">
```
sophstad, line 33: IMO this isn't really offering the user any additional context. Maybe something like this?
```suggestion
          <StyledInlineDefinition definition="Evergreen does not display project variable values in the event log for security reasons.">
```

apps/spruce/src/components/Settings/EventLog/EventDiffTable.tsx
```diff
@@ -69,11 +80,17 @@ const renderEventValue = (value: EventValue): string => {
...
-const columns: LGColumnDef<EventDiffLine>[] = [
+const columns = (
+  customKeyRenderConfig: CustomKeyRenderConfig = {},
+): LGColumnDef<EventDiffLine>[] => [
   {
     header: "Property",
     accessorKey: "key",
-    cell: ({ getValue }) => <CellText>{getValue() as string}</CellText>,
+    cell: ({ getValue }) => (
+      <CellText>
```
sophstad, line 90: This may be more helpful if it were applied on the REDACTED section instead of the variable name. The redacted text is the change that we're actually drawing attention to, not the variable name.

khelif96: That was actually what I originally implemented before I realized we had a [design](https://www.figma.com/design/EgYGXJtGBz43B6StlL46B2/EXPO-4044%3A-Project-Var-Hash-Updates?node-id=48-2624) linked haha. If you think that makes more sense I can move it.

sophstad: Ha! I do think moving it makes more sense, especially since those designs still include the variable values so maybe it wasn't super clear what the ask was.

apps/spruce/src/pages/projectSettings/tabs/EventLogTab/EventLogTab.tsx
```diff
@@ -25,6 +28,13 @@ export const EventLogTab: React.FC<TabProps> = ({ limit, projectType }) => {
   return (
     <EventLog
       allEventsFetched={allEventsFetched}
+      customKeyValueRenderConfig={{
+        "vars.vars": (key) => (
+          <StyledInlineDefinition definition="Evergreen does not display project variable values in the event log for security reasons.">
+            {key}
```
sophstad, line 34: [nit]
```suggestion
        "vars.vars": (val) => (
          <StyledInlineDefinition definition="Evergreen does not display project variable values in the event log for security reasons.">
            {val}
```

sophstad (approved):

PR #512, opened 2024-11-21
DEVPROD-10200: Implement "Go to Date" feature

minnakt:
DEVPROD-10200

### Description

This PR adds the date filter on the new waterfall page. The date is shown in a user-friendly readable format (`YYYY-MM-DD`) in the URL so users can easily manipulate the URL.

The date and pagination query params cannot exist at the same time because it leads to conflicts (and errors) in `maxOrderNumber` and `minOrderNumber`.

Other things in this PR:
* Update `DatePicker` component
* Moved Cypress tests for filtering from `waterfall/waterfall.ts` to `waterfall/filters.ts `

Tips for testing locally:
* Remember to set the year to 2022.
* You may run into [DEVPROD-12594](https://jira.mongodb.org/browse/DEVPROD-12594) when paginating backwards after using the date filter.

### Screenshots

https://github.com/user-attachments/assets/ed5999cc-47d4-49c6-a422-0b8e892bdb76

### Testing
- Added Cypress tests (which will fail until backend PR is merged)

### Evergreen PR
* evergreen-ci/evergreen/pull/8501


minnakt:
evergreen retry

apps/spruce/src/pages/waterfall/DateFilter/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+      sendEvent({ name: "Filtered by date" });
+    }
+  };
+
+  return (
+    <DatePicker
+      data-cy="date-picker"
+      label="Go to Date"
+      onDateChange={handleChange}
+      value={date.length ? new Date(date) : undefined}
```
sophstad, line 39: Using `DatePicker` as an uncontrolled component will provide a better UX by allowing the component to update its date itself and not lag while waiting for the data to come back from the server. I believe the only change necessary is to use `initialValue` instead of `value`. This pattern is used in the RequesterFilter too if you want to take a look.

apps/spruce/src/pages/waterfall/VersionLabel/index.tsx
```diff
@@ -72,7 +72,7 @@ export const VersionLabel: React.FC<Props> = ({
             to={getVersionRoute(id)}
           >
             {shortenGithash(revision)}
-          </InlineCode>
+          </InlineCode>{" "}
```
sophstad, line 75: 🤩

sophstad (changes requested):

apps/spruce/src/pages/waterfall/DateFilter/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+        });
+      });
+      sendEvent({ name: "Filtered by date" });
+    }
+  };
+
+  return (
+    <DatePicker
+      data-cy="date-picker"
+      label="Go to Date"
```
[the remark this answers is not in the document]
minnakt: good idea!

apps/spruce/src/pages/waterfall/DateFilter/index.tsx
minnakt: done

apps/spruce/src/pages/waterfall/DateFilter/index.tsx
```diff
@@ -34,9 +34,11 @@ export const DateFilter = () => {
   return (
     <DatePicker
       data-cy="date-picker"
+      // Use an uncontrolled component so that the transition does not DatePicker rendering.
```
sophstad, line 37:
```suggestion
      // Use an uncontrolled component so that the transition does not block DatePicker rendering.
```

sophstad (approved):
Should we consider adding an index to improve performance on this query?

minnakt:
> Should we consider adding an index to improve performance on this query?

It looks like we have an index that fits pretty well with the query on the versions collection: `identifier_1_r_1_create_time_1_order_1`

But when I ran the query, it actually rejected the index and used this one instead: `identifier_1_r_1_order_1`

Maybe we can add a hint on the query to make it use the first index! I will create a ticket

PR #514, opened 2024-11-22
DEVPROD-10194 Create git commit search component for waterfall

khelif96:
DEVPROD-10194

### Description
Create the git commit selection modal. This is lifted from the project health page with some modifications. I chose not to reuse that component to make it easier to delete the legacy page when we are done.
### Screenshots
image
image


apps/spruce/src/pages/waterfall/InactiveVersions/InactiveVersionsModal.tsx
```diff
@@ -29,6 +34,7 @@ export const InactiveVersionsModal: React.FC<Props> = ({
       {versions?.map((version) => (
         <StyledVersionLabel
           key={version.id}
+          highlighted={revision !== null && version.revision.includes(revision)}
```
sophstad, line 37: This is a nice addition! Do you think it would make sense to apply the highlight around the inactive commits button to provide an indicator that the commit is found there?

apps/spruce/src/pages/waterfall/WaterfallMenu/GitCommitSearch.tsx
```diff
@@ -0,0 +1,80 @@
...
+          onClick: onConfirm,
+          disabled: commitHash.trim().length < 7,
+        }}
+        data-cy="git-commit-search-modal"
+        open={modalOpen}
+        title="Search by Git Commit Hash"
+      >
+        <StyledDescription>
+          Note: This is an experimental feature that works best without any
+          filters. Searching for a git hash will clear all applied filters.
```
sophstad, line 65: IMO this isn't really an experimental feature anymore since it's been used successfully for at least a year. I also don't think filters have any impact on the search now that filtering is client-side.
```suggestion
          Searching for a git hash will clear all applied filters.
```

apps/spruce/src/pages/waterfall/VersionLabel/index.tsx
```diff
@@ -23,6 +23,7 @@ type Props = WaterfallVersionFragment & {
   className?: string;
   shouldDisableText?: boolean;
   view: VersionLabelView;
+  highlighted: boolean;
```
sophstad, line 26: [nit] Alphabetize type
```suggestion
  className?: string;
  highlighted: boolean;
  shouldDisableText?: boolean;
  view: VersionLabelView;
```

apps/spruce/src/pages/waterfall/WaterfallMenu/GitCommitSearch.tsx
```diff
@@ -0,0 +1,80 @@
...
+    onCancel();
+  };
+
+  return (
+    <>
+      <DropdownItem
+        data-cy="git-commit-search"
+        onClick={() => {
+          setModalOpen(true);
+          sendEvent({ name: "Viewed git commit modal" });
```
sophstad, line 44: This seems only marginally useful to me; I guess it would allow us to see how many users looked at the modal without actually submitting a git hash query but that doesn't seem very insightful. I wouldn't mind dropping this if you didn't have a strong reason for it.

sophstad (changes requested):

apps/spruce/src/pages/waterfall/InactiveVersions/InactiveVersionsModal.tsx
khelif96: Great idea! I came up with this glowing effect like the one we use in Parsley. 
image

apps/spruce/src/pages/waterfall/InactiveVersions/index.tsx
```diff
@@ -25,10 +28,19 @@ export const InactiveVersionsButton: React.FC<Props> = ({
       0,
     ) ?? 0;
   const [modalOpen, setModalOpen] = useState(false);
+  const [revisionFilter] = useQueryParam<string | null>(
+    WaterfallFilterOptions.Revision,
+    null,
+  );
```
sophstad, line 34: Is there any reason not to type this consistently and use an empty string instead of null? Seems like it may be easier to only have to handle one non-value, but mostly just curious.

khelif96: I mainly did it to ensure the usage is consistent throughout the app. If the value is null in one place I thought it should be nullable everywhere else since its the same origin. I don't have a strong preference either way.

sophstad (approved):
🥳

PR #515, opened 2024-11-22
DEVPROD-11705: Collapse filtered versions

sophstad:
DEVPROD-11705

### Description

If a version has no active tasks due to filtering, collapse it by marking it as inactive and rendering it in the modal.

### Screenshots

image

### Testing

Update useFilter unit tests, add test for grouping inactive versions

### DIFF
--- a/apps/spruce/src/pages/waterfall/useFilters.ts
+++ b/apps/spruce/src/pages/waterfall/useFilters.ts
@@ -1,12 +1,7 @@
 import { useMemo } from "react";
 import { WaterfallVersionFragment } from "gql/generated/types";
 import { useQueryParam } from "hooks/useQueryParam";
-import {
-  Build,
-  BuildVariant,
-  WaterfallFilterOptions,
-  WaterfallVersion,
-} from "./types";
+import { Build, BuildVariant, WaterfallFilterOptions } from "./types";
 import { groupInactiveVersions } from "./utils";
 
 type UseFiltersProps = {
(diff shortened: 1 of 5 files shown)


apps/spruce/src/pages/waterfall/useFilters.ts
```diff
@@ -117,56 +60,77 @@ export const useFilters = ({
...
   ]);
 
+  const activeVersions = useMemo(() => {
+    const hasActiveBuild = (versionId: string) =>
+      filteredBuildVariants.some((bv) =>
+        bv.builds.some((build) => build.version === versionId),
+      );
+
+    return groupInactiveVersions(flattenedVersions, hasActiveBuild);
+  }, [filteredBuildVariants, flattenedVersions]);
```
minnakt, line 117: I was wondering what you think about moving the requester logic down here, it would be like:

```
  const activeVersions = useMemo(() => {
    const hasActiveBuild = (version: WaterfallVersionFragment) =>
      matchesRequesters(version, requesters) &&
      filteredBuildVariants.some((bv) =>
        bv.builds.some((build) => build.version === version.id),
      );
    return groupInactiveVersions(flattenedVersions, hasActiveBuild);
  }, [filteredBuildVariants, flattenedVersions, requesters]);
```

I think it would mean that you can remove all of the requester logic above

sophstad: This would complicate the render slightly by having a different number of builds vs active versions. I was able to get it to work, but it makes the useFilter test results slightly more confusing by including builds in the results that are not actually rendered. I'm fine with going with this approach if you think this tradeoff makes sense, but I was a bit undecided myself. Good callout though!

PR #525, opened 2024-11-27
DEVPROD-13116 Remove deprecated legacy success status

khelif96:
DEVPROD-13116

### Description
I found some todos that were left over from https://jira.mongodb.org/browse/DEVPROD-650 around removing a legacy success status. It looks like the backend ticket was completed but the frontend TODO's remained

### Evergreen PR
Some backend data will need to change 
https://github.com/evergreen-ci/evergreen/pull/8514

### DIFF
--- a/apps/spruce/src/components/PatchesPage/StatusSelector.tsx
+++ b/apps/spruce/src/components/PatchesPage/StatusSelector.tsx
@@ -35,6 +35,7 @@ export const StatusSelector: React.FC = () => {
 const statusValToCopy = {
   [ALL_PATCH_STATUS]: "All",
   [PatchStatus.Created]: "Created/Unconfigured",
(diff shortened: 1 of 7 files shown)


apps/spruce/src/components/PatchesPage/StatusSelector.tsx
```diff
@@ -56,11 +55,6 @@ const treeData = [
     value: PatchStatus.Started,
     key: PatchStatus.Started,
   },
-  {
```
sophstad, line 59: Should we replace this with `PatchStatus.Success` to allow filtering on success?

khelif96: Good catch

hadjri (approved):

PR #530, opened 2024-12-03
DEVPROD-12987: Fix event log key renderer

sophstad:
DEVPROD-12987

### Description

Fix regex used to format event log entries that contain an array so that numbers in variable names are not incorrectly array-ified.

### Screenshots

image

### Testing

Add test that fails without change


khelif96 (approved):
LGTM!

PR #534, opened 2024-12-05
DEVPROD-2731 Refactor app to use `displayStatus` for task statuses

khelif96:
DEVPROD-2731

### Description
Tasks with a known issue status often showed 2 badges that both displayed known issue which wasn't helpful and made it more difficult to identify what a tasks original status is. This PR adds an`displayStatus` field which will reflect the UI only status used on the app. In a follow up PR `task.status` will be updated to reflect a tasks original status.
### Screenshots
image

https://github.com/evergreen-ci/evergreen/pull/8529


apps/spruce/src/gql/queries/version-tasks.graphql
```diff
@@ -10,9 +10,10 @@ query VersionTasks(
       data {
         aborted
         baseTask {
+          displayStatus
           execution
           id
-          status
+          status: displayStatus
```
minnakt, line 16: I guess I don't fully understand why we need to alias `status` as `displayStatus`. Could we just keep `status` and use the new field `displayStatus` as we prepare for the transition of these fields? I assume that this will require multiple PRs anyways to migrate the field

apps/parsley/src/gql/fragments/base-task.graphql
```diff
@@ -3,7 +3,7 @@ fragment BaseTask on Task {
   execution
   id
   patchNumber
-  status
+  status: displayStatus
```
minnakt, line 6: for example here, it seems like we're not using the `displayStatus` field at all even though it was introduced in the backend PR

apps/spruce/src/gql/queries/version-tasks.graphql
khelif96: I initially kept both statuses as a backup, just in case I missed something during the refactor and didn’t want the page to break. But I decided to go ahead and update everything, and wow, there are so many places that rely on `status`. TypeScript wasn’t too happy with that, so the PR ended up growing more than I expected

khelif96 (commented):
This PR has a lot of type script changes. Most are an attempt at cleaning up and removing ts-expect-errors. In some spots our type safety gets lost so I tried to fix those.

In some spots like the commits folder I did add some `ts-expect-errors` for some more tricky types. But all functionality should still work and we are getting rid of the page soon.

PR #541, opened 2024-12-09
DEVPROD-10167: Move toast context to lib

minnakt:
DEVPROD-10167

### Description
Moves toast up to the `lib` package.


khelif96 (approved):
Great job thanks for doing this!

PR #547, opened 2024-12-18
DEVPROD-13604 Create Holiday tree for Spruce

khelif96:
DEVPROD-13604

### Description
Happy holidays!
Reused the same tree from https://github.com/evergreen-ci/spruce/pull/1590 with some tweaks to make it compatible with `AnimatedIcon`

### Screenshots
image


apps/spruce/src/components/Header/Navbar.tsx
```diff
@@ -72,7 +72,7 @@ export const Navbar: React.FC = () => {
           onClick={() => sendEvent({ name: "Clicked logo link" })}
           to={routes.myPatches}
         >
-          <AnimatedIcon icon={FallTree} />
+          <AnimatedIcon icon={HolidayTree} />
         </LogoLink>
```
minnakt, line 76: super nitpick but I think the icon is slightly uncentered (vertically, not horizontally) and could be moved up by about 4px

minnakt (approved):

PR #548, opened 2024-12-20
DEVPROD-8939: Remove public-read from s3 sync

sophstad:
DEVPROD-8939

### Description

Jonathan pointed out that the default security policy is to not have public-read buckets anymore; security even got flagged when he toggled this setting on for an existing bucket. By removing the public-read ACL flag, this PR facilitates the rest of the individual user staging environments work, which will not use public-read buckets.

### Testing

[Sample patch](https://spruce.mongodb.com/version/6765d6ec567cab000723c66b/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC), staging and beta look good.


ybrill:
I think we need to do some work on the infra side to be ready to make the existing buckets private. Right now CloudFront is able to read from the buckets and serve them because they're public. In order to make them private we need to grant CloudFront permissions to the buckets. (the new stagings are set up this way, but we haven't done it for the existing buckets yet)

khelif96:
I removed UI's review request I think @ybrill  has the most context on this. afaik I thought these were required to make the spruce website accessible over the internet.

ybrill:
@khelif96 the setup we have for the new evergreen stagings is that instead of the bucket and its contents being public to the internet, CloudFront is given permission to serve the bucket. This is done through [OAC](https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/private-content-restricting-access-to-s3.html) which is supposed to be more secure than leaving the bucket public.
If it works well in the stagings we plan to implement it for the main staging/beta/prod distributions too.

ybrill:
Opened a ticket for implementing there: [DEVPROD-13682](https://jira.mongodb.org/browse/DEVPROD-13682)

ybrill (approved):
Sophie and I talked this over. So long as the bucket is public (i.e. the bucket policy allows anyone to GetObject) we shouldn't need the ACL. Sophie has demonstrated this in beta/staging where the bucket configuration is the same as it is for prod.

PR #549, opened 2024-12-20
DEVPROD-8939: Add sync-spruce-staging task

sophstad:
DEVPROD-8939 - depends on #548 

### Description

Introduce the `sync-spruce-staging` task to deploy Spruce to a user's personal staging environment.
- The task must be invoked with `staging_user_key` included as an Evergreen parameter. If not, it yields [setup failed](https://spruce.mongodb.com/version/677834e476fd0800074e2660/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
- The task overwrites a few staging environment variables to point to the correct front and back ends
- If `staging_user_key` is provided, the app will include an `X-Evergreen-Environment` header in its requests. This is used by the evergreen-multi staging backing to correctly route requests
- The kernel-tools CLI introduced in https://github.com/10gen/kernel-tools/pull/457 will check whether or not the staging environment is in sync with Spruce's remote main branch (and provide the user key). To force a push to their staging environment, a user can run the task directly with the `staging_user_key` parameter.

### Testing

Check it out at http://spruce.sophie-stadler.evergreen-staging.devprod.mongodb.com


ybrill:
I tried using this myself to give myself a UI in a new environment and ran into [an error](https://parsley.mongodb.com/evergreen/evergreen_ui_deploy_sync_spruce_staging_patch_5fcede2526a2ac6448609333c00b84a37c69b8a4_6769b7ab4a1233000754e708_24_12_23_19_19_08/0/task?bookmarks=0,229&shareLine=133). I _think_ the issue was that the task was assuming the UI has already been deployed.

.evergreen/deploy.yml
```diff
@@ -58,6 +66,26 @@ functions:
...
+      type: setup
+      params:
+        script: |
+          if ! [ -z "${staging_user_key}" ]; then
+            echo "BUCKET: spruce-${staging_user_key}" >> deploy-vars.yml
+            echo "REACT_APP_USER_KEY: ${staging_user_key}" >> deploy-vars.yml
+            echo "REACT_APP_SPRUCE_URL: https://spruce.${staging_user_key}.evergreen-staging.devprod.mongodb.com" >> deploy-vars.yml
+            echo "REACT_APP_UI_URL: https://evergreen-multi.staging.corp.mongodb.com" >> deploy-vars.yml
+            echo "USER_STAGING_ARN: ${user_staging_arn_prefix}${staging_user_key}.ui_put" >> deploy-vars.yml
+          else
```
khelif96, line 80: I think the step of providing a `staging_user_key` could be automated and just rely on the author field as a default. Would something like this work?
```suggestion
        script: |
          staging_user_key="${staging_user_key}"
          if [ -z "$staging_user_key" ]; then
            echo "No staging_user_key provided. Attempting to determine from author."
            # The staging_user_key is just the author username with a dash instead of a period.
            author=${author}
            staging_user_key=$(echo "$author" | sed 's/\./-/g')
            echo "staging_user_key: '$staging_user_key'"
          fi
          if ! [ -z "$staging_user_key" ]; then
            echo "BUCKET: spruce-$staging_user_key" >> deploy-vars.yml
            echo "REACT_APP_USER_KEY: $staging_user_key" >> deploy-vars.yml
            echo "REACT_APP_SPRUCE_URL: https://spruce.$staging_user_key.evergreen-staging.devprod.mongodb.com" >> deploy-vars.yml
            echo "REACT_APP_UI_URL: https://evergreen-multi.staging.corp.mongodb.com" >> deploy-vars.yml
            echo "USER_STAGING_ARN: ${user_staging_arn_prefix}$staging_user_key.ui_put" >> deploy-vars.yml
          else
```

sophstad: Interested in @ybrill's thoughts here, but my inclination is that we should not make any assumptions about this value in the UI codebase. The value will be passed in by the kernel-tools CLI so users will rarely need to specify it themselves; if they ever do, they could set a [default CLI parameter](https://docs.devprod.prod.corp.mongodb.com/evergreen/Project-Configuration/Parameterized-Builds#command-line-usage).

khelif96: Ahh I missed that the CLI will provide this default value. 

> To force a push to their staging environment, a user can run the task directly with the staging_user_key parameter.

I guess it can still help with this portion.  I would defer to you two on the ideal approach.

ybrill: either way works for me. Though technically the environment could be called anything

sophstad: 👍 For the sake of flexibility, let's stick with the param approach.

khelif96 (approved):

PR #553, opened 2024-12-26
DEVPROD-13052 Create error boundary and handle errors on waterfall page

khelif96:
DEVPROD-13052

### Description
Since we use `useSuspenseQuery` any errors need to be handled by an ErrorBoundary. To prevent the errors from propagating up to the central ErrorBoundary that wrap the entire app I created a custom one for just the Waterfall page.
We didn't have designs for this so I came up with one, open to suggestions.

### Screenshots
image

### Testing
Wrote unit tests


apps/spruce/src/pages/waterfall/WaterfallErrorBoundary/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+              </InlineCode>
+              <ButtonsContainer>
+                <Button onClick={this.handleResetPage} variant="primary">
+                  Reset Page
+                </Button>
+              </ButtonsContainer>
+            </InnerContainer>
+          </Container>
+          <TableSkeleton
+            data-cy="waterfall-skeleton"
```
sophstad, line 92: I think disabling animations here may make sense, to me the motion implies loading but there's nothing incoming to render.
```suggestion
            data-cy="waterfall-skeleton"
            enableAnimations={false}
```

apps/spruce/src/pages/waterfall/WaterfallErrorBoundary/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+                Error: {error?.message ?? "An unexpected error has occurred."}
+              </InlineCode>
+              <ButtonsContainer>
+                <Button onClick={this.handleResetPage} variant="primary">
+                  Reset Page
+                </Button>
+              </ButtonsContainer>
+            </InnerContainer>
+          </Container>
+          <TableSkeleton
```
sophstad, line 91: Can we pull out this skeleton into a separate component to reuse with the loading state?

apps/spruce/src/pages/waterfall/WaterfallErrorBoundary/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+        <div>
+          <Container>
+            <InnerContainer>
+              <H1>Oops! Something went wrong.</H1>
+              <InlineCode>
+                Error: {error?.message ?? "An unexpected error has occurred."}
+              </InlineCode>
+              <ButtonsContainer>
+                <Button onClick={this.handleResetPage} variant="primary">
+                  Reset Page
```
sophstad, line 86: Instead of "Reset Page," I think something like "Return to waterfall" may be a little more user friendly

apps/spruce/src/pages/waterfall/WaterfallErrorBoundary/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+      return (
+        <div>
+          <Container>
+            <InnerContainer>
+              <H1>Oops! Something went wrong.</H1>
+              <InlineCode>
+                Error: {error?.message ?? "An unexpected error has occurred."}
+              </InlineCode>
+              <ButtonsContainer>
+                <Button onClick={this.handleResetPage} variant="primary">
```
sophstad, line 85: Imo an arrow looks nice since this button kinda functions as a link
image
```suggestion
                <Button onClick={this.handleResetPage} variant="primary"                   rightGlyph={<Icon glyph="ArrowRight" />}>
```

apps/spruce/src/pages/waterfall/WaterfallErrorBoundary/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+    }
+    if (redirect && redirectPath) {
+      this.resetState();
+      return <Navigate to={redirectPath} />;
+    }
+    return this.props.children;
+  }
+}
+
+const Container = styled.div`
```
sophstad, line 107: Not sure exactly what rule is causing this overflow, but a horizontal scrollbar is introduced on this page (there wasn't one when I was on the previous waterfall page)
image

sophstad (changes requested):

apps/spruce/src/pages/waterfall/WaterfallErrorBoundary/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+}
+
+interface WaterfallErrorBoundaryState {
+  hasError: boolean;
+  error: Error | null;
+  redirect: boolean;
+  redirectPath?: string;
+}
+
+class WaterfallErrorBoundary extends React.Component<
```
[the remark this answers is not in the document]
khelif96: I'm surprised there isn't a _hooky_ way of creating ErrorBoundarys yet.

apps/spruce/src/pages/waterfall/WaterfallErrorBoundary/index.tsx
khelif96: I think its a combination of the `position:absolute` and the `width: 100%` causing it. Since now it represents 100% of the view window instead of the container.

apps/spruce/src/pages/waterfall/WaterfallErrorBoundary/index.tsx
sophstad, line 1: Using the project select dropdown from this page doesn't work, I think that would be a pretty common use case worth fixing.

apps/spruce/src/pages/waterfall/WaterfallErrorBoundary/index.tsx
```diff
@@ -114,7 +114,7 @@ const Container = styled.div`
   box-sizing: border-box;
   position: absolute;
   background-color: rgba(255, 255, 255, 0.8);
-  width: 100%;
+  width: fill-available;
```
sophstad, line 117: Nice rule 😁 it doesn't work as well in safari lol but what can you do

sophstad (changes requested):

apps/spruce/src/pages/waterfall/WaterfallErrorBoundary/index.tsx
khelif96: :shakes_fist_at_safari: [caniuse](https://caniuse.com/?search=fill-available) says it was supported smh :( Replaced it with vw

apps/spruce/src/pages/waterfall/WaterfallErrorBoundary/index.tsx
khelif96: Fixed!

sophstad (approved):
Not sure why this e2e test is consistently failing across PRs but LGTM!

PR #561, opened 2025-01-06
DEVPROD-12595: Remove transitions for page navigation

sophstad:
DEVPROD-12595

### Description

Remove transition for page navigations. This operation just doesn't represent a transition, which React defines as a non-blocking update. Navigating to the next page should be blocking in the sense that it is of higher priority than interacting with elements on the current page.


khelif96 (approved):
LGTM! 
Navigating to a previously rendered page is still slower than I expected. Do you know if this is a result of some weird cache behavior or just a result of how long it takes to render the page? (Sorry if this is already a known issue just wanted to flag)

https://github.com/user-attachments/assets/ae74de92-0b87-4a34-b470-28862d8e6838

sophstad:
> Navigating to a previously rendered page is still slower than I expected.

One edge case we haven't been able to work around with the Apollo cache is navigating to a page you have already visited, but using the _back_ button. This is because the variables provided to Apollo are different (i.e. `minOrder` is specified instead of `maxOrder`). So ultimately, your example is not using the cache at all. If you were to navigate to the next page again, you wouldn't see a transition state, and it's a lot faster. I would be interested if you have any thoughts around this! 😁 

Also re-requesting a review because I added a little nice-to-have around disabling buttons.

khelif96 (approved):

PR #564, opened 2025-01-08
DEVPROD-13825: Use displayStatusCache instead of displayStatus

sophstad:
DEVPROD-13825

### Description

Use `displayStatusCache` instead of `displayStatus` field. Also use `status` as a fallback for tasks that lack `displayStatusCache`.

### Testing

Updated tests

### Evergreen PR

https://github.com/evergreen-ci/evergreen/pull/8595


minnakt (approved):
haven't been here in a while but it makes sense to me!!

PR #565, opened 2025-01-08
DEVPROD-8669: Move shared styles to lib

minnakt:
DEVPROD-8669

### Description
Moves typography, links, fonts, and global styles to lib directory.

Sorry for number of file changes... most changes are updating snapshots and the import string 🙏


packages/lib/src/components/styles/Link.tsx
```diff
@@ -0,0 +1,30 @@
...
+  line-height: inherit;
+`;
+
+const StyledLink = (props: LinkProps<"a">) => (
+  // @ts-ignore-error: An internal LeafyGreen type causes this error.
+  <Link css={overrideStyles} hideExternalIcon {...props} />
+);
+
+const StyledRouterLink = (props: LinkProps<"span"> & RouterLinkProps) => (
+  // @ts-ignore-error: An internal LeafyGreen type causes this error.
```
sophstad, line 21: This may be preferable in case the error gets fixed
```suggestion
  // @ts-expect-error: An internal LeafyGreen type causes this error.
  <Link css={overrideStyles} hideExternalIcon {...props} />
);

const StyledRouterLink = (props: LinkProps<"span"> & RouterLinkProps) => (
  // @ts-expect-error: An internal LeafyGreen type causes this error.
```

packages/lib/src/components/styles/GlobalStyles.tsx
```diff
@@ -0,0 +1,85 @@
...
+export const resetStyles = css`
+  /* Reset styles, usage recommended by LeafyGreen. */
+  *,
+  *:before,
+  *:after {
+    box-sizing: border-box;
+  }
+`;
+
+const spruceResetStyles = css`
```
sophstad, line 14: My preference would probably be to not move all of these Spruce and Parsley specific styles to lib (unless there's a reason to do so that I'm missing). Instead, they can each import the base `resetStyles` and be used within each application.

packages/lib/src/components/styles/Link.tsx
minnakt: Okay I tried to change this, but it doesn't work unfortunately due to the TypeScript versions mismatching between `lib` and `apps`. lib uses `v5.1.3` and `apps` uses `v5.0.3` and they have a different opinion about what counts as an error. Tried to bump TypeScript but it caused a ton of errors 😞 I opened [DEVPROD-14004](https://jira.mongodb.org/browse/DEVPROD-14004)!

PR #567, opened 2025-01-08
DEVPROD-13931 Ensure mainline commits are defined before trying to get task data

khelif96:
DEVPROD-13931

### Description
Adds an additional check that validates that the mainline commits query has returned data before trying to access it.
### Testing
No longer errors

### DIFF
--- a/apps/spruce/src/hooks/useBreakingTask/index.ts
+++ b/apps/spruce/src/hooks/useBreakingTask/index.ts
@@ -56,8 +56,9 @@ export const useBreakingTask = (taskId: string) => {
       },
     },
   });
-  // @ts-expect-error: FIXME. This comment was added by an automated script.
-  const task = getTaskFromMainlineCommitsQuery(breakingTaskData);
+  const task = breakingTaskData
+    ? getTaskFromMainlineCommitsQuery(breakingTaskData)
+    : undefined;
 
   return {
     task,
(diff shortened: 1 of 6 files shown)


apps/spruce/src/hooks/useBreakingTask/index.ts
```diff
@@ -56,8 +56,9 @@ export const useBreakingTask = (taskId: string) => {
       },
     },
   });
-  // @ts-expect-error: FIXME. This comment was added by an automated script.
-  const task = getTaskFromMainlineCommitsQuery(breakingTaskData);
+  const task = breakingTaskData
+    ? getTaskFromMainlineCommitsQuery(breakingTaskData)
```
sophstad, line 60: Within this util function, do you think calling `reportError` is appropriate? It seems like if that were just removed, this code would function identically to what was deployed.

khelif96: Its not necessary but the schema implies the value can still be undefined since it is nullable. I can add an empty check to satisfy the typescript error but the `reportError` will likely never be called unless it actually encounters an empty value.

sophstad: Yeah, since null is a valid value as per the schema I could see it making sense to not treat that value as an error and just handle it instead. Either works though!

sophstad (approved):

PR #572, opened 2025-01-14
DEVPROD-13254: Only render waterfall task icons when in viewport

minnakt:
DEVPROD-13254

### Description
From what I could tell, Safari was struggling with user inputs (i.e. the filters) when there were a lot of elements on the page. I tried replacing the elements with simpler `<span />`s, and the performance issues persisted. The only solution I found to fixing laggy user inputs was to reduce the number of elements on the page at one time.

Other avenues explored:
- Using css rule [`content-visibility`](https://developer.mozilla.org/en-US/docs/Web/CSS/content-visibility) — this didn't work at all
- Using [`IntersectionObserver`](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) to only render elements in viewport — this worked somewhat but the UX really sucked, choppy scrolling and elements suddenly disappearing and reappearing

In the end the thing that worked the best was adding virtualization. 

### Screenshots
I've also changed it so that the headers are sticky.

https://github.com/user-attachments/assets/5f969799-bcbf-42bb-876d-7cf907c82de6

### Testing
- Tested on Safari for large projects on beta


minnakt:
evergreen refresh

sophstad (commented):
This is interesting, and thank you for the writeup! In my opinion, virtualization feels like a last resort. There are a few reasons:

1. It breaks the functionality for Command-F in order to search for build variants. I know that the filters could replace this, but we've seen often that people like native search to work. It also makes sense that a user may want to jump down the page, not remove results.
2. [DEVPROD-10277](https://jira.mongodb.org/browse/DEVPROD-10277) is an unresolved issue around react-virtuoso breaking the browser. I think breaking this page for users would be unacceptable. Could be worth following up with these users to see if they still experience this, and maybe making sure the dependency is up to date.
3. Safari represents a [tiny portion of users](https://ui.honeycomb.io/mongodb-4b/environments/production?query=%7B%22time_range%22%3A604800%2C%22granularity%22%3A0%2C%22breakdowns%22%3A%5B%22browser.name%22%5D%2C%22calculations%22%3A%5B%7B%22op%22%3A%22COUNT_DISTINCT%22%2C%22column%22%3A%22user.id%22%7D%5D%2C%22filters%22%3A%5B%7B%22column%22%3A%22library.name%22%2C%22op%22%3A%22%3D%22%2C%22value%22%3A%22analytics%22%7D%5D%2C%22filter_combination%22%3A%22AND%22%2C%22orders%22%3A%5B%7B%22column%22%3A%22user.id%22%2C%22op%22%3A%22COUNT_DISTINCT%22%2C%22order%22%3A%22descending%22%7D%5D%2C%22havings%22%3A%5B%5D%2C%22trace_joins%22%3A%5B%5D%2C%22limit%22%3A1000%7D) (~4%). I'm not sure whether this change makes sense for that amount of users given the above limitations.

So those are my thoughts. My preference for now would be to leave this unmerged and circle back; this isn't urgent. I have a few more thoughts that could aid this investigation that I'll leave here:

- It's interesting that the legacy waterfall is zippy on Safari; could there be any clues there?
- LeafyGreen just released a [major version of Popover](https://github.com/mongodb/leafygreen-ui/blob/main/packages/popover/CHANGELOG.md#1200) that changes how LeafyGreen overlays work. Maybe upgrading could help the filter performance?

Thanks for your work on this! Maybe virtualization will end up being the right way to go, but I'm not 100% convinced yet.

minnakt:
Thank you Sophie for the review — I implemented a new method, with a summary below:
* No use of `react-virtuoso` to prevent FireFox bugs
* Build variant labels are always visible in the DOM, making it possible to use Ctrl + F
* Use `IntersectionObserver` along with pre-calculating div heights so that the scrolling isn't choppy. This is how the Project Health page currently works
* I used a Context for now but not sure I want to keep that (or I can add more things to it).
* Kept sticky version labels — let me know if you don't like them

apps/spruce/src/hooks/useDimensions/index.ts
```diff
@@ -11,14 +11,17 @@ interface DimensionState {
...
+export const useDimensions = <T extends HTMLElement>(
+  ref: MutableRefObject<T> | React.RefObject<T>,
+) => {
   const [state, setState] = useState<DimensionState>({ width: 0, height: 0 });
 
   const observer = useMemo(
     () =>
       new ResizeObserver((entries) => {
         requestAnimationFrame(() => {
-          const { height, width } = entries[0]?.contentRect ?? {};
```
minnakt, line 21: removed because [`contentRect`](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserverEntry/contentRect) is a legacy property that may be deprecated in the future

apps/spruce/src/pages/waterfall/BuildVariantContext.tsx
```diff
@@ -0,0 +1,44 @@
...
+const BuildVariantContext = createContext<BuildVariantState | null>(null);
+
+interface BuildVariantProviderProps {
+  children: React.ReactNode;
+}
+
+const BuildVariantProvider: React.FC<BuildVariantProviderProps> = ({
+  children,
+}) => {
+  const [columnWidth, setColumnWidth] = useState(0);
```
sophstad, line 17: Maybe a silly question but how does width come into play here? I would think height is the focus but obviously what you're doing is working... worth leaving a comment perhaps!

apps/spruce/src/pages/waterfall/BuildRow.tsx
```diff
@@ -110,38 +136,67 @@ const BuildGrid: React.FC<{
...
+          );
+        },
+      )}
+    </BuildContainer>
+  );
+};
+
+const squareWithBorder = SQUARE_SIZE + 2;
+const containerPadding = 16 + 2;
+const containerBorder = 2;
```
sophstad, line 180: Super nit but would you mind breaking out these magic vars further? This works nicely but it just scares me... e.g.

```js
const padding = spacing[200];
const border = 1;
const containerPadding = (spacing * 2) + (border * 2);
...
const buildGroupCss = css`
  ...
  border: ${border}px solid ${gray.light2};
  padding: ${padding}px;
`

File: apps/spruce/src/pages/waterfall/BuildRow.tsx [hunk 4]
```diff
@@ -110,38 +136,67 @@ const BuildGrid: React.FC<{
...
+}> = ({ build, handleTaskClick, isRightmostBuild }) => {
+  const rowRef = useRef<HTMLDivElement>(null);
+  const { width } = useDimensions<HTMLDivElement>(rowRef);
+
+  const { columnWidth, setColumnWidth } = useBuildVariantContext();
+  useEffect(() => {
+    if (width !== 0 && columnWidth !== width) {
+      setColumnWidth(width);
+    }
+  }, [setColumnWidth, columnWidth, width]);
```
sophstad, line 148: Might we include this `useEffect` within `useBuildVariantContext`? It seems fairly self-contained

apps/spruce/src/pages/waterfall/BuildRow.tsx
```diff
@@ -75,7 +94,14 @@ export const BuildRow: React.FC<Props> = ({
           {displayName}
         </StyledLink>
       </BuildVariantTitle>
-      <BuildGroup data-cy="build-group">
+      <VisibilityContainer
```
sophstad, line 97: This is awesome, I think it also really  improves the speed of the initial page render 👏👏

sophstad (commented):
I am so pleased with this approach! I am glad that we are not building on top of a dependency that presents some risk to the project, it works great, and doesn't feel over-engineered. Really nice job 😍

apps/spruce/src/pages/waterfall/BuildRow.tsx
minnakt: i couldnt figure out how to get `width` in the context! 😖 let me know if im missing something 🥺

apps/spruce/src/pages/waterfall/WaterfallGrid.tsx
```diff
@@ -159,13 +161,18 @@ export const WaterfallGrid: React.FC<WaterfallGridProps> = ({
...
 const Container = styled.div``;
 
-const StickyHeader = styled(Row)`
+const StickyHeader = styled(Row)<{ atTop: boolean }>`
   position: sticky;
   top: -${size.m};
   z-index: 1;
 
   background: white;
   padding: ${size.xs} 0;
```
sophstad, line 170: If you want to extend the border all the way to the edge it seems like this works
```suggestion
  margin-left: -${size.m};
  margin-right: -${size.m};
  padding: ${size.xs} ${size.m;
```

apps/spruce/src/pages/waterfall/WaterfallGrid.tsx
```diff
@@ -159,13 +161,18 @@ export const WaterfallGrid: React.FC<WaterfallGridProps> = ({
...
+  ${({ atTop }) =>
+    atTop
+      ? "box-shadow: unset"
+      : "box-shadow: 0 2px 2px -2px rgba(0, 0, 0, 0.5); "}
```
sophstad, line 174: [nit] I know this matches settings header shadow but I think a little extra looks good here
```suggestion
      : "box-shadow: 0 4px 4px -4px rgba(0, 0, 0, 0.5); "}
```

apps/spruce/src/pages/projectSettings/index.tsx
```diff
@@ -52,23 +51,8 @@ const ProjectSettings: React.FC = () => {
...
-      } else {
-        setAtTop(false);
-      }
-    }, 250);
-    pageWrapperRef?.current?.addEventListener("scroll", onScroll);
-
-    const wrapper = pageWrapperRef.current;
-    return () => wrapper?.removeEventListener("scroll", onScroll);
-  }, []); // eslint-disable-line react-hooks/exhaustive-deps
+  const { atTop } = useHeaderScroll(pageWrapperRef, 40);
```
sophstad, line 55: Yay!

sophstad (approved):

apps/spruce/src/pages/waterfall/WaterfallGrid.tsx
minnakt: omg yay!

PR #578, opened 2025-01-16
DEVPROD-11891 Capture generic route names for spruce routes in otel traces

khelif96:
DEVPROD-11891

### Description
This PR adds generic path names to routes in OTEL traces so you can filter for events based on a general URL instead of one with ids in it. This will enable more powerful queries where we can get holistic views on general pages instead of super-specific ones with ids embedded. 
* Updated react-router-dom so that `matchPath` supports optional query variables
* Update routes to include tab slugs as optional values.

### Screenshots
image

### Testing
[Example query 
](https://ui.honeycomb.io/mongodb-4b/environments/staging/datasets/spruce/result/dGHGYsAxSwt?vs=hideCompare)


sophstad (commented):
> ### Testing
>
> Example query

Does this mean you meant to add one for us to examine? :)

khelif96:
> > ### Testing
> > Example query
> 
> Does this mean you meant to add one for us to examine? :)

Yup oops I swear I pasted a link in there

packages/lib/src/utils/observability/ReactRouterSpanProcessor/index.ts
```diff
@@ -0,0 +1,35 @@
...
+  }
+
+  onStart(span: Span): void {
+    const matchedRoute = calculateRouteName(
+      window.location.pathname,
+      this.routeConfig,
+    );
+    if (matchedRoute) {
+      span.setAttribute("page.route_name", matchedRoute.name);
+      span.setAttribute("page.route", matchedRoute.route);
```
sophstad, line 20: I'm not 100% sure that overwriting `page.route` is a good choice here. I think it could still be useful to have the distinct route available as part of the trace—or is it available elsewhere and I'm missing it?

[Here's](https://ui.honeycomb.io/mongodb-4b/environments/production/datasets/spruce/board-query/EZxx3bmQsKx/result/wi7gXtUiPyh) a query where I'm using it for grouping. Without a project-specific route, we'd probably need to group on a different field according to each query which is just a bit more cumbersome.

What do you think of creating a new attribute for this value instead?

packages/lib/src/utils/observability/utils.ts
```diff
@@ -22,4 +22,5 @@ const detectGraphqlQuery = (
   }
   return undefined;
 };
+
```
sophstad, line 25: [nit]
```suggestion
```

packages/lib/src/utils/observability/ReactRouterSpanProcessor/index.ts
khelif96: overwriting `page.route` actually enables some powerful functionality in honeycomb that wasn't possible before. 
We can now use the Home dashboard to get page level metrics for the web vitals values. 
image

I added the ability to isolate param values as honeycomb attributes so you can perform the same query you had above but get even more granular project level metrics.
https://ui.honeycomb.io/mongodb-4b/environments/staging/datasets/spruce/result/mACExmv7JGq?vs=hideCompare
image

> or is it available elsewhere and I'm missing it?
The original value still exists as `url.path`

sophstad: Ohhh this explanation is everything! Thanks a lot!

sophstad (approved):

sophstad:
If you wanted, I could see this making sense as a minor release.

PR #579, opened 2025-01-17
DEVPROD-10513 Make new project vars private by default

khelif96:
DEVPROD-10513

### Description
Automatically sets new project vars to be private when creating them.
### Screenshots
image


apps/spruce/src/gql/generated/types.ts
```diff
@@ -234,14 +234,12 @@ export type CommitQueueParams = {
   __typename?: "CommitQueueParams";
   enabled?: Maybe<Scalars["Boolean"]["output"]>;
   mergeMethod: Scalars["String"]["output"];
-  mergeQueue: MergeQueue;
```
Kimchelly, line 237: This looks like it's unrelated?

Kimchelly (approved):
LGTM mod is this supposed to have the other unrelated deletions?

apps/spruce/src/gql/generated/types.ts
khelif96: These are just updates to a generated file. (It will make the waterfall green again)

PR #597, opened 2025-01-27
DEVPROD-14495: Clean up waterfall types

sophstad:
DEVPROD-14495

### Description

Ahead of some changes to the GraphQL schema to better support caching in DEVPROD-12696, clean up waterfall types. Using a few types that are not inferred from generated types will allow us to more easily make changes and ensure that the required data is still present.

### DIFF
--- a/apps/spruce/src/pages/waterfall/types.ts
+++ b/apps/spruce/src/pages/waterfall/types.ts
@@ -1,5 +1,4 @@
-import { Unpacked } from "@evg-ui/lib/types/utils";
-import { WaterfallVersionFragment, WaterfallQuery } from "gql/generated/types";
+import { WaterfallVersionFragment } from "gql/generated/types";
 
 // Although this is pretty much a duplicate of the GraphQL type, it is
 // necessary to resolve type errors.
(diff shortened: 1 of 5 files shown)


apps/spruce/src/pages/waterfall/types.ts
```diff
@@ -10,13 +9,23 @@ export type WaterfallVersion = {
...
+    id: string;
+    status: string;
+  }>;
+  version: string;
+};
 
-export type BuildVariant = Unpacked<
-  WaterfallQuery["waterfall"]["buildVariants"]
->;
+export type BuildVariant = {
```
minnakt, line 24: I think we should use these types to type the [`waterfall`](https://github.com/evergreen-ci/ui/blob/ff2db150784ffeba8510a6805dbfdb333ef5fee7/apps/spruce/src/pages/waterfall/useFilters.test.tsx#L424) variable in `useFilters.test.tsx` 👀 I think it would prompt us to delete the `activated`, `displayName`, and `version` fields there

minnakt (approved):

PR #599, opened 2025-01-29
DEVPROD-14386: Add button to scroll to top of waterfall

minnakt:
DEVPROD-14386

### Description
Adds button to jump to the top of the waterfall. I wasn't really sure where to put it, but I feel like the average user would expect it to be in the bottom right corner.

Also updates floating content styles according to designs.

### Screenshots

https://github.com/user-attachments/assets/070c43b0-27e8-4842-9cbd-88481e5a05f2


apps/spruce/src/pages/waterfall/ScrollToTop/index.tsx
sophstad, line 1: It would be cool to make this generic at some point but we can do that when it's needed. There'd have to be some way of passing in the ID I suppose

apps/spruce/src/components/Content/Layout.tsx
```diff
@@ -57,26 +58,36 @@ export const Layout: React.FC = () => {
       <FloatingContent>
         <TaskStatusIconLegend />
         <Feedback />
+        <WaterfallScrollToTop />
```
sophstad, line 61: For some reason I feel like this belongs at the top. That may make it more noticeable
```suggestion
        <WaterfallScrollToTop />
        <TaskStatusIconLegend />
        <Feedback />
```

apps/spruce/src/pages/waterfall/ScrollToTop/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+import { routes } from "constants/routes";
+
+export const WaterfallScrollToTop: React.FC = () => {
+  const isWaterfallPage = !!useMatch(`${routes.waterfall}/*`);
+
+  return isWaterfallPage ? (
+    <IconButton
+      aria-label="Scroll to top"
+      onClick={() =>
+        document.getElementById("waterfall-page")?.scrollTo({ top: 0 })
```
sophstad, line 13: Works great and I like how there's no dizzying animation

sophstad (approved):

PR #603, opened 2025-01-29
DEVPROD-14335 Remove error logging for missing build variants

khelif96:
DEVPROD-14335

### Description
This error has been very noisy and occurs because sometimes tasks just don't have a last passing version especially if they are newly created so this query will silently error in the background. I removed the error log since this is an acceptable scenario.
* Also fixed a random ref error I saw while I was investigating this also
### Testing
No more errors in the console

### DIFF
--- a/apps/spruce/src/utils/getTaskFromMainlineCommitsQuery.ts
+++ b/apps/spruce/src/utils/getTaskFromMainlineCommitsQuery.ts
@@ -23,7 +23,6 @@ export const getTaskFromMainlineCommitsQuery = (
   }
   const buildVariant = buildVariants[0];
   if (buildVariant === null || buildVariant === undefined) {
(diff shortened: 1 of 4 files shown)


minnakt (approved):
sgtm

apps/spruce/src/utils/getTaskFromMainlineCommitsQuery.ts
```diff
@@ -23,7 +23,6 @@ export const getTaskFromMainlineCommitsQuery = (
   }
   const buildVariant = buildVariants[0];
   if (buildVariant === null || buildVariant === undefined) {
-    reportError(new Error("buildVariant is undefined")).warning();
```
sophstad, line 26: Is there any reason not to remove all of the `reportError` instances above? These cases are handled; I'm not sure what value we get from logging.

PR #604, opened 2025-01-30
DEVPROD-10203: Redirect links to waterfall if beta feature is enabled

minnakt:
DEVPROD-10203

### Description
Redirect all links to waterfall page if user is opted in to beta test. I think this is safe to merge and that I just need to enable the admin flag on Monday.

I changed the "New UI" preferences route and can file a ticket for deleting the redirect but I want some feedback before I do 🤔


apps/spruce/src/pages/waterfall/TaskStatsTooltip.tsx
```diff
@@ -48,24 +48,26 @@ export const TaskStatsTooltip: React.FC<
       >
         <PopoverContainer data-cy="task-stats-tooltip">
           <Table>
-            {taskStatusStats?.counts?.map(({ count, status }) => (
-              <Row key={`task_stats_row_${status}`}>
-                <Count>{count}</Count>
-                <Cell>
-                  <Square status={status as TaskStatus} />
+            <Tbody>
```
minnakt, line 51: Added to fix HTML error in console

apps/spruce/src/components/Header/Navbar.tsx
```diff
@@ -75,13 +79,24 @@ export const Navbar: React.FC = () => {
...
           <StyledAnimatedIcon icon={HolidayTree} />
         </LogoLink>
-        <PrimaryLink
-          data-cy="project-health-link"
-          onClick={() => sendEvent({ name: "Clicked project health link" })}
-          to={getCommitsRoute(projectIdentifier)}
-        >
-          Project Health
-        </PrimaryLink>
+        {spruceWaterfallEnabled ? (
```
sophstad, line 82: Can we add the inverse link to the navbar Auxiliary Dropdown? i.e. project health if a user is in the beta and waterfall if not.

apps/spruce/src/pages/preferences/preferencesTabs/UISettingsTab.tsx
sophstad, line 1: Name change SGTM!

sophstad (changes requested):

apps/spruce/src/components/Header/Navbar.tsx
minnakt: good idea! added!
i additionally gated it on the `showWaterfallPage` condition b/c I don't think we should show the links if we haven't released the project yet... lmk if that doesnt make sense lol

apps/spruce/src/components/Header/AuxiliaryDropdown.tsx
```diff
@@ -19,6 +22,23 @@ export const AuxiliaryDropdown: React.FC<AuxiliaryDropdownProps> = ({
...
+  const inverseLink = spruceWaterfallEnabled
+    ? {
+        "data-cy": "auxiliary-dropdown-project-health",
+        text: "Project Health",
+        to: getCommitsRoute(projectIdentifier),
+        onClick: () => sendEvent({ name: "Clicked project health link" }),
+      }
+    : {
+        "data-cy": "auxiliary-dropdown-waterfall",
+        text: "Waterfall",
```
sophstad, line 37: What do you think of something a bit more eye catching? Would just need to update the `text` prop to be React.ReactNode
image

```suggestion
        text: (
          <span>
            Waterfall{" "}
            <Badge darkMode variant={Variant.Blue}>
              Beta
            </Badge>
          </span>
        ),
```

apps/spruce/src/components/Header/AuxiliaryDropdown.tsx
```diff
@@ -49,6 +68,8 @@ export const AuxiliaryDropdown: React.FC<AuxiliaryDropdownProps> = ({
       to: getProjectSettingsRoute(projectIdentifier),
       onClick: () => sendEvent({ name: "Clicked project settings link" }),
     },
+    // We shouldn't show any inverse links if the waterfall page hasn't been released.
+    ...(showWaterfallPage ? [inverseLink] : []),
```
sophstad, line 72: I _think_ instead of this condition we probably want to use the result of `useAdminBetaFeatures`? That will allow flipping the setting to change this instead of having to remove this value from code.

apps/spruce/src/components/Header/AuxiliaryDropdown.tsx
minnakt: did you want the beta badge on the primary link as well? i wasn't sure :O i can just remove if not!

sophstad: Sorry, not to be too nitpicky but imo no. I feel like the one in the dropdown menu might function as a bit of a call to action/intrigue for users who haven't opted in yet.

apps/spruce/src/pages/preferences/preferencesTabs/uiSettingsTab/BetaFeatures.tsx
sophstad, line 1: Could we add a link to UI Settings in the user dropdown to aid discoverability?

sophstad (approved):

PR #605, opened 2025-01-31
DEVPROD-14591: Remove task stats from waterfall query

sophstad:
DEVPROD-14591

### Description

Move `Version/taskStatusStats` query into the task stats component. This ensures that only active versions have their stats queried for and speeds up the main waterfall request.

### Testing

Combine redundant snapshots


khelif96 (approved):
🔥

PR #607, opened 2025-01-31
DEVPROD-14608 Extract upstream project into its own query 

khelif96:
DEVPROD-14608

### Description
This can be a slow query so this PR extracts it into its own component and performs the query after the initial waterfall loads.


sophstad (approved):
LGTM once snapshot is fixed!

PR #609, opened 2025-02-03
DEVPROD-14608: Remove upstream from waterfall query

sophstad:
DEVPROD-14608

### Description
#607 forgot to remove the `UpstreamProject` fragment from the waterfall query


khelif96 (approved):
oops good catch!

PR #617, opened 2025-02-05
DEVPROD-14722 Embed app version in spans and fix missing route names

khelif96:
DEVPROD-17222

### Description
While looking at Honeycomb, I noticed some strange event names and realized we don't maintain stats on which app version our users are using. This PR ensures we log the app version.
I also noticed that sometimes routes don't get properly generalized so this PR fixes it.

### Screenshots
image

### DIFF
--- a/apps/parsley/src/main.tsx
+++ b/apps/parsley/src/main.tsx
@@ -16,6 +16,7 @@ import routes from "./constants/routes";
 
 initializeErrorHandling();
 initializeHoneycomb({
+  appVersion: process.env.REACT_APP_VERSION || "",
   backendURL: toEscapedRegex(evergreenURL || ""),
   debug: isDevelopmentBuild(),
   endpoint: process.env.REACT_APP_HONEYCOMB_ENDPOINT || "",
@@ -26,6 +27,7 @@ initializeHoneycomb({
 });
 injectOpenTelemetryAttributeStoreIntoWindow();
 
+console.log(process.env.REACT_APP_VERSION);
 ReactDOM.createRoot(document.getElementById("root") as HTMLElement).render(
   <React.StrictMode>
     <App />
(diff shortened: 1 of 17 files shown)


apps/parsley/src/main.tsx
```diff
@@ -12,16 +12,21 @@ import {
   isDevelopmentBuild,
 } from "utils/environmentVariables";
 import App from "./App";
-import routes from "./constants/routes";
+import routes, { slugs } from "./constants/routes";
 
+const routeConfig = {
+  ...routes,
+  testLogs: `${routes.testLogs}/:${slugs.groupID}?`,
```
sophstad, line 19: Why do we want to overwrite this value for Honeycomb? Maybe worth leaving a comment.

apps/spruce/src/constants/routes.ts
sophstad, line 1: Curious about the reason behind removing the named types in this file, do you just find it more readable?

packages/lib/src/utils/observability/honeycomb.ts
```diff
@@ -94,10 +98,11 @@ const initializeHoneycomb = ({
         instrumentations: [
           getWebAutoInstrumentations(webAutoInstrumentationConfig),
         ],
-        // Add user.id as an attribute to all traces.
+        // Add attributes attribute to all traces.
```
sophstad, line 101:
```suggestion
        // Add attributes to all traces.
```

apps/spruce/src/constants/routes.ts
khelif96: It makes IntelliSense better :) 
Before:
image
After:
image

apps/parsley/src/main.tsx
khelif96: It helps with making sure we can identify URLs with the groupID slugs. Otherwise, the ReactRouterSpanProcessor just thinks its a different URL entirely.

sophstad (approved):

PR #618, opened 2025-02-05
DEVPROD-13933: Add custom cache

sophstad:
DEVPROD-13933

### Description

- Implement custom read/merge functions to leverage the Apollo cache for the waterfall page
- Disable normalization for `WaterfallTask` types. This significantly improved the render speed when navigating to a page of cached data.

### Testing

Add navigation test to Cypress

### Evergreen PR
https://github.com/evergreen-ci/evergreen/pull/8701


SupaJoon (changes requested):
Can we have some unit tests for the read and merge functions?

apps/spruce/src/gql/client/cache.ts
```diff
@@ -17,6 +18,144 @@ export const cache = new InMemoryCache({
...
+            if (idx === -1) {
+              return undefined;
+            }
+
+            let startIndex = maxOrder ? idx : 0;
+            let endIndex = maxOrder ? existingVersions.length : idx;
+            let numActivated = 0;
+
+            // Count backwards for paginating backwards.
+            if (minOrder) {
```
minnakt: something kind of funky about the cache is that if you have some versions that match the `maxOrder`/`minOrder` conditions (but less than limit), it'll just read from the versions in the cache and not fetch more versions

not sure if this would have to be addressed by [DEVPROD-11707](https://jira.mongodb.org/browse/DEVPROD-11707) or if it's just too difficult to fix 😞

https://github.com/user-attachments/assets/1515985c-219b-4d08-8af3-03c2d040120e

sophstad: Thanks, good catch. When this happens we just want to initiate a new request via returning undefined; better to fetch eagerly than have weirdness. I added handling and some tests.

SupaJoon (approved):

apps/spruce/src/pages/waterfall/utils.test.ts
minnakt, line 1: could we put the tests in `caching.test.ts`? I think I would forget where they were, and I'd expect `utils.test.ts` to only contain the tests for `utils.ts`

could also possibly make a caching folder if it seems appropriate

apps/spruce/src/pages/waterfall/utils.test.ts
```diff
@@ -137,6 +139,276 @@ describe("groupBuildVariants", () => {
   });
 });
 
+describe("caching functions", () => {
+  describe("mergeVersions", () => {
```
minnakt, line 143: could we pull the `readField` definition up so we only have to write ts-expect-error and the function definition once at the top?
```
  // @ts-expect-error
  const readField: ReadFieldFunction = (field, obj) => obj[field];
```

apps/spruce/src/pages/waterfall/utils.test.ts
```diff
@@ -137,6 +139,276 @@ describe("groupBuildVariants", () => {
...
+          },
+          {
+            // @ts-expect-error
+            readField: (field, obj) => obj[field],
+          } as FieldFunctionOptions,
+        ),
+      ).toStrictEqual({ flattenedVersions: versions, pagination });
+    });
+  });
+
```
minnakt, line 240: Can we add a test for leading inactive versions?

apps/spruce/src/pages/waterfall/caching.ts
```diff
@@ -0,0 +1,139 @@
...
+    if (numActivated < limit) {
+      return undefined;
+    }
+  }
+
+  // Add 1 because slice is [inclusive, exclusive).
+  const flattenedVersions = existingVersions.slice(startIndex, endIndex + 1);
+
+  const prevOrderNumber =
+    startIndex === 0
```
minnakt, line 87: I think checking if the `startIndex` is 0 doesn't work if you start on a page that isn't the first page; for example, starting at http://localhost:3000/project/evergreen-ui/waterfall?maxOrder=1098

i'm guessing we'd probably need to return something like the `mostRecentCommitOrder` from the backend

sophstad: https://github.com/evergreen-ci/evergreen/pull/8701

apps/spruce/src/pages/waterfall/utils.test.ts
sophstad: Since `id: a` is inactive, I think this is covered by "reads the first page and returns limit active versions"? But lmk if you had something else in mind!

apps/spruce/src/pages/waterfall/caching/index.ts
```diff
@@ -0,0 +1,145 @@
+import { FieldMergeFunction, FieldReadFunction } from "@apollo/client";
+import { WaterfallQuery } from "gql/generated/types";
+import { VERSION_LIMIT } from "../constants";
+
+export const readVersions = ((existing, { args, readField }) => {
```
minnakt, line 5: i might be doing something wrong but i think the cache might not be working! i keep getting the loading screen between pagination. maybe the new field also needs to be returned in the read function?

sophstad: duh, added

apps/spruce/src/pages/waterfall/caching/index.ts
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  // Add 1 because slice is [inclusive, exclusive).
+  const flattenedVersions = existingVersions.slice(startIndex, endIndex + 1);
+
+  const zerothOrder = readField<number>("order", flattenedVersions[0]) ?? 0;
+  const prevOrderNumber =
+    mostRecentVersionOrder === zerothOrder
+      ? 0
+      : (readField<number>("order", flattenedVersions[0]) ?? 0);
```
minnakt, line 95: would this be `zerothOrder`?
```suggestion
      : zerothOrder;
```

apps/spruce/src/pages/waterfall/caching/index.ts
```diff
@@ -0,0 +1,146 @@
...
+  const lastVersionOrder = readField<number>(
+    "order",
+    flattenedVersions[flattenedVersions.length - 1],
+  );
+  const nextOrderNumber =
+    !lastVersionOrder || lastVersionOrder === 1 ? 0 : lastVersionOrder;
```
minnakt, line 101: i think you can do a small simplification here:
```suggestion
  const lastVersionOrder =
    readField<number>(
      "order",
      flattenedVersions[flattenedVersions.length - 1],
    ) ?? 0;
  const nextOrderNumber = lastVersionOrder === 1 ? 0 : lastVersionOrder;
```

apps/spruce/src/pages/waterfall/caching/index.ts
```diff
@@ -0,0 +1,146 @@
...
+              startIndex = 0;
+            }
+          }
+          break;
+        }
+      }
+    }
+    if (numActivated < limit) {
+      return undefined;
+    }
```
minnakt, line 69: I wonder if you need this one 👀 i think even if you return `undefined`, if the `minOrder` bound is too high, it won't change the results when you re-query

minnakt (approved):
lgtm i think :)

apps/spruce/src/pages/waterfall/caching/index.ts
sophstad: I tightened up the typing of the caching functions so that future similar changes would not slip through the cracks 👍

PR #621, opened 2025-02-10
DEVPROD-12399: Allow fetching Parsley filters for projects without projectIdentifier

SupaJoon:
DEVPROD-12399

### Description
These code changes allow fetching Parsley filters with the project ID instead of projectIdentifier. This allows fetching repo projects that don't have a projectIdentifier. The Jira ticket has a link to an example project.


apps/parsley/src/gql/queries/repo-filters.graphql
```diff
@@ -0,0 +1,12 @@
+query RepoFilters($repoRefId: String!) {
+  repoSettings(repoId: $repoRefId) {
```
minnakt, line 2: one question i have is regarding permissions -- `repoSettings.projectRef.parsleyFilters` (requires project settings permissions) is stricter than the `project.parsleyFilters` (requires only basic task permissions). will that be a problem?

minnakt (changes requested):

apps/parsley/src/gql/queries/repo-filters.graphql
SupaJoon: @minnakt I think this will probably be an issue. What do you think about temporarily using the task permission for the `repoSettings` query and then sorting out the permission further in a followup ticket like we do for the project query in [DEVPROD-6882](https://jira.mongodb.org/browse/DEVPROD-6882)? `RepoSettingsInput` requires project settings permissions so at least writing is secure.

minnakt: we could also keep `repoSettings` query locked down and make a new query like `repo` that just returns the reporef (instead of returning all the settings too)

it would be similar to how we have `projectSettings` query but also `project` query. what do you think 👀

minnakt (commented):
thanks for adding test! i left a comment above but it doesn't register as a review lol

apps/parsley/src/gql/queries/repo-filters.graphql
SupaJoon: @minnakt that sounds good to me! I will make a note in [DEVPROD-6882](https://jira.mongodb.org/browse/DEVPROD-6882) to update the permission for the `repo` query.

SupaJoon:
This the repo query: https://github.com/evergreen-ci/evergreen/pull/8717

apps/parsley/src/components/ProjectFiltersModal/index.tsx
```diff
@@ -35,13 +35,13 @@ const ProjectFiltersModal: React.FC<ProjectFiltersModalProps> = ({
...
-  const { projectIdentifier = "" } = versionMetadata ?? {};
+  const { projectMetadata } = versionMetadata ?? {};
 
   const { data } = useQuery<ProjectFiltersQuery, ProjectFiltersQueryVariables>(
     PROJECT_FILTERS,
     {
-      skip: !projectIdentifier,
-      variables: { projectIdentifier },
+      skip: !projectMetadata?.id,
+      variables: { projectIdentifier: projectMetadata?.id ?? "" },
```
minnakt, line 44: Can we update the query to take `projectId` so this looks more natural? (not the underlying query itself, but the wrapper definition we have for the query in the GraphQL file)

minnakt (approved):

PR #626, opened 2025-02-11
DEVPROD-14858: Apply lint rules from `@typescript-eslint`

minnakt:
DEVPROD-14858

### Description
Sadly it looks like we weren't extending these rules when we were supposed to be 🥹

I'd like to add them in preparation for [DEVPROD-8866](https://jira.mongodb.org/browse/DEVPROD-8866).


minnakt:
evergreen retry

sophstad (approved):

PR #628, opened 2025-02-12
DEVPROD-14905: Remove `eslint-config-airbnb`

minnakt:
DEVPROD-14905

### Description
Remove `airbnb`'s ESLint rules. I also fixed some warnings just to lower our warnings count.

Reasons to remove `airbnb`:
* They don't support flat config.
* `airbnb`'s config overwrites rules from other ESLint plugins ([example](https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/imports.js)), making it difficult to track down what rules are active and why. We should just be explicit about what rules we want enabled / disabled.
* `airbnb` was a good config when it was introduced but over time it's become less necessary as other ESLint libraries have implemented stronger rules.
* We have a lot of other ESLint rules in place, which already promote best practices.


khelif96 (commented):
This is an interesting PR it looks like the primary motivations are support for eslint 9 flat config and to make our configuration easier to understand? Are there [any rules](https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb/rules/react.js) that we are losing from airbnb that we should probably setup on our own or do you think we have enough coverage? I think that this change might warrant an ADR since it does represent a fairly large shift in our apps configuration/ has the potential to influence our coding conventions. What do you think?

PR #632, opened 2025-02-14
DEVPROD-8866: Upgrade to ESLint v9

minnakt:
DEVPROD-8866

### Description
Upgrades to ESLint v9.

Summary of changes:
* Upgrade all ESLint plugins to latest version.
* Keep all ESLint plugins and rules from `main` branch in the config.
* Delete `// eslint-disable` comments that no longer apply. Some examples (not comprehensive):
   * [`no-param-reassign`](https://eslint.org/docs/latest/rules/no-param-reassign) is no longer enforced, which doesn't seem to matter because the rule [doesn't apply in strict mode](https://eslint.org/docs/latest/rules/no-param-reassign#when-not-to-use-it).
   * [`no-constant-condition`](https://eslint.org/docs/latest/rules/no-constant-condition)'s default value is [`"allExceptWhileTrue"`](https://eslint.org/docs/latest/rules/no-constant-condition#checkloops). So `while (true)` is valid.
* I turned the flat config into a composition of separate objects because I thought it could be more readable, but if it's preferable, I can turn it into one bigger object (with comments). 
  
### How to Review
In order to review this PR, I recommend running the command:
```
yarn eslint --inspect-config
```
This will open the Config Inspector that will allow you to see all installed plugins and active rules. The Config Inspector is a new utility only available for flat config.

You should run this command from the `lib`, `parsley`, and `spruce` directories separately since they will show slightly different configs. For example, Spruce has the `check-file` plugin, so you will only see it in the Spruce ESLint config.


packages/eslint-config/index.js
sophstad, line 1: > In order to review this PR, I recommend running the command:
> 
> ```
> yarn eslint --inspect-config
> ```

You should add a README to this directory with that note!

apps/parsley/package.json
```diff
@@ -119,9 +119,9 @@
     "babel-plugin-import-graphql": "2.8.1",
     "cypress": "12.7.0",
     "env-cmd": "10.1.0",
-    "eslint": "8.56.0",
+    "eslint": "9.19.0",
```
sophstad, line 122: Could we add a caret here and to the other packages?

packages/eslint-config/README.md
```diff
@@ -0,0 +1,15 @@
+### About ESLint
+
+Our ESLint configuration uses the flat config system. Currently, this is the recommended way to use ESLint.
+
+### Reviewing ESLint Changes
```
sophstad, line 5: [nit] Preference to start with an H1
```suggestion
# About ESLint

Our ESLint configuration uses the flat config system. Currently, this is the recommended way to use ESLint.

## Reviewing ESLint Changes
```

sophstad (approved):

minnakt:
evergreen retry

minnakt:
evergreen retry PLEASE

minnakt:
evergreen retry

PR #639, opened 2025-02-19
DEVPROD-15032

khelif96:
DEVPROD-15032

### Description
This issue in this slack [thread](https://mongodb.slack.com/archives/C0V896UV8/p1739902944738679) stemmed from confusion around the differences between scheduling tasks in a version and configuring a patch. 
This PR adds a title attribute to the button to hopefully clarify why its disabled and point users to the correct button. 
Also while investigating the issue I noticed the analytics events weren't as useful as I expected so I am adding some more attributes to make them more useful.

### Testing

Have you have updated the analytics documentation if necessary?  
https://docs.google.com/spreadsheets/d/1s4_nq8ZiphXp5Uq_-9HT6GPqz-KOyaq6HuvmXYaSNzg/edit?usp=sharing 

Yes

### DIFF
--- a/apps/spruce/src/components/LinkToReconfigurePage.tsx
+++ b/apps/spruce/src/components/LinkToReconfigurePage.tsx
@@ -22,6 +22,11 @@ export const LinkToReconfigurePage: React.FC<{
           sendEvent({ name: "Clicked patch reconfigure link" });
         }
       }}
+      title={
+        disabled
+          ? "This is not a reconfigurable patch use the schedule button instead to schedule tasks"
+          : ""
+      }
       to={getPatchRoute(patchId, { configure: true })}
     >
       Reconfigure tasks/variants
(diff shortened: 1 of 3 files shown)


apps/spruce/src/components/LinkToReconfigurePage.tsx
```diff
@@ -22,6 +22,11 @@ export const LinkToReconfigurePage: React.FC<{
           sendEvent({ name: "Clicked patch reconfigure link" });
         }
       }}
+      title={
+        disabled
+          ? "This is not a reconfigurable patch use the schedule button instead to schedule tasks"
```
sophstad, line 27:
```suggestion
          ? "This is not a reconfigurable patch. Use the Schedule button instead to schedule tasks."
```

apps/spruce/src/components/LinkToReconfigurePage.tsx
```diff
@@ -22,6 +22,11 @@ export const LinkToReconfigurePage: React.FC<{
           sendEvent({ name: "Clicked patch reconfigure link" });
         }
       }}
+      title={
```
sophstad, line 25: I wouldn't be opposed to a tooltip here but your call!

sophstad (approved):

PR #640, opened 2025-02-19
DEVPROD-12594: Clear order params when reaching the first page

minnakt:
DEVPROD-12594

### Description
I think that the current behavior of the page is technically correct. If the `minOrder` lower bound is in place, it should only show versions above that lower bound. If we were to show more versions, we would not be correctly obeying the `minOrder` parameter because the additional versions would not satisfy the lower bound. 

So in this PR, I clear the order params when the user reaches the first page. This means we'll fetch the first LIMIT versions without having to worry about any bounds.

### Screenshots

https://github.com/user-attachments/assets/f518ff59-5624-46cf-92be-053726df9ec0

### Testing
- Added unit and Cypress tests
- Confirmed it worked on mms when a new commit had come in

### DIFF
--- a/apps/spruce/cypress/integration/waterfall/pagination.ts
+++ b/apps/spruce/cypress/integration/waterfall/pagination.ts
@@ -111,4 +111,15 @@ describe("pagination", () => {
       cy.location("search").should("not.contain", "minOrder");
     });
   });
+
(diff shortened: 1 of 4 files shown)


minnakt:
I made it on the wrong branch... 🥺

PR #641, opened 2025-02-19
DEVPROD-12594: Clear order params when reaching first page of waterfall

minnakt:
DEVPROD-12594

### Description
I think that the current behavior of the query params is technically correct. If the `minOrder` lower bound is in place, it should only show versions above that lower bound. If we were to show more versions, we would not be correctly obeying the `minOrder` parameter because the additional versions would not satisfy the lower bound. 

So in this PR, I clear the order params when the user reaches the first page. This means we'll fetch the first LIMIT versions without having to worry about any bounds.

### Screenshots

https://github.com/user-attachments/assets/f518ff59-5624-46cf-92be-053726df9ec0

### Testing
- Added unit and Cypress tests
- Confirmed it worked on mms when a new commit had come in
- Checked for performance issues on MongoDB master


apps/spruce/src/pages/waterfall/WaterfallGrid.tsx
```diff
@@ -114,6 +115,24 @@ export const WaterfallGrid: React.FC<WaterfallGridProps> = ({
     },
   );
 
+  // If we've gotten to the first page, we should erase any minOrder / maxOrder query params.
+  useEffect(() => {
+    const { flattenedVersions, pagination } = data.waterfall;
+    if (minOrder > 0 && flattenedVersions.length > 0) {
+      const activeVersions = flattenedVersions.filter((v) => v.activated);
+      const mostRecentCommitOnPage =
+        flattenedVersions[0].order === pagination.mostRecentVersionOrder;
```
sophstad, line 124: Difficult, but did you manage to test this IRL with a new version coming in instead of mocking this scenario using the datepicker? I wonder if that would lead to `mostRecentVersionOrder` being out of date, and thus "hiding" the fact that a new version is present.

If this is in fact the case, one idea is that we could set `nextFetchPolicy` to `cache-and-network` to pull down updated data while still leveraging the cache.

sophstad: Or, maybe polling is likely to update `mostRecentVersionOrder` frequently enough that this is a moot issue and we should avoid complicating things. 🤷‍♀️

minnakt: yes I think the `mostRecentVersionOrder` is updated, which makes this work; I just tested this using the following process

* Go to Cloud project and go 1 page forward.
* Wait and observe GitHub repo to see when a new commit has come in.
* Paginate backwards 1 page. Although this used to be the 0th page, a new commit came in, so the previous page button should not be disabled.
* Click the previous page button. The page is populated with all 5 expected commits.

I'll post a video on the ticket since I don't want to show the commits publicly

sophstad (approved):

PR #642, opened 2025-02-19
DEVPROD-15062 Support corp secure hosted Evergreen and proxy traffic over https for local development

khelif96:
DEVPROD-15062

### Description
I want to apologize in advance for how large this PR is. 

This PR does a few things.
1. Consolidate auth and login page logic between Spruce and Parsley and move it to lib.
2. Introduces the concept of a `ProtectedRoute` this simplifies auth since everything that is a protected route should not be visible.
3. Adds support for a corp secure hosted evergreen. Corp secure requires users to authenticate every 12 hours so this will ensure that we correctly redirect users to the correct login flow.
4. Fixes support for local dev server against a remote backend by properly placing the local UI devserver behind a https hosted domain. (This is also a requirement for corpsecure) TLDR for Evergreen devs: We can use chrome for local dev against a remote backend again.
5. Rename `REACT_APP_UI_URL` -> `REACT_APP_EVERGREEN_URL` This just makes it a little clearer to understand and also reflects what we use in Parsley

To do this there is a 1 time set up process
1. Generate certs for the local environment and add them to your certificate trust authority.
2. Edit your `/etc/hosts` file so that the local environment URLs point to local host.
Detailed instructions for each app are found in the readme's below.

### Testing
Wrote extensive tests for new introduced components and hooks, Created new tests for existing untested code,  Existing e2e tests should still pass.


apps/spruce/.env-cmdrc.json
```diff
@@ -15,6 +15,16 @@
     "REACT_APP_UI_URL": "https://evergreen-staging.corp.mongodb.com",
     "REACT_APP_HONEYCOMB_BASE_URL": "https://ui.honeycomb.io/mongodb-4b/environments/staging"
   },
+  "corp-secure-staging": {
```
ybrill, line 18: Could it work to instead switch over the `staging` and `devStaging` environments to the new URL?

apps/parsley/.env-cmdrc.json
```diff
@@ -43,24 +44,29 @@
...
+    "REMOTE_ENV": "false",
     "REACT_APP_EVERGREEN_URL": "http://localhost:9090",
     "REACT_APP_GRAPHQL_URL": "http://localhost:9090/graphql/query",
     "REACT_APP_LOGKEEPER_URL": "http://localhost:8080",
     "REACT_APP_SPRUCE_URL": "http://localhost:3000"
   },
   "devStaging": {
     "NODE_ENV": "development",
+    "REMOTE_ENV": "true",
     "REACT_APP_EVERGREEN_URL": "https://evergreen-staging.corp.mongodb.com",
```
ybrill, line 56: does this also need to be the new URL?

apps/parsley/README.md
```diff
@@ -17,6 +17,36 @@ Parsley is the UI for Evergreen's log viewer. It is the replacement for
...
+```
+
+3. Run the following command to start the local UI server with the remote
+   Evergreen server.
+
+```sh
+yarn <env_name>  # where env_name is the name of the environment you want to run staging or prod
+```
+
+4. Navigate to `https://parsley-local.corp.mongodb.com:8443` in your browser to
```
ybrill, line 47: this is 8444, right?

apps/spruce/src/gql/client/useCreateGQLClient/index.ts
```diff
@@ -17,15 +17,20 @@ import {
...
+export const useCreateGQLClient = ():
+  | ApolloClient<NormalizedCacheObject>
+  | undefined => {
   const { dispatchAuthenticated, logoutAndRedirect } = useAuthDispatchContext();
   const [secretFields, setSecretFields] = useState<string[]>();
-  const [gqlClient, setGQLClient] = useState<any>();
+  const [gqlClient, setGQLClient] =
+    useState<ApolloClient<NormalizedCacheObject>>();
 
+  console.log(window.location.href);
```
ybrill, line 33: this is left over?

apps/spruce/.env-cmdrc.json
```diff
@@ -38,21 +39,27 @@
...
-    "REACT_APP_SPRUCE_URL": "http://localhost:3000",
-    "REACT_APP_UI_URL": "https://evergreen-staging.corp.mongodb.com"
+    "REACT_APP_SPRUCE_URL": "https://spruce-local.corp.mongodb.com",
+    "REACT_APP_UI_URL": "https://evergreen.staging.corp.mongodb.com",
+    "REACT_APP_HONEYCOMB_BASE_URL": "https://ui.honeycomb.io/mongodb-4b/environments/staging",
+    "REACT_APP_CORP_LOGIN_URL": "https://login.corp.mongodb.com/login"
   },
   "devProduction": {
     "NODE_ENV": "development",
+    "REACT_APP_REMOTE_ENV": "true",
```
ybrill, line 58: do we want to wait on this until we have a new corpsecured prod URL?

apps/spruce/src/gql/client/useCreateGQLClient/index.ts
```diff
@@ -17,15 +17,20 @@ import {
...
 import { environmentVariables } from "utils";
+import { getCorpLoginURL, isRemoteEnv } from "utils/environmentVariables";
 import { leaveBreadcrumb, SentryBreadcrumb } from "utils/errorReporting";
 
 const { getGQLUrl } = environmentVariables;
 
-export const useCreateGQLClient = (): ApolloClient<NormalizedCacheObject> => {
+export const useCreateGQLClient = ():
+  | ApolloClient<NormalizedCacheObject>
+  | undefined => {
```
ybrill, line 27: when does it return `undefined`?

khelif96: It is initially undefined until after the secretFields query responds. https://github.com/khelif96/ui/blob/76880083df11136f85ddcaa2305241fb3f16179b/apps/spruce/src/gql/client/useCreateGQLClient/index.ts#L31

apps/spruce/.env-cmdrc.json
khelif96: This is still fine since 1. It technically still is a remote env and it works the same way either way.

apps/parsley/.env-cmdrc.json
```diff
@@ -43,24 +44,29 @@
...
     "REACT_APP_EVERGREEN_URL": "http://localhost:9090",
     "REACT_APP_GRAPHQL_URL": "http://localhost:9090/graphql/query",
     "REACT_APP_LOGKEEPER_URL": "http://localhost:8080",
     "REACT_APP_SPRUCE_URL": "http://localhost:3000"
   },
   "devStaging": {
     "NODE_ENV": "development",
-    "REACT_APP_EVERGREEN_URL": "https://evergreen-staging.corp.mongodb.com",
-    "REACT_APP_GRAPHQL_URL": "https://evergreen-staging.corp.mongodb.com/graphql/query",
+    "REACT_APP_REMOTE_ENV": "true",
```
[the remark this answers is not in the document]
khelif96: I've refactored the logic to rely on Evergreen for authentication, which makes this variable unnecessary. Previously, I ran into issues where the corp-secure authentication redirect didn't always set the cookie correctly when starting from Spruce. About half the time, it would enter an infinite redirect loop: Spruce would redirect to the corp-secure login, then get redirected back to Spruce, and repeat. By directing everything to the `/login` route on Evergreen instead, the cookie is set correctly each time, and this approach has been consistently more reliable in my tests. It also eliminates the need to maintain this URL and we can use the same authentication strategy between spruce and evergreen.

apps/parsley/.env-cmdrc.json
khelif96, line 1: Update environmental variables to add a new PARSLEY_URL variable so that the app can know its own URL. 
Migrate staging to use the new corpsecure evergreen url

apps/parsley/README.md
```diff
@@ -17,6 +17,36 @@ Parsley is the UI for Evergreen's log viewer. It is the replacement for
...
+If you want to run Parsley against a remote Evergreen server, you can do so by
+taking the following steps.
+
+1. Generate a self signed certificate and add it to your keychain. This will
+   allow you to run Parsley against a remote Evergreen server that uses HTTPS.
+
+```sh
+brew install mkcert
+mkcert -install
+mkcert -key-file localhost-key.pem -cert-file localhost-cert.pem parsley-local.corp.mongodb.com
```
sophstad, line 31: May be worth specifying that this command should be run from within the Parsley directory (and same for Spruce)

apps/parsley/src/App.tsx
```diff
@@ -1,28 +1,52 @@
...
 import Content from "pages";
-import { Login } from "pages/Login";
-import { isDevelopmentBuild } from "utils/environmentVariables";
+import { evergreenURL, isLocal } from "utils/environmentVariables";
+
+const router = createBrowserRouter([
+  {
+    children: [
+      {
+        element: <LoginPage />,
```
sophstad, line 17: Sorry if I'm misunderstanding -- I hit logout on Parsley when running `yarn prod` and am shown this page when I go to the site. I'm able to log back in by visiting prod Parsley and then returning to https://parsley-local.corp.mongodb.com:8444/upload, but I find this page to be a little confusing since it will never be actionable when using the remote evergreen server.
image

Am I correct that there's no way to kick off an Okta login flow besides visiting prod? That's fine if so—but maybe we could display something different on this page if the user is attempting to hit a remote server?

apps/spruce/src/App.tsx
```diff
@@ -1,43 +1,50 @@
...
-    </>,
-  ),
-);
+const router = createBrowserRouter([
+  {
+    element: (
+      <AuthProvider
+        evergreenAppURL={getEvergreenUrl()}
+        localAuthRoute={routes.login}
+        remoteAuthURL={`${getEvergreenUrl()}/login`}
```
ybrill, line 19: I'm confused how this would work in the CorpSecure case.

apps/spruce/.env-cmdrc.json
```diff
@@ -1,58 +1,71 @@
 {
   "local": {
     "NODE_ENV": "production",
+    "REMOTE_ENV": "false",
```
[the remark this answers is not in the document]
khelif96: Good call!

apps/parsley/src/App.tsx
khelif96: There is a way to kick off the okta flow by redirecting to a special okta URL but that solution has been flaky at best.  Redirecting to the evergreen `/login` route to perform the login has been the most reliable in my testing. The only issue is that okta will redirect back to spruce-staging instead of back to spruce-local when doing local development. But this should not be a major inconvenience since it only needs to happen once per day.

apps/spruce/src/App.tsx
khelif96: So evergreen already handles the corp secure login portion with [this login/redirect logic](https://github.com/evergreen-ci/evergreen/blob/ceae5aa100d628053b09b71a6ab4f80aabab4e48/service/ui.go#L290).  Instead of inventing our own I ended up switching to just using Evergreens existing implementation. Its been way more reliable in setting the necessary cookies then sending then relying on a UI only redirect solution.

khelif96: It also does a bunch of [other validation steps](https://github.com/evergreen-ci/gimlet/blob/622a429ac45676bc24b5caa3ff6d086c15226f31/okta/okta.go#L431) that aren't possible with just the regular okta redirect flow we were doing before.

ybrill: but isn't that the redirect to Evergreen's Okta auth? (rather than Corpsecure's)

khelif96:
#optibot review

apps/parsley/src/App.tsx
khelif96: That page will no longer render if connected to a remote server!

apps/parsley/vite.config.ts
```diff
@@ -2,20 +2,73 @@
...
+  // Validate the SSL certificates exist
+  if (
+    !fs.existsSync(path.resolve(__dirname, "localhost-key.pem")) ||
+    !fs.existsSync(path.resolve(__dirname, "localhost-cert.pem"))
+  ) {
+    console.error(`
+    *******************************************************************************************************
+    *                                                                                                     *
+    *  ERROR: localhost-[key|cert].pem is missing. Did you run                                            *
+    *  'mkcert -key-file localhost-key.pem -cert-file localhost-cert.pem ${hostURL}'? *
```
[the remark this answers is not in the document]
khelif96: I find this to be more user-friendly since it tells the user exactly whats wrong. Now that this logic has been consolidated it should be more straight forward.

apps/parsley/vite.config.ts
```diff
@@ -2,20 +2,73 @@
...
 import injectVariablesInHTML from "./config/injectVariablesInHTML";
 
 // Remove when https://github.com/cypress-io/cypress/issues/25397 is resolved.
 dns.setDefaultResultOrder("ipv4first");
 
+let serverConfig: ServerOptions = {
+  host: "localhost",
+  port: 5173,
+};
+const isViteInDevMode = process.env.NODE_ENV === "development";
```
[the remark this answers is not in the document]
khelif96: I tried this out and ran into a ts error and didn't see an immediate solution. We can probably do that in https://jira.mongodb.org/browse/DEVPROD-11638 which is slated for this quarter.

ybrill (approved):
can't speak to the whole strategy here, but the auth flow with CorpSecure makes sense to me!

packages/vite-utils/src/index.ts
```diff
@@ -0,0 +1,89 @@
...
+        process.exit(1);
+      }
+    });
+
+    const dirNameOfParentModule = path.dirname(parentModule() || "");
+
+    // Validate the SSL certificates exist
+    if (
+      !fs.existsSync(path.resolve("localhost-key.pem")) ||
+      !fs.existsSync(path.resolve("localhost-cert.pem"))
```
sophstad, line 61: [nit] Can we pull these file names out into variables?

packages/vite-utils/README.md
```diff
@@ -0,0 +1,32 @@
+# vite-utils
```
sophstad, line 1: Whoa this was a big lift, sorry! Works great though and I think it's reasonable to have a package like this. Do you anticipate the TS compiling as being easy to maintain? Introducing a big support burden here just to avoid duplicate code would be a notable tradeoff.

apps/spruce/README.md
```diff
@@ -23,6 +23,7 @@ taking the following steps.
 
 1. Generate a self signed certificate and add it to your keychain. This will
    allow you to run Spruce against a remote Evergreen server that uses HTTPS.
+   Run the below command in the root of the Parsley repository.
```
sophstad, line 26:
```suggestion
   Run the below command in the root of the Spruce repository.
```

packages/lib/src/context/AuthProvider/index.tsx
```diff
@@ -0,0 +1,216 @@
...
+ */
+const AuthProvider: React.FC<{
+  children: React.ReactNode;
+  /** `evergreenAppURL` is the URL of the backend service we will authenticate against */
+  evergreenAppURL: string;
+  /** `remoteAuthURL` is the URL that we should redirect the user to if they are not currently authenticated when `shouldUseLocalAuth` is false */
+  remoteAuthURL: string;
+  /** `localAuthRoute` is the route that we should redirect the user to if they are not currently authenticated when `shouldUseLocalAuth` is true */
+  localAuthRoute: string;
+  /** `shouldUseLocalAuth` determines which authentication strategy we should use. */
```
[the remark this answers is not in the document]
sophstad: Is there still a need to have these comments alongside the others?

packages/vite-utils/README.md
khelif96: I added the compile step to the [postinstall script](https://github.com/khelif96/ui/blob/ed3b0f15f97f2793c60d4b3391c9789b5849155d/package.json#L18). So it should happen automatically after installing dependencies. I don't anticipate us changing this directory much so I think the support burden would be any more than running `yarn install`.

packages/vite-utils/tsconfig.json
```diff
@@ -0,0 +1,28 @@
...
+    "compilerOptions": {
+        "target": "ES2019",
+        "module": "ESNext",
+        "lib": [
+            "ESNext",
+            "DOM"
+        ],
+        "declaration": true,
+        "declarationMap": true,
+        "sourceMap": true,
```
SupaJoon, line 11: Is the sourceMap actually utilized?

SupaJoon (approved):

sophstad (approved):

PR #643, opened 2025-02-20
DEVPROD-14939: Log e2e mongo output to file

SupaJoon:
DEVPROD-14939

### Description
These code changes remove mongod output from the e2e task log and include them in a separate file called mongod-logs


.evergreen/shared.yml
```diff
@@ -89,14 +89,17 @@ functions:
...
       params:
         background: true
         # Turn off silent to enable DB logging for debugging purposes
         silent: true
         working_dir: ui/evergreen
-        command: make start-mongod
+        shell: bash
+        script: | 
+          mkdir -p bin
+          make start-mongod > bin/mongod-logs.txt
```
sophstad, line 102: Did you try the [`--logpath`](https://www.mongodb.com/docs/manual/reference/program/mongod/#std-option-mongod.--logpath) option? The native way may be preferable (I think as this is we'd omit stderr from the file)

SupaJoon: I included stderr in the log file with 2>&1. This will stream stderr to stdout

sophstad (approved):

PR #652, opened 2025-02-25
DEVPROD-14677: Add menu button to clear all filters

minnakt:
DEVPROD-14677

### Description
Adds a menu button for clearing all filters. I included `date` and `revision` in the filters that get cleared, which I was a little unsure about.

Note that I changed `StatusFilter` and `RequesterFilter` from uncontrolled to controlled components. I realize we did it that way because it's glitchy, but if it's not controlled, the component can't capture events that happen outside of itself (e.g. updating the URL externally). The glitchiness is less apparent on production builds compared to development builds.

### Screenshots

https://github.com/user-attachments/assets/4436bcf6-f5aa-4a10-a7d9-c8233bfc681d

### Testing
* Added Cypress test


apps/spruce/src/pages/waterfall/WaterfallMenu/index.tsx
```diff
@@ -17,6 +18,7 @@ export const WaterfallMenu: React.FC<Props> = ({ projectIdentifier }) => {
   const dropdownItems = [
     <GitCommitSearch key="git-commit-search" setMenuOpen={setMenuOpen} />,
     <JumpToMostRecent key="jump-to-most-recent" setMenuOpen={setMenuOpen} />,
+    <ClearAllFilters key="clear-all-filters" setMenuOpen={setMenuOpen} />,
     <AddNotification
```
sophstad, line 22: Thoughts on adding icons to these? We could do like... `Code`, `ClockWithArrow`, `Eraser`, `Megaphone`?

sophstad (approved):

apps/spruce/src/pages/waterfall/WaterfallMenu/index.tsx
minnakt: how is this:

Screenshot 2025-02-26 at 12 37 12 PM

some more options:

Screenshot 2025-02-26 at 12 36 59 PM

sorry i would ask you on slack but it logged me out and now im locked out for the time being i guess https://slack-status.com/

sophstad: Oh good finds! `NoFilter` and `Bell` are great. I still slightly prefer `ClockWithArrow` over `Return` but your call.

Thanks 😻

PR #653, opened 2025-02-25
DEVPROD-11707: Fetch more versions on filter

sophstad:
DEVPROD-11707

### Description

When a user applies a filter that reduces the number of versions on the page, query the server for more. This only applies for requesters and build variants right now.

### Screenshots

image

### Testing

Add caching tests, update integration tests


apps/spruce/src/pages/waterfall/caching/index.ts
```diff
@@ -10,7 +10,7 @@ export const readVersions = ((existing, { args, readField }) => {
   const minOrder = args?.options?.minOrder ?? 0;
   const maxOrder = args?.options?.maxOrder ?? 0;
   const limit = args?.options?.limit ?? VERSION_LIMIT;
-  const { mostRecentVersionOrder = 0 } =
+  const { activeVersionIds = [], mostRecentVersionOrder = 0 } =
```
minnakt, line 13: Oof, something I noticed (which we don't have to fix for the beta test, since it sounds hard) is that I think the cache also needs to check the build variant match. Otherwise, it just returns 5 activated versions, but some of those could be collapsed due to the filter, so you have less than 5 matching versions on the page

apps/spruce/src/pages/waterfall/WaterfallGrid.tsx
```diff
@@ -188,6 +232,7 @@ export const WaterfallGrid: React.FC<WaterfallGridProps> = ({
               </InactiveVersion>
             );
           })}
+          {isPending && <FetchMoreLoader />}
```
khelif96, line 235: There might be some sort of race happening where the FetchMoreLoader shows up even if we have a full page of tasks loaded.
image

khelif96 (commented):
Sorry if this is already ticketed but should we add an empty state here?

apps/spruce/src/pages/waterfall/WaterfallGrid.tsx
```diff
@@ -141,13 +158,40 @@ export const WaterfallGrid: React.FC<WaterfallGridProps> = ({
...
+      };
+      startTransition(() => {
+        setServerFilters(filters);
+      });
+    } else if (serverQueryParamsEmpty && serverFilters !== resetFilterState) {
+      startTransition(() => {
+        setServerFilters(resetFilterState);
+      });
+    }
+  }, [allQueryParams]);
```
minnakt, line 193: nit. fix or ignore ESLint error

apps/spruce/src/pages/waterfall/WaterfallGrid.tsx
```diff
@@ -141,13 +158,40 @@ export const WaterfallGrid: React.FC<WaterfallGridProps> = ({
...
+  const [allQueryParams] = useQueryParams();
+
+  useEffect(() => {
+    const serverQueryParamsEmpty =
+      !Object.keys(allQueryParams).includes(
+        WaterfallFilterOptions.Requesters,
+      ) &&
+      !Object.keys(allQueryParams).includes(
+        WaterfallFilterOptions.BuildVariant,
+      );
```
minnakt, line 179: nit. this is just a matter of preference, but I think reversing the condition might be slightly more readable
```
 const hasServerParams =
      Object.keys(allQueryParams).includes(WaterfallFilterOptions.Requesters) ||
      Object.keys(allQueryParams).includes(WaterfallFilterOptions.BuildVariant);
```

apps/spruce/src/pages/waterfall/caching/index.ts
```diff
@@ -13,6 +13,7 @@ export const readVersions = ((existing, { args, readField }) => {
   const date = args?.options?.date ?? "";
   const revision = args?.options?.revision ?? "";
 
+  const activeVersionIds = [];
```
minnakt, line 16: nit. can we declare this closer to `allActiveVersions` 👀

apps/spruce/src/pages/waterfall/caching/index.ts
```diff
@@ -107,6 +110,7 @@ export const readVersions = ((existing, { args, readField }) => {
   return {
     flattenedVersions,
     pagination: {
+      activeVersionIds: activeVersionIds.sort(),
```
minnakt, line 113: is the sort just for guaranteeing consistency in testing? Could we add a comment if this is the case that the sort isn't inherently meaningful :0

apps/spruce/src/pages/waterfall/WaterfallGrid.tsx
```diff
@@ -104,18 +119,20 @@ export const WaterfallGrid: React.FC<WaterfallGridProps> = ({
           minOrder,
           revision,
           date: date ? fromZonedTime(date, timezone) : undefined,
+          ...serverFilters,
```
minnakt, line 122: idk why but when I visit mms with params `?buildVariants=ACS1&maxOrder=84102` the page just crashes for me 😭

I searched `"ACS1"` on mms then tried to go to the next page of results

sophstad: This should be addressed by https://github.com/evergreen-ci/evergreen/pull/8807 🤞

sophstad: Omg nvm DEVPROD-15622

sophstad: https://github.com/evergreen-ci/evergreen/pull/8808

apps/spruce/src/pages/waterfall/WaterfallGrid.tsx
```diff
@@ -136,18 +152,34 @@ export const WaterfallGrid: React.FC<WaterfallGridProps> = ({
...
+        requesters: allQueryParams.requesters as string[],
+      };
+      startTransition(() => {
+        setServerFilters(filters);
+      });
+    } else if (!hasServerParams && serverFilters !== resetFilterState) {
+      startTransition(() => {
+        setServerFilters(resetFilterState);
+      });
+    }
```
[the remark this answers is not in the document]
minnakt: or maybe `isPending` can be used to not update the waterfall content until it's ready?

PR #655, opened 2025-02-25
DEVPROD-14759: Disable waterfall datepicker before 1 year

SupaJoon:
DEVPROD-14759

### Description
The date picker will now prevent date selection before 1 year ago.

### DIFF
--- a/apps/spruce/src/pages/waterfall/DateFilter/index.tsx
+++ b/apps/spruce/src/pages/waterfall/DateFilter/index.tsx
@@ -41,7 +41,7 @@ export const DateFilter = () => {
       label="Go to Date"
       max={new Date()}
       // Testing environments should not have a minimum date restriction due to static test data.
-      {...(isProduction() && { min: subDays(subYears(new Date(), 1), 1) })}
+      min={isProduction() ? sub(today, { years: 1 }) : undefined}
       onDateChange={handleChange}
       popoverZIndex={zIndex.popover}
       value={date.length ? new Date(date) : undefined}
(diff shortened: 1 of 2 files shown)


apps/spruce/src/pages/waterfall/DateFilter/index.tsx
```diff
@@ -38,6 +39,7 @@ export const DateFilter = () => {
       data-cy="date-picker"
       label="Go to Date"
       max={new Date()}
+      min={subDays(subYears(new Date(), 1), 1)}
```
minnakt, line 42: (regarding test failures) maybe we should remove this limitation for local & dev environments? Even if we update the test data, if a year passes, we will run into the same failures

apps/spruce/src/pages/waterfall/DateFilter/index.tsx
```diff
@@ -38,6 +40,8 @@ export const DateFilter = () => {
       data-cy="date-picker"
       label="Go to Date"
       max={new Date()}
+      // Testing environments should not have a minimum date restriction due to static test data.
+      {...(isProduction() && { min: subDays(subYears(new Date(), 1), 1) })}
```
minnakt, line 44: nit. It might be a little bit more readable using the `sub` function. I think you can do something like this:
```suggestion
      min={isProduction() ? sub(today, { years: 1 }) : undefined}
```

where `today = new Date()` (could be declared as a constant at the bottom of the file if you want)

minnakt (approved):
LGTM!

PR #656, opened 2025-02-26
DEVPROD-12285: Allow users to modify volume size

minnakt:
DEVPROD-12285

### Description
Shows size as a field in the `EditVolumeModal`.

Volume size can only be increased, not decreased. This is simply a rule that comes from [AWS](https://docs.aws.amazon.com/cli/latest/reference/ec2/modify-volume.html#options)

### Screenshots

Screenshot 2025-02-25 at 8 09 23 PM

### Testing
* Add Cypress tests

### DIFF
--- a/apps/spruce/cypress/integration/spawn/volume.ts
+++ b/apps/spruce/cypress/integration/spawn/volume.ts
@@ -151,7 +151,7 @@ describe("Spawn volume page", () => {
       cy.dataCy("update-volume-modal").should("be.visible");
     });
 
-    it("Volume name & expiration inputs should be populated with the volume display name & expiration on initial render", () => {
+    it("name, size, expiration inputs should be populated on initial render", () => {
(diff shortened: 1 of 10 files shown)


SupaJoon (changes requested):
Saving the volume doesn't persist and I'm unsure if this is a limitation of the testing environment. 

https://github.com/user-attachments/assets/2acb88cf-3fa6-49ae-b36f-c61ef5ce0d32

minnakt:
> Saving the volume doesn't persist and I'm unsure if this is a limitation of the testing environment.
> 
>  bug.mov

since the changes are deployed on production, you can test against production and see that the volume size changes

SupaJoon (approved):

minnakt:
I also just learned you can only edit volume size once every 6 hours; I'll add a disclaimer about that in the modal

PR #657, opened 2025-02-26
DEVPROD-14877: Use `status` if version was created before `displayStatusCache` was added

minnakt:
DEVPROD-14877

### Description

Prioritizes `status` over `displayStatusCache` if the version was created before the `displayStatusCache` changes were finalized. It can be removed after a year (task TTL).

I based the timestamp off of evergreen-ci/evergreen/pull/8644, because it's the last change, but let me know if that's wrong 😅

### Testing
- Tasks mentioned in the ticket now correctly show as succeeded

### DIFF
--- a/apps/spruce/src/pages/waterfall/constants.ts
+++ b/apps/spruce/src/pages/waterfall/constants.ts
@@ -64,5 +64,6 @@ export const tupleSelectOptions = [
 
 /**
  * Timestamp of the last deploy that made changes to `displayStatusCache`, in UTC.
+ * TODO: Remove in DEVPROD-15269.
  */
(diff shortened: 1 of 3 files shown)


apps/spruce/src/pages/waterfall/constants.ts
```diff
@@ -61,3 +61,8 @@ export const tupleSelectOptions = [
     validator: () => true,
   },
 ];
+
+/**
+ * Timestamp of the last deploy that made changes to `displayStatusCache`, in UTC.
+ */
```
sophstad, line 67: Could you make a ticket to remove this code 1 year from this date?

sophstad (changes requested):

sophstad (approved):

PR #658, opened 2025-02-27
DEVPROD-14702: Preserve Task Status filters when switching waterfall projects

SupaJoon:
DEVPROD-14702

### Description
These code changes apply the current waterfall task filters when selecting a new project

### DIFF
--- a/apps/spruce/src/constants/routes.ts
+++ b/apps/spruce/src/constants/routes.ts
@@ -1,5 +1,6 @@
 import { stringifyQuery } from "@evg-ui/lib/src/utils/query-string";
 import { getGithubCommitUrl } from "constants/externalResources";
+import { WaterfallFilterOptions } from "pages/waterfall/types";
 import { TestStatus, HistoryQueryParams } from "types/history";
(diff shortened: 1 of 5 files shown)


apps/spruce/src/constants/routes.ts
```diff
@@ -287,8 +288,16 @@ export const getDistroSettingsRoute = (
 export const getCommitsRoute = (projectIdentifier: string = "") =>
   `${paths.commits}/${encodeURIComponent(projectIdentifier)}`;
 
-export const getWaterfallRoute = (projectIdentifier: string = "") =>
-  `${paths.project}/${encodeURIComponent(projectIdentifier)}${paths.waterfall}`;
+export const getWaterfallRoute = (
+  projectIdentifier?: string,
```
sophstad, line 292: Is there any reason not to require `projectIdentifier`?

SupaJoon: It avoids having to cast projectIdentifier/blocking the render when it comes from useParams or Apollo. Also `getWaterfallRoute` originally allowed an optional value so I preserved that functionality.

PR #660, opened 2025-02-27
DEVPROD-14935: Show metadata title while loading

sophstad:
DEVPROD-14935

### Description

Convert `MetadataTitle` into a `title` prop for the `MetadataCard` component. This allows the card to still show the title while displaying the loading state, and is also a better pattern in general imo.

Would also like to combine `MetadataLabel` and `MetadataItem` in a similar way, but that feels like a different ticket.


khelif96 (approved):
Nice

PR #664, opened 2025-02-27
DEVPROD-15037: Poll version status on Version Page

SupaJoon:
DEVPROD-15037

### Description
These code changes utilize `usePolling` for the Version status field which will reload data when the tab goes from inactive to active.

### DIFF
--- a/apps/spruce/src/pages/Version.tsx
+++ b/apps/spruce/src/pages/Version.tsx
@@ -13,14 +13,9 @@ import {
 } from "components/styles";
 import { Requester } from "constants/requesters";
 import { slugs } from "constants/routes";
-import {
-  VersionQuery,
-  VersionQueryVariables,
-  VersionStatusQuery,
-  VersionStatusQueryVariables,
-} from "gql/generated/types";
-import { VERSION, VERSION_STATUS } from "gql/queries";
-import { usePolling, useSpruceConfig } from "hooks";
+import { VersionQuery, VersionQueryVariables } from "gql/generated/types";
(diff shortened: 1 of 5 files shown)


apps/spruce/src/pages/Version.tsx
```diff
@@ -32,11 +37,10 @@ export const VersionPage: React.FC = () => {
...
   // This query is used to fetch the version data.
-  const {
-    data: versionData,
-    error: versionError,
-    loading: versionLoading,
-  } = useQuery<VersionQuery, VersionQueryVariables>(VERSION, {
+  const { data: versionData, error: versionError } = useQuery<
+    VersionQuery,
+    VersionQueryVariables
+  >(VERSION, {
```
khelif96, line 43: Why don't we just poll this query instead of creating a separate query?

khelif96 (changes requested):

apps/spruce/src/pages/Version.tsx
SupaJoon: I was thinking of doing that but remembered that some users were sensitive to amount of network traffic utilized and I thought it would be a good idea to keep it lower if possible. Here is a ticket that came from user feedback years ago: https://jira.mongodb.org/browse/EVG-12420

khelif96: I don't think this makes a difference you are just adding a second query that will be polling.

This will now run 2 queries for every page load and then poll 1 of them vs Just polling the original query.

khelif96 (approved):

PR #668, opened 2025-02-27
DEVPROD-1031: Update schedule/unschedule button state on Task page

SupaJoon:
DEVPROD-1031

### Description
The schedule and unschedule button disabled states on the Task page were not updating after clicking because the mutations did not fetch canSchedule and canUnschedule. These code changes add those fields so they can get integrated into the gql cache and update the button states.


khelif96 (approved):
Don't forget to add a title before merging

PR #672, opened 2025-02-28
DEVPROD-15343: Replace deprecated `ConfirmationModal` props

minnakt:
DEVPROD-15343

### Description
`buttonText`, `submitDisabled`, `onCancel`, and `onConfirm` are deprecated props. Use updated props to facilitate smoother upgrade when these props are officially removed.

### DIFF
--- a/apps/spruce/src/pages/version/NameChangeModal/index.tsx
+++ b/apps/spruce/src/pages/version/NameChangeModal/index.tsx
@@ -41,7 +41,7 @@ export const NameChangeModal: React.FC<NameChangeModalProps> = ({
     refetchQueries: ["Version"],
   });
 
-  const { newPatchName } = formState;
+  const { newPatchName = "" } = formState;
 
   return (
     <>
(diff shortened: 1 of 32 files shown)


apps/spruce/src/pages/version/NameChangeModal/index.tsx
```diff
@@ -53,18 +53,21 @@ export const NameChangeModal: React.FC<NameChangeModalProps> = ({
...
+        cancelButtonProps={{
+          onClick: () => setIsOpen(false),
+        }}
+        confirmButtonProps={{
+          children: "Confirm",
+          disabled:
+            newPatchName === originalPatchName || hasFormError || loading,
+          onClick: () => {
+            updateDescription({
+              // @ts-expect-error: FIXME. This comment was added by an automated script.
```
khelif96, line 65: I know this was here before but would it be a big lift to fix this expect-error since we are already updating this part of the codebase?

khelif96 (approved):

PR #684, opened 2025-03-04
DEVPROD-15369: Don't cut off waterfall build variant pin

SupaJoon:
DEVPROD-15369

Before:
Screenshot 2025-03-04 at 10 55 22 AM
After:
Screenshot 2025-03-04 at 10 55 07 AM


apps/spruce/src/pages/waterfall/WaterfallGrid.tsx
```diff
@@ -222,8 +222,8 @@ const StickyHeader = styled(Row)<{ atTop: boolean }>`
   z-index: 1;
 
   background: white;
-  margin: 0 -${size.m};
-  padding: ${size.xs} ${size.m};
+  margin: ${size.xs} -${size.m};
+  padding: 0px ${size.m};
```
minnakt, line 226: any way to keep some vertical padding? 👀 (when you scroll down on a large project, the header becomes a little squished now)

would this work? (maybe adjust some values as you need)
```suggestion
  margin: ${size.xxs} -${size.m};
  padding: ${size.xxs} ${size.m};
```

minnakt (approved):

PR #685, opened 2025-03-04
DEVPROD-8665: Move `Icon` and `IconWithTooltip` components to lib

minnakt:
DEVPROD-8665

### Description
Sorry... just lots of moving files and updating imports 😔

I also noticed global styles weren't working in the Lib storybook, so I just copy and pasted the existing static assets. We have multiple copies of static assets and should consider fixing it in [DEVPROD-15453](https://jira.mongodb.org/browse/DEVPROD-15453).


packages/lib/.storybook/main.ts
```diff
@@ -11,14 +11,7 @@ export default {
...
 
     return mergeConfig(config, {
-      plugins: [
-        tsconfigPaths(),
-        {
-          react: {
-            jsxImportSource: "@emotion/react",
-          },
-        },
-      ],
```
khelif96, line 21: Is this no longer necessary?

khelif96 (approved):
Just one question?

SupaJoon (approved):

packages/lib/.storybook/main.ts
minnakt: it's not necessary since we don't use the css prop in any stories in lib

PR #704, opened 2025-03-17
DEVPROD-15801 Add title field to Task history cell statuses

khelif96:
DEVPROD-15801

### Description
Add title to Task History cells so a title shows up on hover. 
Removes duplicate taskStatusToCopy usage

### Screenshots
image


khelif96:
#optibot review

SupaJoon (approved):
I agree with opti-bot that the casting is no good. Ideally, task status has a type other than string and originates from GQL but it may be more plumbing than what it's worth.

PR #706, opened 2025-03-18
DEVPROD-13884: Upgrade Node.js

sophstad:
DEVPROD-13884

### Description

I tried setting a project-level parameter for this variable, but that made it impossible to unset the node_version for our generator variant, where it's helpful to do so to have yarn install fail and therefore complete the task quickly.

So, upgrade to Node 22 in all places where Node 20 was specified.


minnakt (approved):
ok yay!!

PR #707, opened 2025-03-19
DEVPROD-8666 Move error reporting into lib

khelif96:
DEVPROD-8666

### Description
Move the Error Reporting configuration into the lib directory so we can reuse the code


minnakt (commented):
Could we just double check with a staging deploy for each app that the errors will still be properly reported to Sentry?

khelif96:
> Could we just double check with a staging deploy for each app that the errors will still be properly reported to Sentry?

Confirmed https://mongodb-org.sentry.io/issues/?project=4505211042856960&project=4505902479441920&query=is:unresolved%20wrekt&statsPeriod=24h

minnakt (approved):
sgtm

PR #709, opened 2025-03-20
DEVPROD-12427: Add test to verify that filters with commas are parsed correctly

minnakt:
DEVPROD-12427

### Description
This ticket was filed as a bug ticket, but it appears to be functioning correctly.

However, we don't have any tests to verify that it won't break in the future. I have added a test to ensure that this functionality will not break.

### Testing
* [Patch with module](https://spruce.mongodb.com/version/67db8e3101437b0007ccb9cd/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

### Evergreen PR
* evergreen-ci/evergreen/pull/8852

### DIFF
--- a/apps/parsley/cypress/integration/project_filters.ts
+++ b/apps/parsley/cypress/integration/project_filters.ts
@@ -31,6 +31,21 @@ describe("project filters", () => {
     cy.get("[data-cy^='skipped-lines-row-']").should("exist");
   });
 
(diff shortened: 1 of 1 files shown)


SupaJoon (approved):

minnakt:
> here

looks like the test file already has a test with commas, yay

PR #710, opened 2025-03-21
DEVPROD-16014 Fix scripts relying on vite config.

khelif96:
DEVPROD-16014

### Description
After merging #642, UI deploys began to fail. Upon investigation, I discovered that the deploy process itself wasn’t broken — [the issue stemmed from the scripts being run during the deploy.](https://parsley.mongodb.com/evergreen/evergreen_ui_spruce_deploy_spruce_prod__spruce_v5.0.1_25_03_21_15_40_28/0/task?bookmarks=0%2C230&selectedLineRange=L136&shareLine=136)
These [scripts](https://github.com/evergreen-ci/ui/blob/9f8705ff08a993962b905fc9fc0fabe0153bd902/packages/deploy-utils/src/write-previous-deploy/script.ts#L1) are written in TypeScript and executed using vite-node as standalone shell scripts. They rely on vite.config.ts from each project to resolve dependencies and settings.
#642 introduced changes that depend on certain environment configurations at runtime. Since the deploy scripts are meant to be self-contained and don’t require the full Vite project config, they began to break due to the now-required env setup in `vite.config.ts`.
This PR introduces a dedicated environment variable that is used to conditionally bypass Vite config logic when running scripts. This allows the scripts to run in isolation without depending on the broader project environment or config, preventing failures during deployment.

### Testing 
Any scripts in deploy-utils are now runnable manually and would fail if ran without these changes

### DIFF
--- a/.evergreen/deploy.yml
+++ b/.evergreen/deploy.yml
@@ -26,7 +26,7 @@ functions:
           - REACT_APP_PARSLEY_SENTRY_DSN
           - REACT_APP_SPRUCE_SENTRY_DSN
           - REACT_APP_SPRUCE_URL
-          - REACT_APP_UI_URL
+          - REACT_APP_EVERGREEN_URL
           - REACT_APP_USER_KEY
           - SPRUCE_SENTRY_AUTH_TOKEN
         env:
--- a/apps/parsley/vite.config.ts
+++ b/apps/parsley/vite.config.ts
@@ -12,114 +12,119 @@ import path from "path";
 import { generateBaseHTTPSViteServerConfig } from "@evg-ui/vite-utils";
 import injectVariablesInHTML from "./config/injectVariablesInHTML";
 
-// Remove when https://github.com/cypress-io/cypress/issues/25397 is resolved.
-dns.setDefaultResultOrder("ipv4first");
+let finalConfig: Record<string, any> = {};
 
-const serverConfig = generateBaseHTTPSViteServerConfig({
-  port: 5173,
-  appURL: process.env.REACT_APP_PARSLEY_URL,
-  httpsPort: 8444,
-  useHTTPS: process.env.REACT_APP_RELEASE_STAGE !== "local",
-});
+// Only run this configuration if we are not in script mode
+if (process.env.VITE_SCRIPT_MODE !== "1") {
+  // Remove when https://github.com/cypress-io/cypress/issues/25397 is resolved.
+  dns.setDefaultResultOrder("ipv4first");
 
-// https://vitejs.dev/config/
-const viteConfig = defineConfig({
-  server: serverConfig,
-  build: {
-    rollupOptions: {
-      plugins: [],
+  const serverConfig = generateBaseHTTPSViteServerConfig({
+    port: 5173,
+    appURL: process.env.REACT_APP_PARSLEY_URL,
+    httpsPort: 8444,
+    useHTTPS: process.env.REACT_APP_RELEASE_STAGE !== "local",
+  });
+
+  // https://vitejs.dev/config/
+  const viteConfig = defineConfig({
+    server: serverConfig,
+    build: {
+      rollupOptions: {
+        plugins: [],
+      },
+      sourcemap: true,
     },
-    sourcemap: true,
-  },
 
-  plugins: [
-    tsconfigPaths(),
-    react({
-      babel: {
-        // @emotion/babel-plugin injects styled component names (e.g. "StyledSelect") into HTML for dev
-        // environments only. It can be toggled for production environments by modifying the parameter
-        // autoLabel. (https://emotion.sh/docs/@emotion/babel-plugin)
-        plugins: ["@emotion/babel-plugin", "import-graphql"],
-      },
-      // Exclude storybook stories from fast refresh.
-      exclude: /\.stories\.tsx?$/,
-      // Only Typescript files should use fast refresh.
-      include: ["**/*.tsx", "**/*.ts"],
+    plugins: [
+      tsconfigPaths(),
+      react({
+        babel: {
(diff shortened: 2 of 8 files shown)


apps/spruce/vite.config.ts
```diff
@@ -16,152 +16,159 @@ import path from "path";
...
 
-// Remove when https://github.com/cypress-io/cypress/issues/25397 is resolved.
-dns.setDefaultResultOrder("ipv4first");
+// Only run this configuration if we are not in script mode
+if (process.env.VITE_SCRIPT_MODE !== "1") {
+  const require = createRequire(import.meta.url);
 
-// Do not apply Antd's global styles
-fs.writeFileSync(require.resolve("antd/es/style/core/global.less"), "");
-fs.writeFileSync(require.resolve("antd/lib/style/core/global.less"), "");
```
khelif96, line 26: Since this diff is way larger than it should be the only real change here is moving finalConfig into its own variable. and nesting the rest of the config under the conditional on L21

apps/parsley/vite.config.ts
```diff
@@ -12,114 +12,119 @@ import path from "path";
...
+let finalConfig: Record<string, any> = {};
 
-const serverConfig = generateBaseHTTPSViteServerConfig({
-  port: 5173,
-  appURL: process.env.REACT_APP_PARSLEY_URL,
-  httpsPort: 8444,
-  useHTTPS: process.env.REACT_APP_RELEASE_STAGE !== "local",
-});
+// Only run this configuration if we are not in script mode
+if (process.env.VITE_SCRIPT_MODE !== "1") {
```
khelif96, line 18: Since this diff is way larger than it should be the only real change here is moving finalConfig into its own variable. and nesting the rest of the config under the conditional on L18

apps/parsley/vite.config.ts
```diff
@@ -12,114 +12,119 @@ import path from "path";
 import { generateBaseHTTPSViteServerConfig } from "@evg-ui/vite-utils";
```
minnakt, line 12: I think this might break the staging / beta deploys

I just tried in a [patch](https://spruce.mongodb.com/task/evergreen_ui_deploy_deploy_spruce_staging_patch_97c982fb52d4f4a2e1b86b1332f2bc4b9019ef76_67ddd40a24d8d700074f34b0_25_03_21_21_03_18/logs?execution=0) on this PR (no changes, clean) and it failed for me

khelif96: Great catch! Updated the script to disable script mode when running yarn build
https://spruce.mongodb.com/version/67ddd6bc8b22cb0007f95dfe/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

.evergreen/deploy.yml
```diff
@@ -26,7 +26,7 @@ functions:
           - REACT_APP_PARSLEY_SENTRY_DSN
           - REACT_APP_SPRUCE_SENTRY_DSN
           - REACT_APP_SPRUCE_URL
-          - REACT_APP_UI_URL
+          - REACT_APP_EVERGREEN_URL
```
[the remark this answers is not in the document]
khelif96: Moved!

packages/deploy-utils/src/build-and-push/index.ts
```diff
@@ -8,7 +8,8 @@ import { pushToS3 } from "../utils/s3";
  * @param bucket - bucket to push to
  */
 export const buildAndPush = (bucket: string) => {
-  execSync("yarn build", { stdio: "inherit" });
+  // Disable script mode to let vite build the project
+  execSync("VITE_SCRIPT_MODE=0 yarn build", { stdio: "inherit" });
```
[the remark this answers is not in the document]
khelif96: @agent-optibot this is not what the intent is. We want to overwrite VITE_SCRIPT_MODE when we are running the deploy because the env variable will be passed to the `yarn build` command

packages/deploy-utils/src/write-previous-deploy/script.ts
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env -S vite-node --script
+#!/usr/bin/env -S VITE_SCRIPT_MODE=1 vite-node --script
```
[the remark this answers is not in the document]
khelif96: Yes it does!

apps/parsley/vite.config.ts
```diff
@@ -12,114 +12,119 @@ import path from "path";
...
 import injectVariablesInHTML from "./config/injectVariablesInHTML";
 
-// Remove when https://github.com/cypress-io/cypress/issues/25397 is resolved.
-dns.setDefaultResultOrder("ipv4first");
+let finalConfig: Record<string, any> = {};
```
minnakt, line 15: I didn't actually test to see if this works, but it could help with readability to write it as
```suggestion
const useProjectConfig = process.env.VITE_SCRIPT_MODE !== "1";

const finalConfig = useProjectConfig ? getProjectConfig() : {};

export default finalConfig;
```

where `getProjectConfig()` contains the project config logic.

packages/deploy-utils/src/build-and-push/index.ts
minnakt, line 12: Alternatively, you could add the blocking code specifically on `serverConfig` i.e.

```
const serverConfig = useProjectConfig ? { // real config // } : {}
```

and you wouldn't need the environment variable

```suggestion
  execSync("yarn build", { stdio: "inherit" });
```

.evergreen/deploy.yml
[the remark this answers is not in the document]
minnakt: does this thing need to be removed if we deleted the var, or does it exist elsewhere?

minnakt (approved):
lgtm with suggestions

PR #717, opened 2025-03-25
DEVPROD-15849: Render task timeline

minnakt:
DEVPROD-15849

### Description
Create a basic outline for the task timeline. Moved the task box logic into its own component so it can be shared between the Waterfall and Task History.

We will build on this with pagination, date labels, etc.

### Screenshots

https://github.com/user-attachments/assets/38f05720-95f7-40f6-8991-ecd0a865c11c

### Testing
* e2e tests will be added when I add more data in DEVPROD-15980
* Coincidentally fixed some snapshot tests that were broken

### Evergreen PR
* evergreen-ci/evergreen/pull/8865


khelif96 (changes requested):

apps/spruce/src/pages/task/taskTabs/TaskHistory/index.tsx
```diff
@@ -1,12 +1,111 @@
...
 
-const TaskHistory: React.FC<TaskHistoryProps> = ({ execution, taskId }) => (
-  <div data-cy="task-history">
-    This is the history for task {taskId} and execution {execution}
-  </div>
-);
+const TaskHistory: React.FC<TaskHistoryProps> = ({ task }) => {
+  const dispatchToast = useToastContext();
+
+  const [viewOption, setViewOption] = useState(ViewOptions.Collapsed);
```
SupaJoon, line 30: This may be useful as a cookie value

SupaJoon (changes requested):

apps/spruce/src/pages/task/taskTabs/TaskHistory/TaskTimeline/TaskTimeline.stories.tsx
```diff
@@ -0,0 +1,104 @@
...
+    shouldCollapse: true,
+  },
+  argTypes: {
+    shouldCollapse: {
+      control: { type: "boolean" },
+    },
+  },
+};
+
+export const Default: StoryObj<TimelineProps & { shouldCollapse: boolean }> = {
```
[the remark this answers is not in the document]
minnakt: hmm I can't get this to work without type errors. I want `shouldCollapse` as an argument so that you can toggle to collapse or uncollapse the tasks.

apps/spruce/src/pages/task/taskTabs/TaskHistory/index.tsx
minnakt: let's leave it without the cookie for now

apps/spruce/src/pages/task/taskTabs/TaskHistory/TaskTimeline/TaskTimeline.stories.tsx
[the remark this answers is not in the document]
khelif96: Ahh sorry I should have tested it out. You can use a Template component that wraps the component and provides any additional props/functionality
```ts
const Template = (
  args: React.ComponentProps<typeof TaskTimeline> & {
    shouldCollapse: boolean;
  },
) => {
  const groupedTasks = groupTasks(tasks, args.shouldCollapse);
  return <TaskTimeline groupedTasks={groupedTasks} loading={false} />;
};
```

```suggestion
export const Default: StoryObj<typeof Template> = {
```

apps/spruce/src/components/TaskBox/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+export const getTaskStatusStyle = (status: TaskStatus) => {
+  const icon = statusIconMap?.[status];
+  const iconStyle = icon ? `background-image: ${icon};` : "";
+  return css`
+    ${iconStyle}
+    background-color: ${statusColorMap[status]};
+  `;
+};
+
+export const TaskBox = styled(Link)<{ status: TaskStatus }>`
```
khelif96, line 22: So I noticed we have essentially 3 versions of this type of component.
This one.
[TaskStatsTooltip](https://github.com/evergreen-ci/ui/blob/6432255d25a5c676e491edf232335ba1c26711d7/apps/spruce/src/pages/waterfall/TaskStatsTooltip.tsx#L123)
[Waterfall Square](https://github.com/evergreen-ci/ui/blob/d25c9f2a4643e7edee2382f441c59f50dd37b376/apps/spruce/src/pages/waterfall/BuildRow.tsx#L251) Which seems to just be this one with some added tooltip logic. 

I think it would make sense to consolidate the 3 implementations so that we only have one instead of introducing an additional one. I also think the tooltip functionality for the waterfall would work really well here. 

https://github.com/user-attachments/assets/bbd5b2ee-6e45-47f7-8ab7-d441a9355069

apps/spruce/src/pages/task/TaskTabs.tsx
```diff
@@ -171,7 +171,7 @@ export const TaskTabs: React.FC<TaskTabProps> = ({ isDisplayTask, task }) => {
         }
       >
         {/* @ts-expect-error: FIXME. This comment was added by an automated script. */}
```
khelif96, line 173: Unrelated but I think the `ts-expect-error` is related to the `TaskTabs` component being typed wrong. 
It thinks `Task` can be undefined meanwhile we guard against it  being undefined [here](https://github.com/evergreen-ci/ui/blob/5593651aecf0b3105f08c6042450cc96ba361ea8/apps/spruce/src/pages/Task.tsx#L155)

apps/spruce/src/pages/task/taskTabs/TaskHistory/TaskTimeline/index.tsx
```diff
@@ -0,0 +1,107 @@
...
+  flex-direction: row;
+  align-items: center;
+  gap: ${size.xxs};
+`;
+
+const Timeline = styled.div`
+  width: 100%;
+`;
+
+const CollapsedBox = styled.div`
```
khelif96, line 96: Same comment as above. We should consolidate these box-style components so we can reuse the same styling.

apps/spruce/src/components/TaskBox/index.tsx
minnakt: I'm not convinced of the usefulness of the tooltip since the same information will be present directly below; I think we can revisit this idea at another time. I added the hover mechanism that Will specified in his designs in place of this

apps/spruce/src/components/TaskBox/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+    color: ${white};
+    text-align: center;
+    display: none;
+  }
+  :hover:before {
+    display: block;
+  }
+
+  /* Tooltip caret */
+  :hover:after {
```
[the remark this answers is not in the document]
minnakt: thanks, updated with this, except I kept `CollapsedBox` as a separate item since I think it's better for code readability

khelif96 (approved):
Great work! 🔥

SupaJoon (approved):

PR #718, opened 2025-03-25
DEVPROD-15346 Fix broken project events and event log refactor

khelif96:
DEVPROD-15346

### Description

This PR might seem large at first glance, but it primarily includes a refactor of the `EventLog` page along with some necessary fixes and improvements.

### EventLog Page Refactor

- While investigating why some events weren't rendering, I found the current structure hard to follow due to how files and methods were organized.
- I reorganized several files and functions to improve the overall readability and hierarchy of the EventLog component.
- During the refactor, I came across multiple `@ts-expect-error` comments. These have now been resolved, improving type safety across the component.
- Since Aldo is working on rebuilding the admin page (which will reuse the EventLog), I added:
  - Unit tests
  - A Storybook story  

#### Missing Project Event Diffs

- This also fixes an issue where some project events appeared empty in the log. The `before` and `after` fields in the event log return full `Project` objects. Since the project is already fetched earlier when the page loads, Apollo normalizes and merges the cache.
- This results in `before === after`, making the diff viewer think there are no changes.
- The fix was telling Apollo not to cache `Project`. I was unable to tell apollo not to cache the Project field if it orignates within a project event because of the way the cache works so this seemed like a reasonable alternative. 

### Screenshots
Opted not to include screenshots since its an event log for  a production project

### Testing
visit this page running `yarn prod`
https://spruce-local.corp.mongodb.com:8443/project/mongot-master/settings/event-log


apps/spruce/src/gql/client/cache.ts
```diff
@@ -81,6 +81,9 @@ export const cache = new InMemoryCache({
         },
       },
     },
+    Project: {
+      keyFields: false,
+    },
```
khelif96, line 86: This is the actual fix for the cache stuff

apps/spruce/src/components/Settings/EventLog/EventDiffTable/KeyRendererUtils/index.ts
minnakt, line 1: I think it might make sense to instead create a `utils` folder, put the other utils files in there, and put this in the folder as `utils/keyRenderer/index.ts`

apps/spruce/src/components/Settings/EventLog/EventDiffTable/utils.test.ts
```diff
@@ -0,0 +1,238 @@
...
+      patchTriggerAliases: [],
+    },
+    vars: {
+      __typename: "ProjectVars",
+      vars: {},
+    },
+  },
+};
+
+describe("should transform event diffs to key, before and after", () => {
```
minnakt, line 193: nit. can we name the test after the function it is testing
```suggestion
describe("getEventDiffLines", () => {
```

apps/spruce/src/components/Settings/EventLog/EventDiffTable/utils.ts
```diff
@@ -0,0 +1,80 @@
...
+  const eventDiff = diff(beforeNoTypename || {}, afterNoTypename || {});
+  const pathKeys: string[] = getChangedPaths(eventDiff);
+  const eventDiffLines = pathKeys.map((key) => {
+    let previousValue = beforeNoTypename;
+    if (beforeNoTypename !== null && beforeNoTypename !== undefined) {
+      previousValue = getObjectValueByPath(beforeNoTypename, key) as JSONObject;
+    }
+    const changedValue = getObjectValueByPath(eventDiff as JSONValue, key);
+
+    const formattedKey = formatArrayElements(key);
```
minnakt, line 29: I kinda didn't test this at all besides running the corresponding test file; if it doesn't work, we should modify the test file to include a case where this will fail, because I think someone would reasonably think it could be rewritten like this
```suggestion
  const beforeNoTypename = omitTypename(before) || {};
  const afterNoTypename = omitTypename(after) || {};

  const eventDiff = diff(beforeNoTypename, afterNoTypename) as JSONValue;
  const pathKeys: string[] = getChangedPaths(eventDiff);
  const eventDiffLines = pathKeys.map((key) => {
    const previousValue = getObjectValueByPath(beforeNoTypename, key);
    const changedValue = getObjectValueByPath(eventDiff, key);
    const formattedKey = formatArrayElements(key);

```

apps/spruce/src/components/Settings/EventLog/EventDiffTable/utils.ts
```diff
@@ -0,0 +1,80 @@
...
+  return eventDiffLines.filter((el) => el !== null);
+};
+
+/**
+ * `addDelimiter` is a helper function that adds a delimiter between two strings. If the first string is empty, it returns the second string.
+ * @param a - string
+ * @param b - string
+ * @returns - The concatenated string.
+ */
+const addDelimiter = (a: string, b: string): string => (a ? `${a}.${b}` : b);
```
minnakt, line 49: nit. suggesting for clarity
```suggestion
const addDelimiter = (a: string, b: string): string => (a.length > 0 ? `${a}.${b}` : b);
```

apps/spruce/src/utils/object/index.ts
```diff
@@ -1,3 +1,99 @@
...
+
+type WithoutTypename<T> =
+  T extends Record<string, unknown>
+    ? {
+        [K in keyof T as K extends "__typename" ? never : K]: WithoutTypename<
+          T[K]
+        >;
+      }
+    : T;
+/**
```
minnakt, line 78: nit. add extra line to separate from previous function

minnakt (approved):

apps/spruce/src/components/Settings/EventLog/EventDiffTable/utils.ts
khelif96: SGTM it worked!

PR #719, opened 2025-03-26
Disable auto reviews from optibot

khelif96:
Prevent optibot from auto commenting on prs without first being asked.


minnakt (approved):
tyyy

PR #722, opened 2025-03-28
DEVPROD-15374 Add logic to refresh on old bundle errors

khelif96:
DEVPROD-15374 

### Description
Error boundary should now automatically refresh the page when a user encounters a bundle error.

### DIFF
--- a/packages/lib/src/components/ErrorBoundary/utils.ts
+++ b/packages/lib/src/components/ErrorBoundary/utils.ts
@@ -1,4 +1,7 @@
-import { leaveBreadcrumb, SentryBreadcrumbTypes } from "utils/errorReporting";
+import {
+  leaveBreadcrumb,
+  SentryBreadcrumbTypes,
+} from "../../utils/errorReporting";
 
 /**
  * `refreshOnOldBundleError` is a utility function that refreshes the page if the error is due to an old bundle.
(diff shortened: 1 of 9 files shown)


packages/lib/src/components/ErrorBoundary/utils.ts
```diff
@@ -0,0 +1,22 @@
...
+  SentryBreadcrumbTypes,
+} from "../../utils/errorReporting";
+
+/**
+ * `refreshOnOldBundleError` is a utility function that refreshes the page if the error is due to an old bundle.
+ * This is useful when the user has an old version of the application and the bundle has changed.
+ * @param error - The error object
+ */
+const refreshOnOldBundleError = (error: Error) => {
+  if (error.message.includes("error loading dynamically imported module")) {
```
minnakt, line 12: Could we turn this string into a constant (and use the constant in the corresponding test too)?

minnakt (commented):
im not sure if im just testing this wrong, but i was trying to force the error on staging, but I could never reach the Sentry part 🤔 

i deployed multiple times with different bundles (since the deployed version is always 1 version behind, I think you have to deploy once to get the update in, and then deploy once again to test against that update)

minnakt (approved):
lgtm

i think we would expect this to work 2 deploys from now, so if it doesn't work on the first day, we don't need to be alarmed

PR #723, opened 2025-03-28
DEVPROD-1053 Replace antd skeletons with leafygreen 

khelif96:
DEVPROD-1053

### Description
Replaced every antd Skeleton implementation on Spruce with a variant form Leafygreen. I did my best to match or improve the styles. Prioritizing matching the look of the page that is rendered after the Skeleton. 

### Screenshots
Spawn Pages
image

Version and Task Pages
image

Restart modals
image


apps/spruce/src/pages/preferences/preferencesTabs/NotificationsTab.tsx
```diff
@@ -49,7 +48,7 @@ export const NotificationsTab: React.FC = () => {
   });
 
   if (loading) {
-    return <Skeleton active />;
+    return <FormSkeleton />;
```
minnakt, line 51: it looks like we're using `FormSkeleton` a lot for modals but not for pages. Should this have a different skeleton?

apps/spruce/src/pages/task/taskTabs/buildBaronAndAnnotations/BuildBaron.tsx
```diff
@@ -29,7 +29,7 @@ const BuildBaron: React.FC<Props> = ({
   );
   return (
     <>
-      {loading && <Skeleton active paragraph={{ rows: 4 }} title={false} />}
+      {loading && <ParagraphSkeleton />}
       {(data || annotation) && (
         <BuildBaronContent
```
minnakt, line 34: could we block any rendering if `loading` is true by bringing this up? + see next comment

```
if (loading) {
   return (<ParagraphSkeleton />)
}
```

apps/spruce/src/pages/task/taskTabs/buildBaronAndAnnotations/BuildBaronContent.tsx
```diff
@@ -67,7 +67,7 @@ const BuildBaronContent: React.FC<BuildBaronCoreProps> = ({
 
   return (
     <Wrapper data-cy="bb-content">
-      {loading && <Skeleton active paragraph={{ rows: 4 }} title={false} />}
+      {loading && <ParagraphSkeleton />}
```
minnakt, line 70: I think doing the previous comment will eliminate the need for this, since we pass `loading` down to this component. We can delete the `loading` prop as well

```suggestion
```

apps/spruce/src/pages/preferences/preferencesTabs/NotificationsTab.tsx
khelif96: Swapped it with Card since thats what renders here

PR #728, opened 2025-03-31
DEVPROD-16242 The big dramatic evergreen redesign!

khelif96:
DEVPROD-16242

### Description
The culmination of years of design and engineering research has lead us to come up with the most optimal experience for our users

### Screenshots
image

### Testing
Test on staging


apps/spruce/src/components/AprilFoolsDarkMode/index.tsx
minnakt, line 1: Should we remove the Dark Mode files since we decided against it?

apps/spruce/src/constants/cookies.ts
```diff
@@ -16,3 +16,5 @@ export const SUBSCRIPTION_METHOD = "subscription-method";
 export const SEEN_WATERFALL_LAUNCH_MODAL = "seen-waterfall-launch-modal";
 export const SEEN_WATERFALL_ONBOARDING_TUTORIAL =
   "seen-waterfall-onboarding-tutorial";
+export const SEEN_DARK_MODE_PRANK = "seen-dark-mode-prank";
```
minnakt, line 19: same comment

apps/spruce/src/components/AprilFoolsEvergreenRedesign/index.tsx
```diff
@@ -0,0 +1,181 @@
...
+        setComicSansEnabled(true);
+        setIsFolding(false);
+
+        // Now start the 10s timer for the follow-up prank modal
+        setTimeout(() => {
+          setIsFollowUp(true);
+          setModalOpen(true);
+          Cookies.set(SEEN_COMIC_SANS_PRANK, "true", { expires: 2 });
+        }, 10000);
+      }, 2000); // Match animation duration
```
minnakt, line 43: would recommend shortening duration to 500 or under

apps/spruce/src/components/AprilFoolsEvergreenRedesign/index.tsx
```diff
@@ -0,0 +1,181 @@
...
+              width: 100vw;
+              height: 100vh;
+              background: linear-gradient(to bottom, #007d3c 0%, #00642f 100%);
+              animation: curtainReveal 2s ease-in-out forwards;
+              transform-origin: top;
+              z-index: 2147483647 !important;
+              transform: translateZ(0);
+              will-change: transform;
+              pointer-events: all;
+            `}
```
minnakt, line 126: with the gradient, it's clear there are z-index issues; it may be less obvious if we use an opacity-like transition

```
     0% {
                background-color: rgba(255, 255, 255, 0.7);
              }
              100% {
                background-color: none;
              }
```

apps/spruce/src/components/AprilFoolsEvergreenRedesign/index.tsx
```diff
@@ -0,0 +1,181 @@
...
+            css={css`
+              position: fixed;
+              top: 0;
+              left: 0;
+              width: 100vw;
+              height: 100vh;
+              background: linear-gradient(to bottom, #007d3c 0%, #00642f 100%);
+              animation: curtainReveal 2s ease-in-out forwards;
+              transform-origin: top;
+              z-index: 2147483647 !important;
```
minnakt, line 122: (re: previous comment) z-index appears to have no effect

apps/spruce/src/components/AprilFoolsEvergreenRedesign/index.tsx
```diff
@@ -0,0 +1,181 @@
...
+        const trail = document.createElement("span");
+        trail.textContent = "✨"; // or try 🎉, 😎, 🐸, 💚, 🧃
+        trail.style.position = "fixed";
+        trail.style.left = `${e.clientX}px`;
+        trail.style.top = `${e.clientY}px`;
+        trail.style.pointerEvents = "none";
+        trail.style.zIndex = "2147483647";
+        trail.style.transform = "translate(-50%, -50%)";
+        trail.style.fontSize = "20px";
+        trail.style.opacity = "1";
```
minnakt, line 61: If we want to go with a geocities theme, rather than the trailing path, I'd suggest turning the cursor into a specific image. like could the cursor just be a picture of a banana or something

apps/spruce/src/components/AprilFoolsEvergreenRedesign/index.tsx
khelif96: Removed it all together since it didn't have the desired effect

apps/spruce/package.json
```diff
@@ -96,7 +96,7 @@
     "@leafygreen-ui/segmented-control": "^8.2.13",
     "@leafygreen-ui/select": "^12.0.0",
     "@leafygreen-ui/side-nav": "^14.1.3",
-    "@leafygreen-ui/skeleton-loader": "^1.2.0",
+    "@leafygreen-ui/skeleton-loader": "^2.0.8",
```
minnakt, line 99: was this upgrade intentional?

apps/spruce/src/components/AprilFoolsEvergreenRedesign/index.tsx
```diff
@@ -0,0 +1,151 @@
...
+      setModalOpen(false); // Hide original modal
+
+      setComicSansEnabled(true);
+
+      // Now start the 10s timer for the follow-up prank modal
+      setTimeout(() => {
+        setIsFollowUp(true);
+        setModalOpen(true);
+        Cookies.set(SEEN_COMIC_SANS_PRANK, "true", { expires: 2 });
+      }, 8000);
```
minnakt, line 39: nit. comment seems outdated, it looks like it's an 8s timer

apps/spruce/src/components/AprilFoolsEvergreenRedesign/index.tsx
khelif96: Why not both? I went with a unicorn cursor followed by sparkles as the trail. Lmk what you think

apps/spruce/src/components/AprilFoolsEvergreenRedesign/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+      const trailElements: HTMLSpanElement[] = [];
+
+      const handleMouseMove = (e: MouseEvent) => {
+        const trail = document.createElement("span");
+        trail.textContent = "✨";
+        trail.style.position = "fixed";
+        trail.style.left = `${e.clientX}px`;
+        trail.style.top = `${e.clientY}px`;
+        trail.style.pointerEvents = "none";
+        trail.style.zIndex = "2147483647";
```
minnakt, line 54: nit. trail.style.zIndex = `${zIndex.max_do_not_use}`;

apps/spruce/src/analytics/navbar/useNavbarAnalytics.ts
```diff
@@ -17,7 +17,9 @@ type Action =
   | { name: "Clicked profile link" }
   | { name: "Clicked notifications link" }
   | { name: "Clicked UI settings link" }
-  | { name: "Clicked task queue link" };
+  | { name: "Clicked task queue link" }
+  | { name: "Clicked disable april fools link" }
+  | { name: "Clicked enable april fools link in navbar" };
```
minnakt, line 22: nit. capitalization
```suggestion
  | { name: "Clicked disable April Fools link" }
  | { name: "Clicked enable April Fools link in navbar" };
```

minnakt, line 22: I also think these events might belong in `useAprilFoolsAnalytics`

apps/spruce/src/components/AprilFoolsEvergreenRedesign/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+                  auto;
+              }
+            `}
+          />
+        )}
+
+        <MarketingModal
+          buttonText={
+            isFollowUp
+              ? "The Evergreen team rocks for this!"
```
minnakt, line 102: My only problem with this copy is that it's not 100% clear that clicking this button will cause the Comic Sans state to persist

minnakt (approved):
approving since it's the end of the day

apps/spruce/src/analytics/navbar/useNavbarAnalytics.ts
khelif96: Good catch!

apps/spruce/package.json
khelif96: def not

apps/spruce/src/components/AprilFoolsEvergreenRedesign/index.tsx
khelif96: I think the text above "Comic Sans is here to stay" indicates that happy to swap it if you have any suggestions

minnakt: something like `"Continue the experience"` could be appropriate

PR #754, opened 2025-04-15
DEVPROD-16050: Support pagination on Task Timeline

minnakt:
[DEVPROD-16050](https://jira.mongodb.org/browse/DEVPROD-16050)

### Description
Adds basic pagination logic to Task History. A key consideration in this implementation that is that no matter if you're paginating backwards or forwards, visible tasks should always expand/shrink from the **right side**. This is because as a user, it feels unnatural to suddenly see tasks appear/disappear on the left side.

Further pagination (and tangentially, performance) improvements will be investigated in [DEVPROD-16186](https://jira.mongodb.org/browse/DEVPROD-16186).

### Performance Analysis
The performance actually seems pretty OK. [DEVPROD-16595](https://jira.mongodb.org/browse/DEVPROD-16595) helped a lot to make the overall query faster. Attaching an [example](https://ui.honeycomb.io/mongodb-4b/environments/production/datasets/spruce/result/mg2xnbYaXrJ/trace/34pQ8Xckm1R?fields[]=s_name&fields[]=s_serviceName&span=85ad741113129852) from a Server task.

### Screenshots

https://github.com/user-attachments/assets/db924a61-8880-4fc8-8c1a-493cac02a641

### Testing
* Cypress tests
* Unit tests

### DIFF
--- a/apps/spruce/cypress/integration/task/task_history.ts
+++ b/apps/spruce/cypress/integration/task/task_history.ts
@@ -54,8 +54,9 @@ describe("task history", () => {
   describe("pagination", () => {
     describe("can paginate forwards and backwards", () => {
       beforeEach(() => {
-        // Change the viewport size so that tasks overflow to the next page.
-        cy.viewport(1400, 1080);
+        // Change the viewport size so that tasks overflow to the next page. CI environment has a large scrollbar due to
+        // distorts the viewport size, so adjustments are made here.
+        cy.viewport(Cypress.env("CI") ? 1410 : 1400, 1080);
       });
 
       it("collapsed view", () => {
--- a/apps/spruce/cypress/integration/task/task_history.ts
+++ b/apps/spruce/cypress/integration/task/task_history.ts
@@ -54,9 +54,7 @@ describe("task history", () => {
         "/task/evg_lint_generate_lint_c6672b24d14c6d8cd51ce2c4b2b88b424aaacd64_25_03_27_14_56_09/history?execution=0";
 
       beforeEach(() => {
-        // Change the viewport size so that tasks overflow to the next page. CI environment has a large scrollbar due to
-        // Linux, which distorts the viewport size, so adjustments are made here.
-        cy.viewport(Cypress.env("CI") ? 1410 : 1400, 1080);
+        cy.viewport(1400, 1080);
       });
 
       it("collapsed view", () => {
--- a/apps/spruce/cypress/integration/task/task_history.ts
+++ b/apps/spruce/cypress/integration/task/task_history.ts
@@ -54,7 +54,9 @@ describe("task history", () => {
         "/task/evg_lint_generate_lint_c6672b24d14c6d8cd51ce2c4b2b88b424aaacd64_25_03_27_14_56_09/history?execution=0";
 
       beforeEach(() => {
-        cy.viewport(1400, 1080);
+        // Change the viewport size so that tasks overflow to the next page. CI environment has a large scrollbar due to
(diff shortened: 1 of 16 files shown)


apps/spruce/src/pages/task/taskTabs/TaskHistory/index.tsx
```diff
@@ -78,17 +80,43 @@ const TaskHistory: React.FC<TaskHistoryProps> = ({ task }) => {
...
   const { taskHistory } = data ?? {};
-  const { tasks = [] } = taskHistory ?? {};
+  const { pagination, tasks = [] } = taskHistory ?? {};
+  const { mostRecentTaskOrder, oldestTaskOrder } = pagination ?? {};
 
   const groupedTasks = groupTasks(tasks, shouldCollapse);
   const numVisibleTasks = Math.floor(timelineWidth / SQUARE_WITH_BORDER);
-  const visibleTasks = groupedTasks.slice(0, numVisibleTasks);
+
+  const visibleTasks =
```
SupaJoon, line 89: The number of visible tasks should be the next limit for the history query.

SupaJoon (changes requested):

apps/spruce/cypress/integration/task/task_history.ts
```diff
@@ -47,4 +50,110 @@ describe("task history", () => {
       });
     });
   });
+
+  describe("pagination", () => {
+    describe("can paginate forwards and backwards", () => {
+      beforeEach(() => {
+        // Change the viewport size so that tasks overflow to the next page. CI environment has a large scrollbar due to
+        // distorts the viewport size, so adjustments are made here.
+        cy.viewport(Cypress.env("CI") ? 1410 : 1400, 1080);
```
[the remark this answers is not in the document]
minnakt: no, it's not possible to set both to the same number because 1410 is just enough to change the number of visible tasks on the local Mac environment

apps/spruce/src/pages/task/taskTabs/TaskHistory/index.tsx
minnakt: I think that it's safer to request 50 in case the user decides to expand the page (the content will look really short otherwise since we don't re-request on page resize events). It could also be an improvement made in a future ticket

apps/spruce/cypress/integration/task/task_history.ts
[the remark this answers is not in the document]
minnakt: I just tried editing webkit scrollbar (not sure if that was a good idea) but this is another option, let me know what you think

apps/spruce/src/pages/task/taskTabs/TaskHistory/index.tsx
minnakt: oh yeah no data should be missing... if it's missing that is a bug. Let me know if you find any issues with missing data!

SupaJoon: I tested this a bunch more and found that the taskbox indices are preserved when paginating regardless of page width!

SupaJoon (approved):

apps/spruce/cypress/integration/task/task_history.ts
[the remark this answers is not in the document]
minnakt: ok I actually decided styling the scrollbar is a really bad idea, so I'm going to put it back the way it was, but I can file a ticket for the discrepancies between Mac and Linux screen sizes

PR #756, opened 2025-04-18
DEVPROD-16598 Add timechart view for variants/tasks.

ahartschen:
[DEVPROD-16598](https://jira.mongodb.org/browse/DEVPROD-16598)

### Description
This change adds a new tab to the version page that will display a Gantt chart of variants or tasks. If no variants are selected the view will show the chart of all variants. If a variant is selected then it will show a paginated chart of tasks within that variant.

The idea behind this view is that it will provide more context into build times than simply looking at task duration because task duration does not show waiting time. When looking at the Gantt chart you can see exactly when on a timeline a task started/finished.

#### Variants
image

#### Tasks
image

#### Clicking around showing loading, no data available, tasks, and variant views

Kapture 2025-04-18 at 14 09 14

### Testing
I wrote Cypress tests to verify that the two different views are visible. I added unit tests for utility methods.

Manual verification was also done.

### DIFF
--- a/apps/spruce/cypress/integration/version/version_timing.ts
+++ b/apps/spruce/cypress/integration/version/version_timing.ts
@@ -102,7 +102,7 @@ describe("Version Timing Tab with a variant selected", () => {
         page.forEach((task) => {
           expect(textFound).to.include(task);
         });
-        cy.get('[data-cy="next-page-button"]').click();
+        cy.dataCy("next-page-button").click();
       });
     });
 
@@ -113,7 +113,7 @@ describe("Version Timing Tab with a variant selected", () => {
         page.forEach((task) => {
           expect(textFound).to.include(task);
         });
-        cy.get('[data-cy="prev-page-button"]').click();
+        cy.dataCy("prev-page-button").click();
       });
     });
   });
@@ -136,7 +136,7 @@ describe("Version Timing Tab with a variant selected", () => {
   });
 
   it("allows the user to clear all filters", () => {
-    cy.get('[data-cy="clear-all-filters"]').click();
+    cy.dataCy("clear-all-filters").click();
 
     cy.url().should(
       "equal",
@@ -166,9 +166,9 @@ describe("Version Timing Tab with a variant selected", () => {
     cy.get("[id^=reactgooglegraph]").within(() => {
       cy.contains("test-agent").click();
     });
-    cy.url().should(
+    cy.location("pathname").should(
       "equal",
-      "http://localhost:3000/task/evergreen_ubuntu1604_test_agent_patch_5e823e1f28baeaa22ae00823d83e03082cd148ab_5e4ff3abe3c3317e352062e4_20_02_21_15_13_48/logs?execution=0",
+      "/task/evergreen_ubuntu1604_test_agent_patch_5e823e1f28baeaa22ae00823d83e03082cd148ab_5e4ff3abe3c3317e352062e4_20_02_21_15_13_48/logs",
     );
   });
 });
--- a/apps/spruce/src/pages/version/Tabs/VersionTiming/index.tsx
+++ b/apps/spruce/src/pages/version/Tabs/VersionTiming/index.tsx
@@ -31,11 +31,12 @@ interface Props {
   taskCount: number;
 }
 
+const defaultSort = `${TaskSortCategory.Duration}:${SortDirection.Desc}`;
+
 const VersionTiming: React.FC<Props> = ({ taskCount, versionId }) => {
   const dispatchToast = useToastContext();
   const { search } = useLocation();
 
-  const defaultSort = `${TaskSortCategory.Duration}:${SortDirection.Desc}`;
   const [queryParams, setQueryParams] = useQueryParams();
   const versionAnalytics = useVersionAnalytics(versionId);
   const queryVariables = useQueryVariables(search, versionId);
--- a/apps/spruce/src/pages/version/Tabs/VersionTiming/types.ts
+++ b/apps/spruce/src/pages/version/Tabs/VersionTiming/types.ts
@@ -1,46 +1,13 @@
 import { Unpacked } from "@evg-ui/lib/types/utils";
 import { VersionTaskDurationsQuery } from "gql/generated/types";
 
-type GanttChartColumnHeader = {
-  type: "string" | "date" | "number";
-  label: string;
-};
-
 export type DateTimeRange = {
   start: Date;
   finish: Date;
   buildVariantDisplayName: string;
 };
 
-type GanttChartColumnHeaders = [
-  GanttChartColumnHeader,
-  GanttChartColumnHeader,
-  GanttChartColumnHeader,
-  GanttChartColumnHeader,
-  GanttChartColumnHeader,
-  GanttChartColumnHeader,
-  GanttChartColumnHeader,
-  GanttChartColumnHeader,
-];
-
-export type GanttChartDataRow = [
-  string,
-  string,
-  string | null,
-  Date,
-  Date,
-  number | null,
-  number,
-  string | null,
-];
-
-export type TaskDurationData = Unpacked<
-  VersionTaskDurationsQuery["version"]["tasks"]["data"]
->;
-
-export type GanttChartData = [GanttChartColumnHeaders, ...GanttChartDataRow[]];
-
-export const GANTT_CHART_COLUMN_HEADERS: GanttChartColumnHeaders = [
+export const GANTT_CHART_COLUMN_HEADERS = [
   { type: "string", label: "Task ID" },
   { type: "string", label: "Task Name" },
   { type: "string", label: "Resource" },
@@ -49,4 +16,55 @@ export const GANTT_CHART_COLUMN_HEADERS: GanttChartColumnHeaders = [
   { type: "number", label: "Duration" },
   { type: "number", label: "Percent Complete" },
   { type: "string", label: "Dependencies" },
+] as const;
+
+type GanttChartColumnHeaders = typeof GANTT_CHART_COLUMN_HEADERS;
+
+/**
+ * GanttChartDataRow represents a single row of data in the Gantt chart.
+ * Each row contains information about a task or variant, including its
+ * ID, name, assigned resource, start and end dates, duration, completion
+ * percentage, and dependencies.
+ *
+ *
+ */
+export type GanttChartDataRow = [
+  /**
+   * A unique identifier for the row.
+   */
+  id: string,
+  /**
+   * The name or description of the row (e.g., task or variant name).
+   */
+  displayName: string,
+  /**
+   * The resource assigned to the row. Can be null if no resource is assigned.
+   */
+  resource: string | null,
+  /**
+   * The date and time when the resource started running.
+   */
+  startDate: Date,
+  /**
+   * The date and time when the resource finished running.
+   */
+  endDate: Date,
+  /**
+   * The duration of the task in milliseconds. Can be null if the duration should be calculated from startDate and endDate.
+   */
+  duration: number | null,
+  /**
+   * The percentage of the task that has been completed, represented as a number between 0 and 100.
+   */
+  percentComplete: number,
+  /**
+   * A comma-separated list of IDs that this resource depends on. Can be null if there are no dependencies.
+   */
+  dependencies: string | null,
 ];
+
+export type TaskDurationData = Unpacked<
+  VersionTaskDurationsQuery["version"]["tasks"]["data"]
+>;
+
+export type GanttChartData = [GanttChartColumnHeaders, ...GanttChartDataRow[]];
--- a/yarn.lock
+++ b/yarn.lock
@@ -2144,18 +2144,6 @@
   resolved "https://registry.yarnpkg.com/@humanwhocodes/retry/-/retry-0.4.1.tgz#9a96ce501bc62df46c4031fbd970e3cc6b10f07b"
   integrity sha512-c7hNEllBlenFTHBky65mhq8WD2kbN9Q6gk0bTk8lSBvc554jpXSkST1iePudpt7+A/AQvuHs9EMqjHDXMY1lrA==
 
-"@isaacs/cliui@^8.0.2":
-  version "8.0.2"
-  resolved "https://registry.yarnpkg.com/@isaacs/cliui/-/cliui-8.0.2.tgz#b37667b7bc181c168782259bab42474fbf52b550"
-  integrity sha512-O8jcjabXaleOG9DQ0+ARXWZBTfnP4WNAqzuiJK7ll44AmxGKv/J2M4TPjxjY3znBCfvBXFzucm1twdyFybFqEA==
-  dependencies:
-    string-width "^5.1.2"
-    string-width-cjs "npm:string-width@^4.2.0"
-    strip-ansi "^7.0.1"
-    strip-ansi-cjs "npm:strip-ansi@^6.0.1"
-    wrap-ansi "^8.1.0"
-    wrap-ansi-cjs "npm:wrap-ansi@^7.0.0"
-
 "@joshwooding/vite-plugin-react-docgen-typescript@0.5.0":
   version "0.5.0"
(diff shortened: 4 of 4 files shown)


apps/spruce/src/pages/version/Tabs/TimeChart/VariantTimeChart/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { usePolling } from "hooks";
+import { useQueryVariables } from "../../useQueryVariables";
+import TimeChartGraph from "../TimeChartGraph";
+import { transformVersionDataToVariantGanttChartData } from "../utils";
+
+interface Props {
+  versionId: string;
+}
+
+const VariantTimeChart: React.FC<Props> = ({ versionId }) => {
```
khelif96, line 19: I don't its necessary to have a second VariantTimeChart component. This is just a subset of whats shown in TaskTimeChart. We can just bake the additional variant filtering logic into the original component and reuse it.

apps/spruce/src/pages/version/Tabs/TimeChart/TaskTimeChart/index.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const { version } = data || {};
+  const { tasks } = version || {};
+  const { count = 0, data: tasksData = [] } = tasks || {};
+
+  return (
+    <>
+      <TableControl
+        filteredCount={count}
+        label="tasks"
+        // @ts-expect-error: FIXME. This comment was added by an automated script.
```
khelif96, line 65: Can you address the ts errors. Assigning a default value should be enough to suppress these errors.

apps/spruce/src/pages/version/Tabs/TimeChart/TimeChartGraph/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  return (
+    <Chart
+      chartType="Gantt"
+      data={data}
+      height={Math.max(
+        data.length * CHART_ROW_HEIGHT_IN_PIXELS,
+        MINIMUM_CHART_HEIGHT_IN_PIXELS,
+      )}
+      options={{
```
khelif96, line 38: The time labels on the x-axis don't mean anything since they don't have any associated dates.  12pm and 12am show up twice in this example but they aren't associated with any particular date.
image

apps/spruce/src/pages/version/Tabs/TimeChart/TimeChartGraph/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+interface Props {
+  data: GanttChartData;
+  loading: boolean;
+}
+
+const TimeChartGraph: React.FC<Props> = ({ data, loading }) => {
+  if (loading || !data || data.length === 1) {
+    return (
+      <div
```
khelif96, line 17: I think a LoadingSkeleton would work well here.
https://www.mongodb.design/component/skeleton-loader/live-example

apps/spruce/src/pages/version/Tabs/TimeChart/types.ts
```diff
@@ -0,0 +1,52 @@
+import { Unpacked } from "@evg-ui/lib/types/utils";
+
```
khelif96, line 2: nit. remove the space

khelif96 (changes requested):
Great work so far I had a few comments on how these components are set up and styled. 

On my super wide monitor I'm noticing task names get cut off. This would be even worse on smaller screens. Can we do something about it? [Example](https://spruce-local.corp.mongodb.com:8443/version/mongodb_mongo_master_f5dfc887f947751367eb8195ecce450fd5464e3b/timechart?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)
image

apps/spruce/src/pages/version/Tabs/TimeChart/VariantTimeChart/index.tsx
khelif96: I think the same query can be used for both. `VersionTaskDurations` takes in variant as a parameter so you can technically just reuse the same query and add an optional variant filter. Its how the task-duration tab works.

khelif96:
> > On my super wide monitor I'm noticing task names get cut off. This would be even worse on smaller screens. Can we do something about it?
> 
> I can not figure out how to change the wrapping/underflow of the text within the component.

There appears to be some additional params for the Chart component that can be used to update the styles 
https://developers.google.com/chart/interactive/docs/gallery/ganttchart#configuration-options Do any of these work?

apps/spruce/src/pages/version/Tabs/VersionTiming/VersionTimingGraph/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+      <div
+        style={{
+          display: "flex",
+          justifyContent: "center",
+          alignItems: "center",
+          height: `${MINIMUM_CHART_HEIGHT_IN_PIXELS}px`,
+        }}
+      >
+        <p>No data available</p>
+      </div>
```
khelif96, line 46: Can we use emotion to style this element? 

it would look something like this.

Declare the styled component. 
```ts
const Container = styled.div`
display: flex;
justify-content: center;
align-items: center;
height: ${MINIMUM_CHART_HEIGHT_IN_PIXELS}px;
`
```

And then you can just reuse the component .
```suggestion
      <Container>
              <p>No data available</p>
      </Container>

```

apps/spruce/src/pages/version/Tabs/VersionTiming/VersionTimingGraph/index.tsx
```diff
@@ -0,0 +1,121 @@
...
+      options={{
+        timeline: {
+          groupByRowLabel: false,
+          showRowLabels: false,
+        },
+
+        gantt: {
+          criticalPathEnabled: false,
+          labelMaxWidth: labelMaxWidth,
+          labelStyle: {
```
khelif96, line 103: Can we apply these font styles to these tooltips as well
image

apps/spruce/src/pages/version/Tabs/VersionTiming/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+          onPageSizeChange={(l) => {
+            versionAnalytics.sendEvent({
+              name: "Changed page size",
+              "page.size": l,
+            });
+          }}
+          page={page || 0}
+          totalCount={taskCount}
+        />
+      )}
```
khelif96, line 117: As a general design principle it's kind of jarring when large navigational UI elements are suddenly removed or added to the same view. I think we can keep this consistent and just disable the buttons when they are not modifiable. 


https://github.com/user-attachments/assets/55470c07-d6f4-4c26-b52c-ef5c33ed5cea

apps/spruce/src/pages/version/Tabs/VersionTiming/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+    ? transformTaskDurationDataToTaskGanttChartData(tasksData)
+    : transformTaskDurationDataToVariantGanttChartData(tasksData);
+
+  return (
+    <>
+      <Body>
+        This page provides a timeline view of{" "}
+        {hasBuildVariantQueryParam ? "task" : "variant"} wall clock duration in
+        the version.
+      </Body>
```
khelif96, line 100: How do you feel about this description instead
```suggestion
  const description = isVariantTimingView ? (
    <>
      This page is showing a timeline view of task run times in the{" "}
      <b>{queryParams.variant}</b> variant. This is a Gantt chart showing when
      each task started and finished running. You can click on a task to visit
      the task page.
    </>
  ) : (
    "This page is showing a timeline view of variant run times in this version. This is a Gantt chart showing when each variant started and finished running. You can click on a variant to see a view of the tasks that ran."
  );
  return (
    <>
      <H2>{isVariantTimingView ? "Variant" : "Version"} Timing</H2>
      <StyledBody>{description}</StyledBody>
```

When viewing the overall version and its variants
image

When viewing a variant
image

apps/spruce/src/pages/version/Tabs/VersionTiming/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+const VersionTiming: React.FC<Props> = ({ taskCount, versionId }) => {
+  const dispatchToast = useToastContext();
+  const { search } = useLocation();
+
+  const defaultSort = `${TaskSortCategory.Duration}:${SortDirection.Desc}`;
+  const [queryParams, setQueryParams] = useQueryParams();
+  const versionAnalytics = useVersionAnalytics(versionId);
+  const queryVariables = useQueryVariables(search, versionId);
+  const hasQueryVariables = Object.keys(parseQueryString(search)).length > 0;
+  const hasBuildVariantQueryParam = !!queryParams.variant;
```
khelif96, line 42: This might be a more descriptive name for this variable. 
```suggestion
  const isVariantTimingView = !!queryParams.variant;
```

apps/spruce/src/pages/version/Tabs/VersionTiming/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+  };
+
+  const { data, loading, refetch, startPolling, stopPolling } = useQuery<
+    VersionTaskDurationsQuery,
+    VersionTaskDurationsQueryVariables
+  >(VERSION_TASK_DURATIONS, {
+    variables: {
+      ...queryVariables,
+      taskFilterOptions: {
+        ...queryVariables.taskFilterOptions,
```
khelif96, line 71: I find the fact that we are using the filters from the tasks table in the gantt chart but there isn't any indication that we actually are, pretty confusing. I think we should provide those inputs to users on the page or not use them at all.

khelif96 (changes requested):

apps/spruce/cypress/integration/version/version_timing.ts
```diff
@@ -0,0 +1,174 @@
+import { clickOnPageSizeBtnAndAssertURLandTableSize } from "../../utils";
+
+describe("Version Timing Tab without a variant selected", () => {
+  beforeEach(() => {
+    cy.visit("/version/5e4ff3abe3c3317e352062e4/version-timing");
+  });
+  it("shows a chart of all variants in the version", () => {
```
SupaJoon, line 7: This doesn't show all the variants. The only visible variant in this test is Ubuntu 16.04. I found that out by running the test with `cy.pause()` at the bottom of the test body.

apps/spruce/cypress/integration/version/version_timing.ts
```diff
@@ -0,0 +1,174 @@
...
+    const expectedVariants = ["Ubuntu 16.04", "Lint", "Race Detector"];
+    cy.get("svg > g > text").then(($items) => {
+      const textFound = Array.from($items, (item) => item.innerHTML);
+      expectedVariants.forEach((variant) => {
+        expect(textFound).to.include(variant);
+      });
+    });
+  });
+  it("allows the user to select a variant and see the tasks in it", () => {
```
SupaJoon, line 16: The test doesn't assert the task visibility. You can change the test name or assert that the tasks are visible.

apps/spruce/cypress/integration/version/version_timing.ts
```diff
@@ -0,0 +1,174 @@
...
+      });
+    });
+  });
+
+  it("allows the user to select a task and navigate to it", () => {
+    cy.get("[id^=reactgooglegraph]").within(() => {
+      cy.contains("test-agent").click();
+    });
+    cy.url().should(
+      "include",
```
SupaJoon, line 170: Can you make both assertions use "include" or "equal"

apps/spruce/cypress/tsconfig.json
```diff
@@ -5,7 +5,7 @@
     "lib": [
       "es5",
       "dom",
-      "es2017"
+      "es2019"
```
SupaJoon, line 8: Why did you this need to be bumped?

SupaJoon (commented):
Clicking on the build variant box takes you away from the version timing tab. Is that intentional?

apps/spruce/src/pages/version/Tabs/VersionTiming/types.ts
```diff
@@ -0,0 +1,52 @@
...
+export type GanttChartDataRow = [
+  string,
+  string,
+  string | null,
+  Date,
+  Date,
+  number | null,
+  number,
+  string | null,
+];
```
khelif96, line 35:
```suggestion
/**
 * GanttChartDataRow represents a single row of data in the Gantt chart.
 * Each row contains information about a task or variant, including its
 * ID, name, assigned resource, start and end dates, duration, completion
 * percentage, and dependencies.
 *
 *
 */

export type GanttChartDataRow = [
  /**
   * A unique identifier for the row.
   */
  id: string,
  /**
   * The name or description of the row (e.g., task or variant name).
   */
  displayName: string,
  /**
   * The resource assigned to the row. Can be null if no resource is assigned.
   */
  resource: string | null,
  /**
   * The date and time when the resource started running.
   */
  startDate: Date,
  /**
   * The date and time when the resource finished running.
   */
  endDate: Date,
  /**
   * The duration of the task in milliseconds. Can be null if the duration is not available.
   */
  duration: number | null,
  /**
   * The percentage of the task that has been completed, represented as a number between 0 and 100.
   */
  percentComplete: number,
  /**
   * A comma-separated list of IDs that this resource depends on. Can be null if there are no dependencies.
   */
  dependencies: string | null,
];

```

This will make intellisense a little smarter and help the implementer know what each field refers to
image

apps/spruce/src/pages/version/Tabs/VersionTiming/types.ts
```diff
@@ -0,0 +1,52 @@
...
+export const GANTT_CHART_COLUMN_HEADERS: GanttChartColumnHeaders = [
+  { type: "string", label: "Task ID" },
+  { type: "string", label: "Task Name" },
+  { type: "string", label: "Resource" },
+  { type: "date", label: "Start Date" },
+  { type: "date", label: "End Date" },
+  { type: "number", label: "Duration" },
+  { type: "number", label: "Percent Complete" },
+  { type: "string", label: "Dependencies" },
+];
```
khelif96, line 52: This can be rewritten with `as const` and will preserve the type safety while removing the redundant `GanttChartColumnHeaders` typing
```suggestion
export const GANTT_CHART_COLUMN_HEADERS = [
  { type: "string", label: "Task ID" },
  { type: "string", label: "Task Name" },
  { type: "string", label: "Resource" },
  { type: "date", label: "Start Date" },
  { type: "date", label: "End Date" },
  { type: "number", label: "Duration" },
  { type: "number", label: "Percent Complete" },
  { type: "string", label: "Dependencies" },
] as const;

 type GanttChartColumnHeaders = typeof GANTT_CHART_COLUMN_HEADERS;
```

apps/spruce/src/pages/version/Tabs/VersionTiming/types.ts
```diff
@@ -0,0 +1,52 @@
...
+type GanttChartColumnHeaders = [
+  GanttChartColumnHeader,
+  GanttChartColumnHeader,
+  GanttChartColumnHeader,
+  GanttChartColumnHeader,
+  GanttChartColumnHeader,
+  GanttChartColumnHeader,
+  GanttChartColumnHeader,
+  GanttChartColumnHeader,
+];
```
khelif96, line 24: Same comment as below

apps/spruce/src/pages/version/Tabs/VersionTiming/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+interface Props {
+  versionId: string;
+  taskCount: number;
+}
+
+const VersionTiming: React.FC<Props> = ({ taskCount, versionId }) => {
+  const dispatchToast = useToastContext();
+  const { search } = useLocation();
+
+  const defaultSort = `${TaskSortCategory.Duration}:${SortDirection.Desc}`;
```
khelif96, line 38: Nit. you can move this outside the component like we do in TaskDuration since it never changes.

PR #766, opened 2025-04-22
DEVPROD-17033 Remove mainline commits page [major]

khelif96:
DEVPROD-17033

### Description
Gets rid of the mainline commits page and all associated code. 

The tests will fail until #765 is merged in.

Do not merge until May 6th


apps/spruce/cypress/integration/projectHealth/variant_history.ts
```diff
@@ -5,16 +5,12 @@ describe("variant history", () => {
...
 
-  it("should link to a specific commit from the project health page", () => {
-    cy.visit("/commits/spruce");
-    cy.dataCy("variant-header").should("exist");
-    cy.dataCy("variant-header").should("contain.text", "Ubuntu 16.04");
-    cy.dataCy("variant-header").first().click();
+  it("should link to variant history from the waterfall page", () => {
+    cy.visit("/project/spruce/waterfall");
+    cy.dataCy("build-variant-link").should("exist");
+    cy.dataCy("build-variant-link").should("contain.text", "Ubuntu 16.04");
```
minnakt, line 11: nit
```suggestion
    cy.dataCy("build-variant-link").first().should("contain.text", "Ubuntu 16.04");
```

apps/spruce/src/gql/fragments/projectSettings/viewsAndFilters.graphql
```diff
@@ -5,7 +5,6 @@ fragment ProjectViewsAndFiltersSettings on Project {
     exactMatch
     expression
   }
-  projectHealthView
```
minnakt, line 8: Should we create a ticket to remove this setting? It was designed specifically for the Project Health page... and then no one ever used it. 😭😭

apps/spruce/src/gql/fragments/projectSettings/viewsAndFilters.graphql
minnakt, line 1: Besides this field, would we also be able to stop querying for `hasUsedMainlineCommitsBefore` and `hasUsedSpruceBefore` from `SpruceConfig`?

minnakt (approved):

apps/spruce/src/gql/fragments/projectSettings/viewsAndFilters.graphql
khelif96: :( yes! I'll do it along side the other mainline commits deprecation work

PR #769, opened 2025-04-24
DEVPROD-16174: Create Inactive Commits Button

SupaJoon:
DEVPROD-16174

### Description
These code changes create buttons to expand and collapse inactive commits.
Screenshot 2025-04-24 at 10 39 00 AM


apps/spruce/src/pages/task/taskTabs/TaskHistory/index.tsx
```diff
@@ -89,17 +94,41 @@ const TaskHistory: React.FC<TaskHistoryProps> = ({ task }) => {
...
 
-  const prevPageCursor = getPrevPageCursor(visibleTasks[0]);
+  const prevPageCursor = getPrevPageCursor(timelineTasks[0]);
   const nextPageCursor = getNextPageCursor(
-    visibleTasks[visibleTasks.length - 1],
+    timelineTasks[timelineTasks.length - 1],
+  );
+
+  const commitDetailsList: GroupedTask[] = expandVisibleInactiveTasks(
+    timelineTasks,
```
minnakt, line 130: Is there a reason all this logic can't be self-contained within a component? e.g.

```
import { useState } from "react";
import styled from "@emotion/styled";
import Button from "@leafygreen-ui/button";
import Icon from "@leafygreen-ui/icon";
import { Size } from "@leafygreen-ui/tokens";
import pluralize from "pluralize";
import CommitDetailsCard from "../CommitDetailsCard";
import { TaskHistoryTask } from "../types";

interface InactiveCommitsProps {
  inactiveTasks: TaskHistoryTask[];
  currentTaskId: string;
  owner: string | undefined;
  repo: string | undefined;
}

const InactiveCommits: React.FC<InactiveCommitsProps> = ({
  currentTaskId,
  inactiveTasks,
  owner,
  repo,
}) => {
  const [isExpanded, setIsExpanded] = useState(false);

  return (
    <>
      <StyledButton
        data-cy="collapsed-card"
        leftGlyph={
          isExpanded ? (
            <Icon glyph="ChevronDown" />
          ) : (
            <Icon glyph="ChevronRight" />
          )
        }
        onClick={() => setIsExpanded(!isExpanded)}
        size={Size.XSmall}
      >
        {inactiveTasks.length}{" "}
        {isExpanded
          ? "EXPANDED"
          : pluralize("INACTIVE COMMIT", inactiveTasks.length)}
      </StyledButton>
      {isExpanded
        ? inactiveTasks.map((t) => (
            <CommitDetailsCard
              key={t.id}
              isCurrentTask={t.id === currentTaskId}
              owner={owner}
              repo={repo}
              task={t}
            />
          ))
        : null}
    </>
  );
};

const StyledButton = styled(Button)`
  width: fit-content;
`;

export default InactiveCommits;
```

The component tracks its own expanded state and is rendered as a child of `CommitDetailsList`.

I guess the component won't keep its expanded state between pages, but that seems fine to me.

minnakt (changes requested):

apps/spruce/src/pages/task/taskTabs/TaskHistory/index.tsx
SupaJoon: The main benefits of the current implementation is that the CommitDetails view is completely controlled by the input data so that lets us do things like initiate and persist state. For example, this implementation enables
- persisting state between pages
- allows the storybook to show expanded state without user interaction
- building a feature in the future where clicking on an inactive commit task box in the timeline scrolls and expands inactive commits.

It is a little bit more code but it's more functional and give us better flexibility over view states. Let me know what you think.

minnakt: I would still prefer the component. While I agree with your points that the current implementation is more amenable to being expanded in the future, I also think it is a premature optimization. Using the component will simplify the code a lot and make it more maintainable; in case we do need to update the behavior, we can always come back to this implementation.

minnakt: State can also be expanded without user interaction through the use of a prop.

apps/spruce/src/pages/task/taskTabs/TaskHistory/CommitDetailsList/index.tsx
```diff
@@ -33,12 +36,13 @@ const CommitDetailsList: React.FC<CommitDetailsListProps> = ({
                 task={task}
               />
             );
-          } else if (t.inactiveTasks) {
-            // TODO DEVPROD-16174: Replace with Inactive Commits Button.
+          } else if (inactiveTasks && shouldCollapse) {
```
minnakt, line 39: `shouldCollapse` is likely redundant (since it's already part of the calculation for `groupTasks`)
```suggestion
          } else if (inactiveTasks) {
```

apps/spruce/src/pages/task/taskTabs/TaskHistory/CommitDetailsList/index.tsx
```diff
@@ -3,17 +3,20 @@ import { ParagraphSkeleton } from "@leafygreen-ui/skeleton-loader";
...
   currentTask: NonNullable<TaskQuery["task"]>;
   tasks: GroupedTask[];
   loading: boolean;
+  shouldCollapse: boolean;
 }
 
 const CommitDetailsList: React.FC<CommitDetailsListProps> = ({
   currentTask,
   loading,
+  shouldCollapse,
```
minnakt, line 19: (after above change)
```suggestion
}

const CommitDetailsList: React.FC<CommitDetailsListProps> = ({
  currentTask,
  loading,
```

apps/spruce/src/pages/task/taskTabs/TaskHistory/CommitDetailsList/index.tsx
```diff
@@ -33,12 +36,13 @@ const CommitDetailsList: React.FC<CommitDetailsListProps> = ({
...
             );
-          } else if (t.inactiveTasks) {
-            // TODO DEVPROD-16174: Replace with Inactive Commits Button.
+          } else if (inactiveTasks && shouldCollapse) {
             return (
-              <span key={t.inactiveTasks[0].id} data-cy="collapsed-card">
-                {t.inactiveTasks.length} Collapsed
-              </span>
+              <InactiveCommitsButton
+                key={inactiveTasks[0].id}
```
minnakt, line 42: nit. could we `key` this by something else, perhaps the range 👀 I think the `key` might be duplicated due to also using it for the first card (if expanded).

e.g. `inactive-tasks-${inactiveTasks[0].id]}- ${inactiveTasks[inactiveTasks.length-1].id}` (or whatever else you can think of)

minnakt (approved):
thank you for exploring a bunch of different approaches for this ticket!

PR #772, opened 2025-04-25
DEVPROD-5935: Migrate Tasks table to LeafyGreen table

minnakt:
DEVPROD-5935

### Description
Delete the last antd table. This makes it so that [DEVPROD-1037](https://jira.mongodb.org/browse/DEVPROD-1037) is the only ticket left open with antd dependencies.

The reason we couldn't complete this ticket in the past was because the performance of expanding nested rows was really bad. But now that is fixed 😃

### Screenshots

Screenshot 2025-04-25 at 10 24 51 AM

### Testing
* Update Cypress tests 
* Add story
* Add unit tests


apps/spruce/src/pages/version/Tabs/Tasks/index.tsx
```diff
@@ -82,36 +80,17 @@ const Tasks: React.FC<Props> = ({ taskCount, versionId }) => {
...
-      />
-    </TableWrapper>
+    <VersionTasksTable
+      clearQueryParams={clearQueryParams}
+      filteredCount={count}
+      isPatch={isPatch ?? false}
+      limit={limit ?? 0}
+      loading={tasksData.length === 0 && loading}
+      page={page ?? 0}
+      // @ts-expect-error: Expects TaskTableInfo which overlaps with this type
```
khelif96, line 90: Would it be possible to fix the typing in https://github.com/evergreen-ci/ui/blob/02acfa1bd2fa369fd75c379f9a78701b4455f39f/apps/spruce/src/components/TasksTable/types.ts so that it reflects the real types? (And maybe file a ticket to get rid of the null or undefined stuff in evergreen)?

khelif96 (approved):
Thanks for doing this! One step closer to no more antd

PR #773, opened 2025-04-25
DEVPROD-12937 Add task ownership information to Task metadata

khelif96:
DEVPROD-12937

### Description
Adds a new section to task metadata that surfaces task ownership information. I put this in its own component and query since the FWS API may not always have something for a task so I wanted to prevent an error from surfacing.

### Screenshots
image


apps/spruce/src/pages/task/metadata/TaskOwnership/TaskOwnership.test.tsx
```diff
@@ -0,0 +1,223 @@
...
+> = {
+  request: {
+    query: TASK_OWNER_TEAM,
+    variables: { taskId, execution },
+  },
+  result: { data: undefined },
+};
+
+describe("TaskOwnership", () => {
+  it("renders nothing while loading", async () => {
```
minnakt, line 103: nit. doesn't need to be `async`

apps/spruce/src/pages/task/metadata/TaskOwnership/TaskOwnership.test.tsx
```diff
@@ -0,0 +1,223 @@
...
+      ).toBeInTheDocument();
+    });
+
+    expect(screen.getByText("Task Owner:")).toBeInTheDocument();
+    expect(screen.getByText("Platform Team")).toBeInTheDocument();
+
+    // Hover over the tooltip
+    act(() => {
+      screen.getByDataTestid("info-sprinkle-icon").focus();
+    });
```
minnakt, line 135: We should be using `userEvent` library instead to test hover
Not sure if you'll still need `act` if you're going to be `await`ing the `hover` event

apps/spruce/src/pages/task/metadata/Metadata.stories.tsx
minnakt, line 1: Can this field be represented in a story?

minnakt (approved):

PR #780, opened 2025-04-28
DEVPROD-16618: Improve restart task interaction on Task History tab

minnakt:
DEVPROD-16618

### Description
Right now, if you restart a task through task history timeline, a toast gets dispatched but nothing happens visually. This can be a confusing user experience for users who expect to see the timeline update in real time.

Instead of re-requesting the query every time, we can update the cached object. Except in the case of changing the URL, in which updating the cache doesn't really work well. But overall, this should be an uncommon scenario.

### Screenshots

https://github.com/user-attachments/assets/0b4e1fd5-6de3-4b2a-8a4e-02b8c58dc2a7

### Testing
- Added Cypress tests


khelif96 (approved):
Dope!

PR #783, opened 2025-05-01
DEVPROD-16185: Connect Commit Details interactions with Task Timeline

minnakt:
DEVPROD-16185

### Description
Implements the following interactions (confirmed with Will):
* When hovering over a commit card, the corresponding task box should be outlined in blue.
* When clicking on a task box, the corresponding commit details should be outlined in blue and jump into view.

### Screenshots

https://github.com/user-attachments/assets/9212e391-505c-4b27-b8ff-715815c17d62

### Testing
* Cypress tests

Note that I couldn't test some styles due to Cypress not being able to test `:hover` styles.
Screenshot 2025-05-01 at 11 25 14 AM


apps/spruce/src/pages/task/taskTabs/TaskHistory/TaskTimeline/index.tsx
```diff
@@ -68,8 +77,25 @@ const TaskTimeline = forwardRef<HTMLDivElement, TimelineProps>(
                     <TaskBox
                       key={task.id}
                       data-cy="timeline-box"
+                      hovered={hoveredTask === task.id}
+                      id={`task-box-${task.id}`}
+                      onClick={() => {
```
SupaJoon, line 82: nit: this function logic is easier to read outside of the prop assignment

SupaJoon (changes requested):
I think the UX could be smoother if the page scrolled to the "Task History Overview" automatically when clicking any task to prevent the mouse from moving away from task timeline. 

Uploading Screen Recording 2025-05-05 at 10.51.31 AM.mov…

khelif96 (approved):

PR #784, opened 2025-05-02
DEVPROD-16620: Implement "Jump to Date" feature for Task History

minnakt:
DEVPROD-16620

### Description
Implement the Jump to Date feature. It's a little complicated because JavaScript doesn't store timezone information in the `Date` object, so we have to calculate the correct UTC timestamp for midnight depending on the timezone. Let me know if you find a bug in the logic.

Also, expose helpful util [`waitForElementToBeRemoved`](https://testing-library.com/docs/dom-testing-library/api-async/#waitforelementtoberemoved) from our test library package.

### Screenshots

Note: design was tweaked because LeafyGreen doesn't expose these components from `DatePicker` independently.

https://github.com/user-attachments/assets/82897bbc-80cf-489c-a3d9-540c203e6246

### Testing
* Wrote Cypress tests
* Wrote unit tests for now shared `DateFilter` component

### Evergreen PR
* evergreen-ci/evergreen/pull/8987 (tests won't pass until merged)


apps/spruce/src/components/DateFilter/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+    }
+  };
+
+  return (
+    <DatePicker
+      data-cy="date-picker"
+      label={showLabel ? "Go to Date" : ""}
+      max={today}
+      // Testing environments should not have a minimum date restriction due to static test data.
+      min={isProduction() ? subDays(subYears(today, 1), 1) : undefined}
```
khelif96, line 41: This makes me sad :( 
I wonder if we should use [this](https://docs.cypress.io/api/commands/clock) in the future to ensure consistent dates so our tests don't get stale

khelif96 (approved):

apps/spruce/src/components/DateFilter/index.tsx
minnakt: I think we had looked at `cy.clock()` but there was some kind of problem with modals? Anyways, I will create a ticket for it

minnakt: Actually, nevermind, I do think it might be a good idea to keep it this way; or else we can't test the component on our local environments.

PR #791, opened 2025-05-07
DEVPROD-17568 Redirect legacy commits page to waterfall page

khelif96:
DEVPROD-17568

### Description
Introduce a temporary redirect to redirect links from external facing apps to the new waterfall. While deprecating the page I came to the realization that we still have external apps we do not own sending traffic to the project health page. This  introduces a temporary redirect to preserve links while we identify those apps and make the necessary updates.

=

### DIFF
--- a/apps/spruce/cypress/integration/waterfall/navigation.ts
+++ b/apps/spruce/cypress/integration/waterfall/navigation.ts
@@ -13,4 +13,12 @@ describe("navigation", () => {
     cy.location("pathname").should("equal", "/project/spruce/waterfall");
     cy.dataCy("waterfall-page").should("be.visible");
   });
+
+  it("is redirected to the waterfall page when a user visits a legacy route", () => {
+    cy.visit("/commits/evergreen");
+    cy.location("pathname").should("equal", "/project/evergreen/waterfall");
+    cy.visit("/commits/evergreen?taskNames=test");
+    cy.location("pathname").should("equal", "/project/evergreen/waterfall");
+    cy.location("search").should("contain", "tasks=test");
+  });
 });
--- a/apps/spruce/src/analytics/waterfall/useWaterfallAnalytics.ts
+++ b/apps/spruce/src/analytics/waterfall/useWaterfallAnalytics.ts
@@ -39,7 +39,7 @@ type Action =
       name: "Viewed waterfall modal";
       navigated_to_waterfall: boolean;
     }
-  | { name: "System Event redirect"; referrer: string };
+  | { name: "Redirected to waterfall page"; referrer: string };
 
 export const useWaterfallAnalytics = () => {
   const { [slugs.projectIdentifier]: projectIdentifier } = useParams();
--- a/apps/spruce/src/analytics/waterfall/useWaterfallAnalytics.ts
+++ b/apps/spruce/src/analytics/waterfall/useWaterfallAnalytics.ts
@@ -38,7 +38,8 @@ type Action =
   | {
       name: "Viewed waterfall modal";
       navigated_to_waterfall: boolean;
-    };
+    }
+  | { name: "System Event redirect"; referrer: string };
 
 export const useWaterfallAnalytics = () => {
   const { [slugs.projectIdentifier]: projectIdentifier } = useParams();
--- a/apps/spruce/src/components/Redirects/WaterfallCommitsRedirect.tsx
+++ b/apps/spruce/src/components/Redirects/WaterfallCommitsRedirect.tsx
@@ -1,8 +1,46 @@
+import { useEffect } from "react";
 import { useParams, Navigate } from "react-router-dom";
+import { useWaterfallAnalytics } from "analytics";
 import { getWaterfallRoute, slugs } from "constants/routes";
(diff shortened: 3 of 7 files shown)


apps/spruce/src/components/Redirects/WaterfallCommitsRedirect.tsx
```diff
@@ -1,8 +1,46 @@
...
+    [],
+  );
+  const { sendEvent } = useWaterfallAnalytics();
+  useEffect(() => {
+    const { referrer } = document;
+    sendEvent({
+      name: "System Event redirect",
+      referrer,
+    });
+    console.log("System Event redirect", referrer);
```
minnakt, line 33:
```suggestion
```

apps/spruce/src/analytics/waterfall/useWaterfallAnalytics.ts
```diff
@@ -38,7 +38,8 @@ type Action =
   | {
       name: "Viewed waterfall modal";
       navigated_to_waterfall: boolean;
-    };
+    }
+  | { name: "System Event redirect"; referrer: string };
```
minnakt, line 42: I'm not sure what system event means but to be more in line with our other events I would reword it slightly
```suggestion
  | { name: "Redirected via system event"; referrer: string };
```

apps/spruce/src/constants/routes.ts
```diff
@@ -286,11 +287,20 @@ export const getDistroSettingsRoute = (
 
 export const getWaterfallRoute = (
   projectIdentifier?: string,
-  options?: { taskFilters?: string[] },
+  options?: {
+    statusFilters?: string[];
```
minnakt, line 291: just checking; the old commits page had umbrella statuses, the new page does not. will it be ok?

apps/spruce/src/components/Redirects/WaterfallCommitsRedirect.tsx
```diff
@@ -1,8 +1,46 @@
...
+    WaterfallFilterOptions.Task,
+    [],
+  );
+  const [requesterFilters] = useQueryParam<string[]>(
+    WaterfallFilterOptions.Requesters,
+    [],
+  );
+  const [variantFilters] = useQueryParam<string[]>(
+    WaterfallFilterOptions.BuildVariant,
+    [],
```
minnakt, line 24: the old commits page uses different query params for example `buildVariants` versus `variants`,  `taskNames` vs `tasks`, `requester` vs `requesters`. does this account for that?

minnakt (changes requested):
could we write an e2e test for this as well?

apps/spruce/src/analytics/waterfall/useWaterfallAnalytics.ts
khelif96: I forgot that redirected was one of our analytics event actions thank you!

apps/spruce/src/constants/routes.ts
khelif96: I think this ommission is ok. The goal is to primarily redirect external links and old links. The umbrella statuses always included the normal statuses as well so filters should still work the same. Besides that I don't think anyone actually filtered on status from what I've seen it was mainly task and variant filters being applied.

apps/spruce/src/components/Redirects/WaterfallCommitsRedirect.tsx
khelif96: Good catch!

apps/spruce/cypress/integration/waterfall/navigation.ts
```diff
@@ -13,4 +13,12 @@ describe("navigation", () => {
     cy.location("pathname").should("equal", "/project/spruce/waterfall");
     cy.dataCy("waterfall-page").should("be.visible");
   });
+
+  it("is redirected to the waterfall page when a user visits a legacy route", () => {
+    cy.visit("/commits/evergreen");
+    cy.location("pathname").should("equal", "/project/evergreen/waterfall");
+    cy.visit("/commits/evergreen?taskNames=test");
```
minnakt, line 20: nit. could we add variant just for completeness

apps/spruce/src/constants/routes.ts
minnakt: you could filter out using `SortedTaskStatus` if desired to generate the most "correct" link but sounds ok

apps/spruce/src/components/Redirects/WaterfallCommitsRedirect.tsx
```diff
@@ -1,8 +1,32 @@
...
+  const [requesterFilters] = useQueryParam<string[]>("requester", []);
+  const [variantFilters] = useQueryParam<string[]>("buildVariants", []);
+  const { sendEvent } = useWaterfallAnalytics();
+  useEffect(() => {
+    const { referrer } = document;
+    sendEvent({
+      name: "Redirected to waterfall page",
+      referrer,
+    });
+  }, []);
```
minnakt, line 20: nit. ignore or fix eslint warning

minnakt (approved):

PR #796, opened 2025-05-09
DEVPROD-16186: Support caching for Task History query

minnakt:
DEVPROD-16186

### Description
Currently, when you page forwards and then backwards, a network request is dispatched for both. This is odd because you are paging back to a page that you should have already fetched data for.

This PR introduces caching to reduce the number of network requests and give more immediate feedback to users. Note that this implementation tends to eagerly refetch.

### How to test this PR
I would recommend testing on staging or production. Go to any task, page forwards 2x or more, and then page backwards; you should be able to see that there's no loading state or network request, and the page is reading from the cache. 

(_Note: I would **not** recommend testing on local environment. It doesn't have enough activated tasks to accurately represent the behavior._)

### Testing
- Unit tests


apps/spruce/src/pages/task/taskTabs/TaskHistory/caching/index.ts
```diff
@@ -0,0 +1,170 @@
...
+  readField: ReadFieldFunction,
+) => {
+  for (let i = 0; i < tasks.length; i++) {
+    const activated = readField<boolean>("activated", tasks[i]) ?? false;
+    if (activated) {
+      return false;
+    }
+  }
+  return true;
+};
```
khelif96, line 22: nit you can probably just rewrite this to
```suggestion
): boolean => tasks.every((t) => !readField<boolean>("activated", t));
```

apps/spruce/src/pages/task/taskTabs/TaskHistory/caching/index.ts
```diff
@@ -0,0 +1,170 @@
...
+  if (activeTasks.size < limit) {
+    const allTaskOrders =
+      readField<Set<number>>("allTaskOrders", existing) ?? new Set<number>();
+    if (pagingBackwards && !allTaskOrders.has(mostRecentTaskOrder)) {
+      return undefined;
+    }
+    if (pagingForwards && !allTaskOrders.has(oldestTaskOrder)) {
+      return undefined;
+    }
+  }
```
khelif96, line 120: (optional comment) This has a bit of repition and I think we can extract this out into utils and DRY out the logic a bit.

```suggestion
  let endIndex = existingTasks.length - 1;
  let pagingDirection: Direction | null = null;

  if (direction === TaskHistoryDirection.After) pagingDirection = "backwards";
  if (direction === TaskHistoryDirection.Before) pagingDirection = "forwards";

  if (pagingDirection) {
    ({ endIndex, startIndex } = getPaginationIndices(
      [...existingTasks],
      readField,
      idx,
      limit,
      pagingDirection,
      includeCursor,
    ));
  }

  // Check if we have enough active tasks
  const activeTasksCount = existingTasks
    .slice(startIndex, endIndex + 1)
    .filter((t) => readField<boolean>("activated", t)).length;

  if (activeTasksCount < limit) {
    const allTaskOrders =
      readField<Set<number>>("allTaskOrders", existing) ?? new Set<number>();
    if (
      (pagingDirection === "backwards" &&
        !allTaskOrders.has(mostRecentTaskOrder)) ||
      (pagingDirection === "forwards" && !allTaskOrders.has(oldestTaskOrder))
    ) {
      return undefined;
    }
  }
```

khelif96: These are the utility functions for the suggestion above.
```ts
type Direction = "backwards" | "forwards";
interface Indices {
  startIndex: number;
  endIndex: number;
}

const getPaginationIndices = (
  tasks: TaskHistoryTask[],
  readField: ReadFieldFunction,
  cursorIdx: number,
  limit: number,
  direction: Direction,
  includeCursor: boolean,
): Indices =>
  direction === "backwards"
    ? paginateBackwards(tasks, readField, cursorIdx, limit, includeCursor)
    : paginateForwards(tasks, readField, cursorIdx, limit, includeCursor);

const paginateBackwards = (
  tasks: TaskHistoryTask[],
  readField: ReadFieldFunction,
  cursorIdx: number,
  limit: number,
  includeCursor: boolean,
): Indices => {
  const startAt = includeCursor ? cursorIdx : cursorIdx - 1;
  const boundary = findBoundary(tasks, readField, startAt, -1, limit);
  const endIndex = startAt;
  const startIndex =
    boundary == null || isAllInactive(tasks.slice(0, boundary), readField)
      ? 0
      : boundary;
  return { startIndex, endIndex };
};

const paginateForwards = (
  tasks: TaskHistoryTask[],
  readField: ReadFieldFunction,
  cursorIdx: number,
  limit: number,
  includeCursor: boolean,
): Indices => {
  const startAt = includeCursor ? cursorIdx : cursorIdx + 1;
  const boundary = findBoundary(tasks, readField, startAt, +1, limit);
  const startIndex = startAt;
  const endIndex =
    boundary == null || isAllInactive(tasks.slice(boundary + 1), readField)
      ? tasks.length - 1
      : boundary;
  return { startIndex, endIndex };
};

const findBoundary = (
  tasks: TaskHistoryTask[],
  readField: ReadFieldFunction,
  idx: number,
  step: 1 | -1,
  limit: number,
): number | null => {
  const seenIds = new Set<string>();
  for (let i = idx; i >= 0 && i < tasks.length; i += step) {
    if (readField<boolean>("activated", tasks[i])) {
      seenIds.add(readField<string>("id", tasks[i]) ?? "");
      if (seenIds.size === limit) return i;
    }
  }
  return null;
};

```

khelif96 (approved):
This works really well! I left some optional suggestions below on how I think it can be made a bit more readable but I don't have a strong opinion on if you use them or not.

PR #804, opened 2025-05-19
DEVPROD-17458 Remove reviewers field from dependabot config

khelif96:
DEVPROD-17458

### Description
Rely on the codeowners file instead.


khelif96:
> i didnt even know we had a codeowners file

Its used to auto tag us on prs

PR #816, opened 2025-05-22
DEVPROD-16863 Allow setting of providerAccount field

khelif96:
DEVPROD-16863

### Description
This adds the providerAccount field to the distro settings UI. This field only works for ec2 instances so I built it to only support those field types

### Screenshots
image
image

### Evergreen PR
https://github.com/evergreen-ci/evergreen/pull/9052/files


apps/spruce/src/pages/distroSettings/tabs/ProviderTab/getFormSchema.ts
```diff
@@ -9,18 +9,21 @@ import {
...
 import { textAreaCSS } from "./styles";
 
+const ec2Providers = [Provider.Ec2Fleet, Provider.Ec2OnDemand];
 export const getFormSchema = ({
   awsRegions,
   fleetRegionsInUse,
   onDemandRegionsInUse,
   poolMappingInfo,
   pools,
+  providerName,
```
minnakt, line 19: can we pass this information as a variable `isEC2Provider` instead?

apps/spruce/src/pages/distroSettings/tabs/ProviderTab/getFormSchema.ts
```diff
@@ -242,3 +248,9 @@ export const getFormSchema = ({
     },
   },
 });
+
+const providerAccountField = {
+  type: "string" as const,
+  title: "Provider Account",
+  default: "",
+};
```
minnakt, line 256: can we put this in `schemaFields.ts` with a comment that it's only for EC2 providers (or a name it something like `ec2ProviderAccountField`)

minnakt (approved):

PR #834, opened 2025-05-30
DEVPROD-11739 Only run parallel e2e tasks if we have changes in the spruce directory or the lib directory

khelif96:
DEVPROD-11739

### Description
This PR optimizes our CI pipeline by making E2E tests conditional on relevant code changes. It introduces a check that only runs E2E tests when there are modifications in either the app-specific directory or the shared packages directory.

## Changes
- Created a new `git-utils.js` module to centralize git-related functionality:
  - `getMergeBase()`: Gets the merge base between current branch and main
  - `whatChanged()`: Returns list of modified files in current PR/patch
  - `hasChangesInDirectory()`: Utility to check for changes in specific directories

- Modified `generate-parallel-e2e-tasks.js` to:
  - Check for changes in the relevant app directory (e.g., `apps/spruce`) or shared packages
  - Skip E2E task generation if no relevant changes are detected

### Testing
- Verified that E2E tasks are generated when changes exist in app/spruce directory
- Verified that E2E tasks are generated when changes exist in shared packages
- Verified that E2E tasks are skipped when changes are in unrelated directories
- Confirmed empty task list is written when skipping to maintain file structure

PR #845, opened 2025-06-06
Revert "DEVPROD-11739 Only run parallel e2e tasks if we have changes in the spruce directory or the lib directory"

minnakt:
Reverts evergreen-ci/ui#834


minnakt:
evergreen retry

PR #846, opened 2025-06-09
DEVPROD-11739 REVERT REVERT Only run parallel e2e tasks if we have changes

khelif96:
DEVPROD-11739

### Description
Fixes the bug in #845 by checking if it is a trigger version and if so it will always run the tasks.


.evergreen/scripts/generate-parallel-e2e-tasks.js
```diff
@@ -140,15 +143,31 @@ const generateParallelE2ETasks = (bv) => {
...
 const main = () => {
   const buildVariant = process.env.BUILD_VARIANT;
-  if (buildVariant) {
-    const fileDestPath = join(process.cwd(), "/.evergreen", "generate-parallel-e2e-tasks.json");
+  // If the task is triggered by a upstream task, we should always generate the tasks
+  const isDownstreamTask = process.env.REQUESTER === "trigger"
+
+  // Check if there are any changes in the given build variant directory
+  if (!isDownstreamTask && !hasChangesInDirectory(`${APPS_DIR}/${buildVariant}`) && !hasChangesInDirectory(PACKAGES_DIR)) {
+    console.log(`No changes detected in ${buildVariant} or packages directory, skipping e2e task generation`);
```
minnakt, line 151: Could we also check if it comes from a patch request? Sometimes I want to generate in patch request but it won't work due to there being no changes

example use case is if I want to module patch Evergreen changes against UI main

minnakt (changes requested):

PR #853, opened 2025-06-09
DEVPROD-11421 Do not overwrite default sorts on tasks table and task duration table

khelif96:
DEVPROD-11421

### Description
Do not overwrite default sorts when sorts exist in the url

### DIFF
--- a/apps/spruce/src/pages/version/Tabs/TaskDuration/index.tsx
+++ b/apps/spruce/src/pages/version/Tabs/TaskDuration/index.tsx
@@ -34,14 +34,16 @@ const TaskDuration: React.FC<Props> = ({ taskCount, versionId }) => {
   const versionAnalytics = useVersionAnalytics(versionId);
   const queryVariables = useQueryVariables(search, versionId);
   const hasQueryVariables = Object.keys(parseQueryString(search)).length > 0;
-  const { limit, page } = queryVariables.taskFilterOptions;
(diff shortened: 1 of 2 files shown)


apps/spruce/src/pages/version/Tabs/TaskDuration/index.tsx
```diff
@@ -34,14 +34,16 @@ const TaskDuration: React.FC<Props> = ({ taskCount, versionId }) => {
   const versionAnalytics = useVersionAnalytics(versionId);
   const queryVariables = useQueryVariables(search, versionId);
   const hasQueryVariables = Object.keys(parseQueryString(search)).length > 0;
-  const { limit, page } = queryVariables.taskFilterOptions;
+  const { limit, page, sorts } = queryVariables.taskFilterOptions;
   const [hasInitialized, setHasInitialized] = useState(false);
 
   useEffect(() => {
```
minnakt, line 40: the only thing is that we currently have a behavior where if you switch to task duration tab, it will sort by duration descending because we assume that's what users want to see; I think this new condition prevents that

khelif96: Hmm good point. I updated this so that we try to mirror the sorts across the task duration tab and tasks tab and if there aren't valid sorts it will go back to the default flow. If we find that people don't like the new behavior it should be easy enough to just go back to the behavior you have outlined.

PR #859, opened 2025-06-10
DEVPROD-17172: Fix visual errors on Task History tab

minnakt:
DEVPROD-17172

### Description
Fixes the issue where the timeline shoots off into the page.

The problem is that the date separators were a "shadow element" of the tasks which makes it impossible (or at least very hard) to predict how many elements will fit within the given width. I refactored so that they are an entirely separate item in the list, which makes it easier to estimate the amount of items that will fit. 

### Screenshots
Fix overflow problem for large numbers:
Screenshot 2025-06-10 at 1 14 20 PM

Added sticky header shadow:
Screenshot 2025-06-10 at 3 07 43 PM


minnakt:
converting to draft to fix some problems

khelif96 (approved):
LGTM I wasn't able to reproduce the bug!

PR #869, opened 2025-06-17
DEVPROD-4051: Uninstall `antd`

minnakt:
DEVPROD-4051

### Description
Uninstalls `antd` (and some other libraries that became unnecessary)


sophstad (approved):
Byeee 🐜

PR #871, opened 2025-06-20
DEVPROD-18192: Don't show toast for annotation error

sophstad:
DEVPROD-NNNNN

### DIFF
--- a/apps/spruce/src/pages/Task.tsx
+++ b/apps/spruce/src/pages/Task.tsx
@@ -51,10 +51,15 @@ export const Task = () => {
     pollInterval: DEFAULT_POLL_INTERVAL,
     fetchPolicy: "network-only",
     errorPolicy: "all",
-    onError: (err) =>
-      dispatchToast.error(
-        `There was an error loading the task: ${err.message}`,
-      ),
+    onError: (err) => {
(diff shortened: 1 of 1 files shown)


minnakt (approved):

apps/spruce/src/pages/Task.tsx
```diff
@@ -51,10 +51,15 @@ export const Task = () => {
...
     fetchPolicy: "network-only",
     errorPolicy: "all",
-    onError: (err) =>
-      dispatchToast.error(
-        `There was an error loading the task: ${err.message}`,
-      ),
+    onError: (err) => {
+      // We shouldn't show errors about annotation permissions resulting from the task resolver, but we can't separate out the query because we need to identify if the user has permissions to hide the tab accordingly.
+      // Thus, if an error comes from the annotation resolver, don't show a toast.
+      if (!err?.graphQLErrors?.some((e) => e?.path?.includes("annotation"))) {
```
khelif96, line 57: I think it might make more sense to filter out annotation errors. Since graphql resolvers can have multiple errors this would prevent all of them from being shown if one of them is an annotation error.
```suggestion
      const hasNonAnnotationErrors = err?.graphQLErrors?.some(
        (e) => !e?.path?.includes("annotation")
      );
      if (hasNonAnnotationErrors) {```

khelif96 (approved):
Approved (To not block) mod my suggestion!

PR #872, opened 2025-06-20
DEVPROD-11539 Remove sortBy and sortDir query params from useTableSort

khelif96:
DEVPROD-11539

### Description
Removes singleQueryParam handling in the useTableSort hook and its references. Update tests to support new format


apps/spruce/src/pages/hosts/utils.ts
```diff
@@ -7,30 +7,37 @@ import {
...
   const [statuses] = useQueryParam(HostsQueryParams.Statuses, []);
 
+  // Parse the sorts query param into an array of sort objects
+  const sortsArray = parseSortString(sorts ?? "", {
+    sortByKey: "sortBy",
+    sortDirKey: "sortDir",
+    sortCategoryEnum: HostSortBy,
+  });
+
+  // Default to status sort if no sorts are provided. We only support one sort at a time.
```
minnakt, line 29: I wonder if this can be supported natively by the table. There is a field called `maxMultiSortColCount` that we use in other places, which could help simplify the logic here

khelif96: I found a different command `enableMultiSort` to disable multi sort. Since we are trying to be consistent about the sorts param I think we still need this handling just to handle the array returned by `parseSortString`. But sorry if i'm misunderstanding your comment and you intended something else

PR #879, opened 2025-06-23
DEVPROD-11739 Add handling if activated_by parent patch

khelif96:
DEVPROD-11739

### Description
@minnakt  pointed out that we are still not generating e2e tasks if the task is generated by a patch. Looking at the task document it looks like the `activated_by` field will have a special value called `parent_patch`  I think this can be used to determine if we should generate tasks for evergreen patches. There doesn't appear to be any other field that signifies if a downstream task was activated by an upstream one for patches. 
image

### Testing
Do it live? Since we can't use set module since this relies on a downstream task trigger instead of a module

### DIFF
--- a/.evergreen/scripts/generate-parallel-e2e-tasks.js
+++ b/.evergreen/scripts/generate-parallel-e2e-tasks.js
@@ -4,6 +4,7 @@ import { APPS_DIR, PACKAGES_DIR, PARALLEL_COUNT, Tasks } from "./constants.js"
 import { hasChangesInDirectory } from "./git-utils.js";
 
 const ALWAYS_GENERATE_TASKS_REQUESTERS = ["trigger", "patch", "commit"];
+const PARENT_PATCH_USER = "parent_patch";
 const EVERGREEN_DIR = join(process.cwd(), "/.evergreen");
 const TASKS_FILE = join(EVERGREEN_DIR, "generate-parallel-e2e-tasks.json");
 /**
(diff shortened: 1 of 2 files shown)


.evergreen/scripts/generate-parallel-e2e-tasks.js
```diff
@@ -150,7 +151,9 @@ const main = () => {
   // If the task is triggered by a upstream task, we should always generate the tasks
   const shouldAlwaysGenerateTasks = ALWAYS_GENERATE_TASKS_REQUESTERS.includes(requester);
   const isDownstreamTask = triggerId !== undefined && triggerId !== "";
-  const mustGenerateTasks = shouldAlwaysGenerateTasks || isDownstreamTask;
+  const activatedBy = process.env.ACTIVATED_BY;
```
minnakt, line 154: nit. move to line 150 where the other env variables are

.evergreen/scripts/generate-parallel-e2e-tasks.js
```diff
@@ -150,7 +151,9 @@ const main = () => {
   // If the task is triggered by a upstream task, we should always generate the tasks
   const shouldAlwaysGenerateTasks = ALWAYS_GENERATE_TASKS_REQUESTERS.includes(requester);
   const isDownstreamTask = triggerId !== undefined && triggerId !== "";
```
minnakt, line 153: nit. I don't think you need `TRIGGER_ID` anymore since it seems it's logically equivalent to `trigger` requester (a case you already covered)

minnakt (approved):

PR #881, opened 2025-06-24
DEVPROD-822: Include ANSI results in search

sophstad:
DEVPROD-822

### Description

Strip ANSI formatting during search so that matches with formatting can be identified.

### Screenshots

image

### Testing

Add unit test

### DIFF
--- a/apps/parsley/src/utils/searchLogs/index.ts
+++ b/apps/parsley/src/utils/searchLogs/index.ts
@@ -1,3 +1,4 @@
+import stripAnsi from "strip-ansi";
 import { ProcessedLogLines } from "types/logs";
 import {
   isLogRow,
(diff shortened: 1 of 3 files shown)


apps/parsley/src/utils/searchLogs/index.ts
```diff
@@ -59,7 +60,7 @@ const searchLogs = (options: searchOptions): number[] => {
       }
     } else if (isLogRow(processedLogLine)) {
       if (processedLogLine >= lowerBound) {
-        if (searchRegex.test(getLine(processedLogLine))) {
+        if (searchRegex.test(stripAnsi(getLine(processedLogLine)))) {
```
minnakt, line 63: i think you also have to add this on Line 57; for the case that all sections and all subsections are collapsed, it won't detect a match

Screenshot 2025-06-24 at 2 24 07 PM

minnakt (approved):
awesome sauce....

PR #886, opened 2025-06-25
DEVPROD-18478 Add ability to toggle never activated tasks 

khelif96:
DEVPROD-18478

### Description
This PR builds off of https://github.com/evergreen-ci/evergreen/pull/9149 and adds a toggle in the menu item to conditionally show never activated tasks on the ui.
This supersedes https://github.com/evergreen-ci/ui/pull/885 since I needed to make a bunch of fundemental changes. 
### Screenshots
Before:
image
After:
image

### Testing
Tested on staging with this [patch](https://spruce-staging.corp.mongodb.com/version/66db6da8d9ad170007246cb1/tasks?includeNeverActivatedTasks=false&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC) which has a never activated task.


sophstad (changes requested):
I'm curious if you considered putting this on a user's UI Settings page instead of within version actions. On the one hand, it seems like a sort of power-user feature that doesn't need to be surfaced on every build. On the other, it may be handy for users to toggle the setting easily (although if it's enabled, filters could easily be applied to remove those tasks from view).

khelif96:
I opted to remove the default values for `includeNeverActivatedTasks` so that it falls back to the default behavior if the field is not explicetly set in the url. I think this makes it so that the patch/mainline commit behavior defaults to its original behavior and if people want to see more tasks for a version they should explicelty set it. lmk if you think it makes more sense to set it back

apps/spruce/src/gql/queries/display-task.graphql
```diff
@@ -1,12 +0,0 @@
-query DisplayTask($taskId: String!, $execution: Int) {
```
sophstad, line 1: Is this just unused query cleanup?

sophstad (approved):
I'm sorry for the delay! LGTM mod snapshots

apps/spruce/src/gql/queries/display-task.graphql
khelif96: Yup!

PR #888, opened 2025-06-26
DEVPROD-18446: Update deprecation banner for Task History page

minnakt:
DEVPROD-18446

### Description
* Update the deprecation banner to include a date
* Remove task-level "See History" button because the history tab is right underneath
* Keep test-level "See History" button as a convenience for test failure filtering

### Screenshots
Screenshot 2025-06-26 at 12 54 26 PM


sophstad (approved):
Sorry for delay 🫶

PR #903, opened 2025-07-02
DEVPROD-13870: Add ANSI formatting to resmoke logs

sophstad:
DEVPROD-13870


khelif96 (approved):
Thanks for doing this! Can you please follow up with the stakeholders once this is deployed on the linked ticket and see if the change addresses their needs and is enough to unblock them. There are a bunch of watchers on the other ticket but not many watching this ticket which was split from the original.

PR #915, opened 2025-07-08
DEVPROD-17214 Remove deleted BackgroundCleanupDisabled feature flags

khelif96:
DEVPROD-17214

### Description
This was deleted in the backend pr but was not deleted here


minnakt (approved):
also need to remove it from AdminSettings

bynn:
I think it's because in this [PR](https://github.com/evergreen-ci/evergreen/pull/9166/files#diff-8c17273db15c21d24beed3604e377ebd7263573288cf14fa7539286f2ae536d6) the EvergreenTestResultsDisabled flag was deleted right before. 

Do you think we can include the UI fix for that flag as well in this PR?

PR #916, opened 2025-07-08
DEVPROD-18618: Add Multichip Input widget 

aldo-aguilar:
[DEVPROD-18618](https://jira.mongodb.org/browse/DEVPROD-18618)

### Description
This PR adds a component for generic / form chips. Based on the existing filter chips component 

### Screenshots
image

### Testing
Local testing and storybook component added

### DIFF
--- a/apps/spruce/src/components/FilterChips/FilterChips.test.tsx
+++ b/apps/spruce/src/components/FilterChips/FilterChips.test.tsx
@@ -85,7 +85,7 @@ describe("filterChips", () => {
     render(
       <FilterChips chips={chips} onClearAll={onClearAll} onRemove={vi.fn()} />,
     );
-    await user.click(screen.getByRole("button", { name: "Clear filters" }));
+    await user.click(screen.getByRole("button", { name: "Clear all" }));
     expect(onClearAll).toHaveBeenCalledTimes(1);
   });
 


apps/spruce/src/components/FilterChips/FilterChip.tsx
minnakt, line 1: Maybe we should just change the text of `"Clear filters"` to say `"Clear all"`

Screenshot 2025-07-09 at 12 00 01 PM

apps/spruce/src/pages/adminSettings/tabs/FieldTemplates.tsx
```diff
@@ -0,0 +1,100 @@
...
+import TextInput from "@leafygreen-ui/text-input";
+import { ArrayFieldTemplateProps } from "@rjsf/core";
+import { size } from "@evg-ui/lib/constants/tokens";
+import { PlusButton, Variant } from "components/Buttons";
+import { FilterChipType } from "../../../components/FilterChips/FilterChip";
+import FilterChips from "../../../components/FilterChips/index";
+
+export const ArrayFieldTemplate: React.FC<
+  Pick<ArrayFieldTemplateProps, "items">
+> = ({ items }) => {
```
minnakt, line 12: So this is confusing because `ArrayFieldTemplate` makes intuitive sense here, but I don't think it's going to work. This is because to sync the changes with the form state, you have to use the native functions exposed from the array. But the only function it exposes is `onAddClick`, which isn't sufficient for our needs.

Instead we might have to use a widget. The widget is more flexible because we can use its native function `onChange` to modify the form state more easily. You can try this solution and remember to update the form schema to use this widget.

Also you should modify it however you need, as this is just a sketch of what it might look like:

```suggestion
export const LeafyGreenChipInput: React.FC<SpruceWidgetProps> = ({
  disabled,
  label,
  onChange,
  options,
  readonly,
  value = [],
}) => {
  const [text, setText] = useState("");

  const { "data-cy": dataCy, elementWrapperCSS, placeholder } = options;
  const isDisabled = disabled || readonly;

  const chips = value.map((v: string) => ({
    key: v,
    value: v,
  }));

  const addChip = () => {
    onChange([...value, text]);
  };

  const removeChip = (chip: string) => {
    const newItems = value.filter((v: string) => v !== chip);
    onChange(newItems);
  };

  return (
    <ElementWrapper
      css={css`
        ${chipStyles}
        ${elementWrapperCSS};
      `}
    >
      <InputWrapper>
        <ChipInput
          data-cy={dataCy}
          disabled={isDisabled}
          label={label}
          onChange={(e) => setText(e.target.value)}
          placeholder={placeholder}
          value={text}
        />
        <Button
          disabled={isDisabled || text.trim().length === 0}
          leftGlyph={<Icon glyph="Plus" />}
          onClick={addChip}
          variant={ButtonVariant.Primary}
        />
      </InputWrapper>
      <FilterChips
        chips={chips}
        onClearAll={() => onChange([])}
        onRemove={(chip) => removeChip(chip.value)}
      />
    </ElementWrapper>
  );
};

const chipStyles = css`
  display: flex;
  flex-direction: column;
  gap: ${size.xs};
`;

const ChipInput = styled(TextInput)`
  flex-grow: 1;
`;

const InputWrapper = styled.div`
  min-width: 100%;
  display: flex;
  align-items: flex-end;
  gap: ${size.xs};
`;

```

apps/spruce/src/pages/adminSettings/tabs/FieldTemplates.tsx
```diff
@@ -0,0 +1,100 @@
...
+    onAdd(chipKey, chipValue);
+    setChipKey("");
+    setChipValue("");
+  };
+  return (
+    <StyledFilterChipInput>
+      <TextInput
+        className="text-input"
+        label="Disabled GraphQL Queries"
+        onChange={(e) => {
```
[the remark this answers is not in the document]
minnakt: Also, the content of the text input should be cleared when a user submits

apps/spruce/src/components/SpruceForm/SpruceForm.stories.tsx
```diff
@@ -202,3 +214,53 @@ const example3Def = {
...
+            width: "100%",
+          },
+          "ui:elementWrapperCSS": {
+            width: "100%",
+            maxWidth: "100%",
+          },
+        },
+      },
+    },
+  },
```
minnakt, line 262:
```suggestion
  schema: {
    type: "object" as const,
    properties: {
      testText: {
        title: "This is the only visible page element",
        type: "array" as const,
        uniqueItems: true,
        default: [],
        items: {
          type: "string" as const,
          properties: {
            value: {
              type: "string" as const,
            },
          },
        },
      },
    },
  },
  uiSchema: {
    testText: {
      "ui:widget": widgets.ChipInputWidget,
    },
  },
```

apps/spruce/src/components/SpruceForm/Widgets/ChipInput.tsx
```diff
@@ -0,0 +1,81 @@
...
+`;
+const InputWrapper = styled.div`
+  min-width: 100%;
+  display: flex;
+  align-items: flex-end;
+  gap: ${size.xs};
+`;
+const StyledButton = styled(PlusButton)`
+  margin-top: 25px;
+`;
```
minnakt, line 81: Do we need the additional styles?
```suggestion
```

minnakt (changes requested):

apps/spruce/src/components/FilterChips/FilterChip.tsx
```diff
@@ -9,20 +9,25 @@ const maxChipLength = 25;
 
 interface FilterChipType {
   key: string;
-  value: string;
   title: string;
+  value?: string;
```
minnakt, line 13: I think it should be the opposite; title should be optional, value should be required

PR #917, opened 2025-07-09
DEVPROD-19219: Rename codegen files

sophstad:
DEVPROD-19219

### Description

Rename our codegen file to one of the [supported filenames](https://the-guild.dev/graphql/config/docs/user/usage). This allows the ESLint plugin to use information specified there in order to have more context on the GraphQL setup.

This resolves some errors I was encountering in DEVPROD-19103; opening a separate PR for reviewability purposes!


khelif96:
I was wondering why this broke. I remember having really good linter support in the past

PR #918, opened 2025-07-09
DEVPROD-18958 use password input type for rdp password

khelif96:
DEVPROD-18958

### Screenshots
image


sophstad (commented):
Do we know if this value needs to be redacted from the backend/network requests too? Or is shoulder surfing the only risk?

khelif96:
> Do we know if this value needs to be redacted from the backend/network requests too? Or is shoulder surfing the only risk?

It already is redacted on all logs and traces via 
image
This pr is primarily for shoulder surfing

PR #919, opened 2025-07-09
DEVPROD-19103: Task review UI

sophstad:
DEVPROD-19103

### Description

Add client-level field (i.e., it does not exist in the database) for a user to mark a task as reviewed. You can mark a task either from the version tasks table, or on the task page itself with the button. Execution tasks behave independently, with the ability to mark all execution tasks as un/reviewed from the display task, which may also show an indeterminate state.

Because the cache is being leveraged, these changes should persist across the app. However, this PR does not implement local backend support. This means changes will not yet persist if you reload/close the page.

This UI exists behind a feature flag so it should not render on prod. It also would translate nicely to saving `reviewed` status on the backend; we could just introduce that field on the backend and replace the cache updates with some mutations. The bulk of the UI work could remain unchanged.

### Screenshots

image
image

### Testing

Testing with the Apollo cache is hard 😬 I've fixed existing tests for now and will introduce new tests for this functionality in a separate PR!


sophstad:
Closing in favor of #942. I ended up consolidating a lot of the caching works into hooks so may as well just have that reviewed once.

PR #921, opened 2025-07-10
DEVPROD-13867: Display other failing commands in task metadata

minnakt:
DEVPROD-13867

### Description
Show other failing commands.

### Screenshots

Screenshot 2025-07-10 at 2 10 29 PM


apps/spruce/src/pages/task/metadata/DetailsDescription/index.tsx
```diff
@@ -28,30 +29,72 @@ const DetailsDescription = ({
...
+              popoverZIndex={zIndex.tooltip}
+            />
+          </>
+        ) : (
+          fullText
+        )}
+      </MetadataItem>
+      {otherFailingCommands.length > 0 ? (
+        <MetadataItem data-cy="task-metadata-other-failing-commands">
+          <MetadataLabel>Other Failing Commands: </MetadataLabel>
```
sophstad, line 56: What do you think of using [`<details>` and `<summary>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/details) to make this collapsible? Based on our tests it seems like there may be a lot of failures that aren't helpful most of the time. Could also include a count in the total to give some context

Image

minnakt: good idea! I changed it

minnakt:
evergreen refresh

PR #922, opened 2025-07-10
DEVPROD-19277: Enable task history tab on patches

minnakt:
DEVPROD-19277

### Description
Enable history tab for patches using the base task ID. A user requested this because it used to be possible to go to the task history tab from patches.

Note that the task history page only ever showed mainline commit history, so the same is done here.


minnakt:
> Nice Refactor! Do you think we should have some sort of alternative workaround for the "This Task indicator?"
> 
> Image

good idea! I'll update it

PR #933, opened 2025-07-16
DEVPROD-19304: Add pagination buttons to bottom of patches page

minnakt:
DEVPROD-19304

### Description
Add pagination buttons to bottom of the patches page so that users don't have to scroll all of the way up to go to the next page.

### Screenshots

Screenshot 2025-07-16 at 11 42 45 AM

### DIFF
--- a/apps/spruce/src/components/PatchesPage/index.tsx
+++ b/apps/spruce/src/components/PatchesPage/index.tsx
@@ -1,12 +1,9 @@
 import styled from "@emotion/styled";
 import Checkbox from "@leafygreen-ui/checkbox";
 import Cookies from "js-cookie";
-import PageSizeSelector from "@evg-ui/lib/components/PageSizeSelector";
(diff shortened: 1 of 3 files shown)


apps/spruce/src/components/PatchesPage/index.tsx
```diff
@@ -97,33 +86,25 @@ export const PatchesPage: React.FC<Props> = ({
...
       <ListArea
         loading={loading}
         pageType={pageType}
         patches={patches?.patches || []}
       />
+      {!loading && (
+        <PaginationButtons
+          filteredPatchCount={patches?.filteredPatchCount}
+          pageType={pageType}
+        />
```
khelif96, line 102: On the version page we conditionally show the pagination buttons if the length of the page is > 10 should we do that here?

khelif96 (approved):

PR #936, opened 2025-07-16
DEVPROD-15453: Remove duplicate copies of static assets

minnakt:
DEVPROD-15453

### Description
Remove duplicate copies of static assets

I couldn't get the paths to work for production Storybook 😞 The problem is the application expects absolute paths but Storybook needs a relative path. So I just designated `woff` to be for the Storybook fonts but it is kind of wack

I tried all of the methods outlined in https://storybook.js.org/docs/configure/integration/images-and-assets


packages/lib/src/components/styles/fonts.ts
```diff
@@ -1,5 +1,11 @@
 import { css } from "@emotion/react";
 
+/**
+ * Definitions for font faces.
+ *
+ * WOFF2 files use an absolute path because they're used for the actual application.
+ * WOFF files use a relative path because they're only used for Storybook, which is less important.
```
sophstad, line 7: Maybe we should just create a separate stylesheet for storybook? It would probably be more proper to separate our concerns, since it can just go in the storybook package

minnakt:
I don't understand why but I can't get this to work, so I'm abandoning this ticket

PR #937, opened 2025-07-16
DEVPROD-17219 remove spot fleet options from distro UI

ablack12:
DEVPROD-17219
### Description
Remove fleet options from the the distro UI; we should only have on-demand for this.

### Screenshots
Screenshot is kind of lame since I removed something haha, but it looks normal, and i was able to save without issues.
image

### Testing
Removed from tests; I think this doesn't have to update the analytics table.

### Evergreen PR
https://github.com/evergreen-ci/evergreen/pull/9223


minnakt (approved):
lgtm mod fixing the unit test

PR #942, opened 2025-07-18
DEVPROD-19103: Add task review UI

sophstad:
DEVPROD-19103

### Description

Add client-level field (i.e., it does not exist in the database) to mark a task as reviewed. You can mark a task either from the version tasks table, or on the task page itself with the button. Execution tasks behave independently, with the ability to mark all execution tasks as un/reviewed from the display task, which may also show an indeterminate state. Successful tasks have this feature disabled.

Because the cache is being leveraged, these changes should persist across the app. However, this PR does not implement local backend support. This means changes will not yet persist if you reload/close the page.

This UI exists behind a feature flag so it should not render on prod. It also would translate nicely to saving reviewed status on the backend; we could just introduce that field on the backend and replace the cache updates with some mutations. The bulk of the UI work could remain unchanged.

### Screenshots

image
image

### Testing

This PR includes a few storysnaps that are now failing to render. I'm working on fixing snapshots that utilize the Apollo cache, so for now checking in the failed snapshots prevents the test from failing. I'll include a fix in a followup PR!

Add UI tests including checking cache

### DIFF
--- a/apps/spruce/src/components/TaskReview/useTaskReview.ts
+++ b/apps/spruce/src/components/TaskReview/useTaskReview.ts
@@ -50,20 +50,7 @@ export const useTaskReview = ({
     cache.updateFragment(
       {
         id: cacheTaskId,
-        fragment: gql`
-          fragment TaskReviewed on Task {
-            id
-            displayStatus
-            execution
-            reviewed
-            executionTasksFull {
-              id
-              displayStatus
-              execution
-              reviewed
-            }
-          }
-        `,
+        fragment: REVIEWED_TASK_FRAGMENT,
       },
       (existing) => ({
         ...existing,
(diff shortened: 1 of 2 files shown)


apps/spruce/src/gql/client/schema.graphql
```diff
@@ -0,0 +1,6 @@
+# eslint-disable @graphql-eslint/executable-definitions
+
+# For defining local-only fields
+extend type Task {
```
khelif96, line 4: TIL

apps/spruce/src/components/TaskReview/useTaskReview.ts
```diff
@@ -0,0 +1,109 @@
...
+  );
+
+  // We have to break out display tasks and non-display mostly for testing purposes :(
+  // In prod, non-nullable executionTasksFull handles this better.
+  const updateDisplayTask = useCallback(() => {
+    cache.updateFragment(
+      {
+        id: cacheTaskId,
+        fragment: gql`
+          fragment TaskReviewed on Task {
```
khelif96, line 66: Whats the difference between this fragment and the one above?

khelif96 (approved):
Great job on this!

PR #956, opened 2025-07-23
DEVPROD-19923 Rename githubPatchAliases to githubPRPatchAliases

ZackarySantana:
DEVPROD-19923

### Description

This renames `githubPatchAliases` and adds `githubMQPatchAliases` graphql schema changes.

### Screenshots

NA

### Testing

Updated tests. I deployed the evg pr first to staging, then project settings would fail because of the field change. I then deployed spruce to staging- and it would load & save correctly.

The PR is ready for review, I've ran `yarn` and `yarn codegen` in both apps.

### Evergreen PR

[PR](https://github.com/evergreen-ci/evergreen/pull/9249) (it's ready to merge)

PR #967, opened 2025-07-29
DEVPROD-19923 Rename githubPatchAliases to githubPRPatchAliases

ZackarySantana:
DEVPROD-19923

Unreverts #956


sophstad (approved):
I can merge this after tomorrow's Spruce deploy so that we wouldn't have to revert Spruce if tomorrow's backend deploy is reverted.

PR #968, opened 2025-07-30
DEVPROD-19104: Task review backend

sophstad:
DEVPROD-19104

### Description

- Adds a simple [IndexedDB](https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API) key-value store to indicate whether a task has been reviewed or not. This collection is modified as a side effect of the cache update.
- When reading a task's review field, if a value does not exist in the cache, check for it in the DB and add it to the cache.
- Added the lightweight dependency [idb](https://github.com/jakearchibald/idb) to work with IndexedDB
- Upgrade LG table to get new test utils

### Testing

- Unit tests stub out IndexedDB usage to test cache logic
- Cypress tests can use IndexedDB and effectively test persistence

### DIFF
--- a/packages/lib/src/components/Badge/TaskStatusBadge/__snapshots__/TaskStatusBadge_AllBadges.storyshot
+++ b/packages/lib/src/components/Badge/TaskStatusBadge/__snapshots__/TaskStatusBadge_AllBadges.storyshot
@@ -3,80 +3,80 @@
     class="emotion-0"
   >
     <div
-      class="emotion-1 leafygreen-ui-c62i3c"
+      class="emotion-1 leafygreen-ui-sti25g"
       data-cy="task-status-badge"
     >
       Inactive
     </div>
     <div
-      class="emotion-1 leafygreen-ui-c62i3c"
+      class="emotion-1 leafygreen-ui-sti25g"
       data-cy="task-status-badge"
     >
       Unstarted
     </div>
     <div
-      class="emotion-1 leafygreen-ui-c62i3c"
+      class="emotion-1 leafygreen-ui-sti25g"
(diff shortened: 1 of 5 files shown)


apps/spruce/src/components/TaskReview/caching/index.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    const taskId = readField<Task["id"]>("id") ?? "";
+    const execution = readField<Task["execution"]>("execution") ?? 0;
+    getItem([taskId, execution]).then((data) => {
+      if (data !== undefined) {
+        storage.var(data);
+      }
+    });
+  }
+  return storage.var();
```
khelif96, line 24: I noticed that this is a bit flaky between refreshes and page navigations. I see the value still exists in the db but does not persist consistently when you refresh the page. 

https://github.com/user-attachments/assets/961df87f-1174-4589-877f-f3b6ec589225

apps/spruce/src/components/TaskReview/db.ts
```diff
@@ -0,0 +1,52 @@
+import { IDBPDatabase, openDB, DBSchema } from "idb";
+
+const DB_NAME = "spruce-db";
+const STORE_NAME = "task_reviews";
+
+// CAUTION! Updating the store's schema requires updating the DB_VERSION and writing an IndexedDB migration.
```
khelif96, line 6: Can you please link a doc on how to do a migration (Doesn't need to be a custom one just some external documentation on how to do this if we need to in the future)

khelif96 (commented):
This is pretty cool I love learning about new tools like this!

khelif96 (approved):

PR #971, opened 2025-07-31
DEVPROD-20268 Avoid new line when manually copy and pasting a log range

khelif96:
DEVPROD-20268

### Description
The menuitem component seems to have added a hidden new line character to every log line. So when a user drags a range of log lines to copy to their clipboard the end beginning of every line has a new line character. Adding the user-select:none attribute prevents that from happening. 
On Firefox this bug also extended to the implementation of the `pre` tag. The `pre` tag is treated specially in firefox and Firefoxes selection engine adds new lines to the end of pre tags when they are selected. This is a super obscure bug. https://bugzilla.mozilla.org/show_bug.cgi?id=1528442 

### Testing
Tried it out on my clipboard


sophstad (commented):
This works on Safari, but not Firefox 😭 is there any way to add broader support?

Also fwiw, I don't think this is related to MenuItem—the issue the user submitted is what Minna had identified on the initial ticket. The "Copy raw" button is one workaround for the issue, but would be great to have this work too!

PR #978, opened 2025-08-05
DEVPROD-19611 Update date handling in form schemas to reset time to midnight for date fields

khelif96:
DEVPROD-19611 

### Description
The date inputs seem to default to midnight so we should have the disabled state be the same so that we don't prevent people from setting the same day as a disabled date option.

### Screenshots
Before:
image
image

After:
image

image


sophstad (commented):
Is there a reason not to make this update where `exemptionRange` is defined? (in `components/Spawn/utils/hostUptime.ts`)

sophstad (approved):

PR #1012, opened 2025-08-29
DEVPROD-21432: Move `useBetaFeatures` hook to lib

minnakt:
DEVPROD-21432

### Description
I want to use the beta features hook across both apps, so I have to move it to Lib. But it requires GraphQL so I had to set up GraphQL in Lib too.


minnakt:
evergreen refresh

khelif96 (approved):
Woah I didn't realize we would need to do this too, thank you so much!

minnakt:
evergreen refresh

PR #1046, opened 2025-09-18
DEVPROD-20523: Fix disableBefore bug

sophstad:
DEVPROD-20523

### Description

A datetime picker bug in handling `disableBefore` prevented users from selecting the current day when the intent was to disable _yesterday_. Applying `setToUTCMidnight` to the `disableBefore` prop fixes the issue

### Testing

Update periodic build integration test since this is where a user encountered the issue

### DIFF
--- a/apps/spruce/cypress/integration/projectSettings/not_defaulting_to_repo.ts
+++ b/apps/spruce/cypress/integration/projectSettings/not_defaulting_to_repo.ts
@@ -172,12 +172,8 @@ describe("Project Settings when not defaulting to repo", () => {
       });
 
       cy.get("input[id='year']").as("startOfDateInput");
-      cy.get("input[id='day']").as("endOfDateInput");
 
-      // LG Date Picker does not respond well to .clear()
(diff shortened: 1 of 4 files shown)


apps/spruce/cypress/integration/projectSettings/not_defaulting_to_repo.ts
```diff
@@ -158,6 +158,63 @@ describe("Project Settings when not defaulting to repo", () => {
...
+        day: "16",
+      });
+
+      cy.get("input[id='year']").as("startOfDateInput");
+      cy.get("input[id='day']").as("endOfDateInput");
+
+      // LG Date Picker does not respond well to .clear()
+      cy.get("@endOfDateInput").type(
+        "{backspace}{backspace}{backspace}{backspace}{backspace}",
+      );
```
minnakt, line 180: maybe we should make a Cypress function for this `clearDateInput`

minnakt (approved):

PR #1055, opened 2025-09-22
DEVPROD-21812: Remove apollo/server

sophstad:
DEVPROD-21812

### Description

`path-to-regexp@1` has security vulnerabilities. It is included because of `@apollo/server`, which we use so trivially that we may as well remove. Apollo client error handling in v4 is different enough that we would probably have to change this line anyway.

Note that `path-to-regexp@3` is still included by Parsley's `serve` dependency, but this version has no vulnerability.

### DIFF
--- a/apps/parsley/package.json
+++ b/apps/parsley/package.json
@@ -42,7 +42,6 @@
   },
   "dependencies": {
     "@apollo/client": "3.10.4",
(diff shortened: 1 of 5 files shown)


khelif96 (approved):
SGTM good catch

PR #1067, opened 2025-09-24
DEVPROD-20854: Disable animations for 404 story

minnakt:
DEVPROD-20854

### Description
This story keeps getting flagged by Chromatic due to it having animations. Disables animation for stories and rewrites SVG in a more modern way (I'm not sure what format it was in but it threw a bunch of errors in TypeScript).

### DIFF
--- a/apps/spruce/src/components/404/NotFoundSvg.tsx
+++ b/apps/spruce/src/components/404/NotFoundSvg.tsx
@@ -1,1269 +1,1246 @@
-import { useEffect, useRef } from "react";
-
-interface NotFoundSvgProps {
-  disableAnimations?: boolean;
-}
-
-const NotFoundSvg: React.FC<NotFoundSvgProps> = ({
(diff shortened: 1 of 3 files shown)


apps/spruce/src/components/404/NotFoundSvg.tsx
```diff
@@ -1,13 +1,1269 @@
...
+interface NotFoundSvgProps {
+  disableAnimations?: boolean;
+}
+
+const NotFoundSvg: React.FC<NotFoundSvgProps> = ({
+  disableAnimations = false,
+}) => {
+  const svgRef = useRef<SVGSVGElement>(null);
+
+  useEffect(() => {
```
sophstad, line 12: Apparently Chromatic is supposed to [automatically disable](https://www.chromatic.com/docs/animations/) animations; is the useEffect still necessary with the updated SVG element?

PR #1071, opened 2025-09-26
DEVPROD-20393: Make task review disable-able

sophstad:
DEVPROD-20393


apps/spruce/src/pages/preferences/preferencesTabs/uiSettingsTab/PreferenceToggles.tsx
```diff
@@ -60,45 +61,50 @@ export const PreferenceToggles: React.FC = () => {
...
+  };
+
   return loading ? (
     <ParagraphSkeleton />
   ) : (
     <>
       <PreferenceItem>
-        <Toggle
-          aria-label="Toggle new Evergreen UI"
+        <Checkbox
```
minnakt, line 78: To me, it is a little odd that the checkboxes have immediate effects. I think they should be accompanied by some kind of "Submit" button in this case.

If we want to have immediate effects without any buttons, then I think keeping the toggles makes sense. I think we could just add a description ourselves to the toggle in that case!

apps/spruce/src/pages/preferences/preferencesTabs/uiSettingsTab/PreferenceToggles.tsx
```diff
@@ -60,45 +61,50 @@ export const PreferenceToggles: React.FC = () => {
     window.location.reload();
   };
 
+  const handleToggleTaskReview = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const nextState = !e.target.checked;
+    sendEvent({
+      name: "Toggled task review",
+      value: nextState ? "Enabled" : "Disabled",
+    });
+    Cookies.set(DISABLE_TASK_REVIEW, nextState.toString());
```
minnakt, line 70: I think this function would benefit from launching a success toast

sophstad: Ended up needing to have a page refresh to get consistent behavior with the cookies so I think a toast wouldn't really work now

PR #1083, opened 2025-10-02
DEVPROD-22249: Add cypress test to check save button of admin settings page

aldo-aguilar:
[DEVPROD-22249](https://jira.mongodb.org/browse/DEVPROD-22249)

### Description
Cypress test that modifies params throughout the page, saves, verifies, modifies a single section, then verifies that it didn't change other sections 

### Testing
- local 
- pr check


apps/spruce/cypress/integration/adminSettings/save_function.ts
minnakt, line 1: I think this test is fine to keep, but I'm not as concerned about non-Parameter Store fields being erased because they'll show up in the event log. Could we specifically write a test to check that all of the Parameter Store fields are correctly preserved?

minnakt (changes requested):

minnakt (approved):
great job!

PR #1089, opened 2025-10-06
DEVPROD-22707: Add task timing links

sophstad:
DEVPROD-22707

### Description

- Add links for wait time, run time, and total time. Each query shows a heatmap, P99 runtime, and run count as separate graphs.
- Allow filtering on only successful tasks and only commits to support the same functionality as the legacy page
  - Persist this setting via local storage
- I diverged slightly from the designs and added a new metadata card since we need to support multiple link types and config. This should be easy to expand in the future too
- Add new `total_time_min` and `wait_time_min` calculated fields to the evergreen-agent dataset on Honeycomb. They are defined for [prod](https://ui.honeycomb.io/mongodb-4b/environments/production/datasets/evergreen-agent/schema?dc=wait_time_min) and [staging](https://ui.honeycomb.io/mongodb-4b/environments/staging/datasets/evergreen-agent/schema?dc=wait_time_min)

### Screenshots

image

### Testing

- Add unit tests ensuring the links are correct and that localStorage works correctly
- 25 snapshot updates

### Evergreen PR

https://github.com/evergreen-ci/evergreen/pull/9422

This PR added activated time to task spans. Since this field is new note that wait time and total time metrics are limited to the past few days.


khelif96 (changes requested):
> This PR added activated time to task spans. Since this field is new note that wait time and total time metrics are limited to the past few days.

 We should flag this when announcing deprecation. 

Overall this looks great! I just had some thoughts on the actual links since they differ from the research Will did. 

I also think it would be great if we can link this ticket to the task timing epic and close it out once this is released

khelif96 (approved):

PR #1163, opened 2025-11-03
DEVPROD-23596: Fix public key test

sophstad:
DEVPROD-23596

### Description

Fix test again by clearing input before typing


khelif96 (approved):
reasonable to me if the input was already populated

PR #1254, opened 2025-12-18
DEVPROD-25351 add release mode description

ablack12:
DEVPROD-25351 

### Description
Just adding a small description to the admin page; this came up during the 24 hour release process.

I can't figure out how to make the css allow the description to be full size and the labels to still be nested; lmk if you know how to make this better.

Also, I don't understand the codegen thing that was needed, but I just did things until I got through the pre-commit hooks 🤷‍♀️😭

### Screenshots
image


minnakt (approved):
I think that's how you wanted the description to be styled but let me know if it's wrong

PR #1273, opened 2026-01-06
DEVPROD-23988 Add warning banner for OAuth spawn hosts

ZackarySantana:
DEVPROD-23988

### Description

Users often get confused about OAuth with spawn hosts, since it's a new workflow and hard to communicate to the users that need it. This helps that workflow, hopefully the new warning helps guide users on how to use spawn hosts with oauth.

### Additional Context
In a follow up PR, I want to make that checkbox for OAuth enabled based off a feature flag (which we will soon enable). This will force OAuth on all spawnhosts. I think it should not even let the user disable it, because the spawn host will simply **not** work unless it's checked (which is something that happened that confused users)

### Screenshots

Checked:
Screenshot 2026-01-06 at 9 48 51 AM

Unchecked:
Screenshot 2026-01-06 at 9 48 40 AM

### Testing

I didn't modify the existing tests because I wasn't sure if any were needed for this banner, please let me know if they are! I'm happy to put some in!

### DIFF
--- a/apps/spruce/src/components/Spawn/spawnHostModal/getFormSchema.tsx
+++ b/apps/spruce/src/components/Spawn/spawnHostModal/getFormSchema.tsx
@@ -238,6 +238,9 @@ export const getFormSchema = ({
                         title:
                           "Use OAuth authentication to download the task data from Evergreen. This will soon be required, see DEVPROD-4160",
                       },
+                      warningBanner: {
+                        type: "null" as const,
+                      },
                     },
                   },
                 ],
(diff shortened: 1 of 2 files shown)


apps/spruce/src/components/Spawn/spawnHostModal/getFormSchema.tsx
```diff
@@ -432,6 +448,21 @@ export const getFormSchema = ({
...
           },
+          warningBanner: {
+            "ui:showLabel": false,
+            "ui:warnings": [
+              <>
+                Spawn hosts with OAuth require additional setup. After SSHing in
+                to your spawn host, please run the command{" "}
+                <InlineCode>evergreen host fetch</InlineCode>. For more details,
+                refer to the{" "}
+                <StyledRouterLink to="https://docs.devprod.prod.corp.mongodb.com/evergreen/Hosts/Spawn-Hosts#spawning-a-host-from-a-task">
```
minnakt: Can you build this link on top of the existing variable `hostsDocumentationUrl`?

minnakt (approved):

PR #1274, opened 2026-01-06
DEVPROD-23988 Spawn hosts oauth option should respect service flags

ZackarySantana:
DEVPROD-23988

### Description

This makes the spawn host modal respect the service flag for jwt (oauth) usage. It forces the checkbox checked when it is enabled

### Screenshots

With the feature flag enabled:
Screenshot 2026-01-08 at 2 34 14 PM

Demo video: 

https://github.com/user-attachments/assets/61329605-82e8-4607-9d87-5ffdce5cb33a

### Testing

I updated the mock and deployed to staging, demo videoo above ^!

### Evergreen PR

[#9653](https://github.com/evergreen-ci/evergreen/pull/9653)

### DIFF
--- a/apps/spruce/src/components/Spawn/spawnHostModal/getFormSchema.tsx
+++ b/apps/spruce/src/components/Spawn/spawnHostModal/getFormSchema.tsx
@@ -36,7 +36,7 @@ interface Props {
   };
   isMigration: boolean;
   isVirtualWorkstation: boolean;
-  jwtTokenForCLIDisabled?: boolean;
+  oAuthDisabled?: boolean;
   myPublicKeys: MyPublicKeysQuery["myPublicKeys"];
   noExpirationCheckboxTooltip: string;
   spawnTaskData?: SpawnTaskQuery["task"];
@@ -56,9 +56,9 @@ export const getFormSchema = ({
   hostUptimeWarnings,
   isMigration,
   isVirtualWorkstation,
-  jwtTokenForCLIDisabled = false,
   myPublicKeys,
   noExpirationCheckboxTooltip,
+  oAuthDisabled = false,
   spawnTaskData,
   timeZone,
   useProjectSetupScript = false,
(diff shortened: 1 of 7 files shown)


apps/spruce/src/gql/queries/spruce-config.graphql
```diff
@@ -25,6 +25,9 @@ query SpruceConfig {
         }
       }
     }
+    serviceFlags {
+      jwtTokenForCLIDisabled
+    }
```
ZackarySantana: This is the part where I need help. This isn't queryable on SpruceConfig. I'm not sure where in the tree this would be queryable (if any?). Cursor does not know either

minnakt: Yeah, `SpruceConfig` is just a subset of `AdminSettings`. Querying `AdminSettings` directly isn't an option because it requires admin permissions. Maybe it's appropriate to allow `SpruceConfig` to expose service flags (if these flags are considered non-sensitive), which would require changes [here](https://github.com/evergreen-ci/evergreen/blob/545ca6fef8f1a0d46c70993ee63c87013c42bcb3/graphql/schema/types/config.graphql#L171-L183).

apps/spruce/src/components/Spawn/spawnHostModal/getFormSchema.tsx
```diff
@@ -238,6 +240,7 @@ export const getFormSchema = ({
                         type: "boolean" as const,
                         title:
                           "Use OAuth authentication to download the task data from Evergreen. This will soon be required, see DEVPROD-4160",
+                        default: !jwtTokenForCLIDisabled,
```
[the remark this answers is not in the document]
ZackarySantana: Even bots get confused by our service flag names 😢 . It's a bit confusing, but I think the demo shows this is the correct behavior (maybe I'm getting confused instead though). When the feature flag is switched on (it is off in prod rn), then the checkbox will be forced. If the feature flag is off, it is not forced.

minnakt (changes requested):

minnakt (changes requested):

minnakt (approved):

PR #1324, opened 2026-01-29
DEVPROD-24713: Correctly render names with user field

sophstad:
DEVPROD-24713

### Description

- Remove queries to author/authorDisplayName fields in favor of `User`. This should allow us to leverage more caching and perhaps [dataloaders](https://gqlgen.com/reference/dataloaders/) at some point
- Add cache read function to `displayName` where we return the user ID if display name is unset. This effectively makes `displayName` always defined (though we can't really capture that in typing, so I've just used `!`)

### Testing

- Update hook tests


minnakt (approved):
tbh that's it that's all I had to say so just going to approve it

PR #1325, opened 2026-01-30
DEVPROD-24931 Add debug toggle on spawn host modal

bynn:
DEVPROD-24931

### Description
add toggle for spawn host

### Screenshots
image

### Testing
made sure the toggle was there only when the correct settings were enabled 
staging

### DIFF
--- a/apps/spruce/src/components/Spawn/spawnHostModal/transformer.ts
+++ b/apps/spruce/src/components/Spawn/spawnHostModal/transformer.ts
@@ -27,6 +27,7 @@ export const formToGql = ({
   const {
     expirationDetails,
     homeVolumeDetails,
+    isDebug,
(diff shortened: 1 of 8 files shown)


apps/spruce/src/components/Spawn/spawnHostModal/transformer.ts
```diff
@@ -70,6 +71,7 @@ export const formToGql = ({
 
   return {
     isVirtualWorkStation,
+    ...(isDebug ? { isDebug: true } : {}),
```
minnakt: Slight preference for 
```suggestion
    isDebug: isDebug ?? false,
```
as long as it still accomplishes what you're trying to do

minnakt (approved):

PR #1342, opened 2026-02-04
DEVPROD-27470: Fix Parsley AI login

sophstad:
DEVPROD-27470

### Description

LeafyGreen commit [`9ee19a2`](https://github.com/mongodb/leafygreen-ui/commit/9ee19a2b6988bdde4a8be5489fd36e6e25d60856) introduced behavior where a button that is `disabled` has any `href` props set to `undefined`.

The fact that we set `isPolling = true` to disable the button on click introduced a race condition where the `href` was erased before it was actually opened. To get around this, we can just open the authentication directly from the onClick function.

### Testing

Add unit tests, validated locally


khelif96 (approved):

khelif96:
I don't think the unit tests were committed

PR #1352, opened 2026-02-08
DEVPROD-27703: Fix e2e test due to aria label changes

lavigneer:
DEVPROD-XXXX

### Description

A small tweak to test logic to fix issues caused by the change to the aria-label format in leafygreen. See [here](https://github.com/mongodb/leafygreen-ui/blob/a048363a1397c54efac4eaf3139af80e85957053/packages/date-picker/src/DatePicker/DatePickerMenu/DatePickerMenuHeader/DatePickerMenuHeader.tsx#L113) for the code that adds a suffix now.

I noticed this issue in some PR downstream tasks from the evergreen repo

### Testing

CI passes


sophstad (approved):
Thank you!

PR #1376, opened 2026-02-19
DEVPROD-19077: Dynamic Service Flags admin UI

sophstad:
DEVPROD-19077 - ~tagging app for their UX opinion!~ Can't tag the app team for some reason 🤔 

### Description

- Move service flags to their own tab since they use a separate save function. This also means we don't use the admin settings context or save button.
- Use checkbox instead of Enabled/Disabled radio buttons. Since we show the JSON key name in order to be dynamic/use Go reflection, it's pretty confusing to have `field_disabled` with Enabled/Disabled options. Double negative vibes...
- Slightly clean up sidebar nav
- Use "Service Flags" instead of "Feature Flags" for consistency with backend implementation

### Screenshots

image

### Testing

Add Cypress test

### Evergreen PR

https://github.com/evergreen-ci/evergreen/pull/9830


malikchaya2 (approved):
I think this is a great improvement and makes it more intuitive, thank you!

ablack12 (approved):
LOVE this matching the DB, the double negative really was hard to reason about. Assuming this defaults to false as well ?

athammer (approved):

minnakt (approved):
if App team likes it, sgtm

PR #1389, opened 2026-02-25
DEVPROD-28603: Replace TaskStatusIcon with TaskBox across all pages

khelif96:
## Summary
- Replace the old `TaskStatusIcon` (LeafyGreen glyph icons) with `TaskBox` (colored square icons) across all pages for visual consistency with the waterfall and task history views
- Update the variant history page, GroupedTaskStatusBadge tooltip, VersionRestartModal checkboxes, and DependsOn metadata icons
- Delete the now-unused `TaskStatusIcon` component, its stories, and snapshots (691 lines removed)

## Testing
- Type checks pass (`pnpm check-types`)
- All relevant unit tests pass (HistoryTableIcon, Cell, GroupedTaskStatusBadge, VersionRestartModal)
- ESLint passes on all modified files
- Verify visually in Storybook that HistoryTableIcon stories render colored squares
## Screenshots
image
image


apps/spruce/src/pages/task/metadata/DependsOn.tsx
```diff
@@ -61,10 +61,10 @@ const StyledBadge = styled(Badge)`
...
 const metStatusToIcon = {
-  [MetStatus.Started]: <TaskStatusIcon status={TaskStatus.Started} />,
-  [MetStatus.Met]: <TaskStatusIcon status={TaskStatus.Succeeded} />,
-  [MetStatus.Unmet]: <TaskStatusIcon status={TaskStatus.Failed} />,
-  [MetStatus.Pending]: <TaskStatusIcon status={TaskStatus.Pending} />,
+  [MetStatus.Started]: <TaskBox status={TaskStatus.Started} />,
+  [MetStatus.Met]: <TaskBox status={TaskStatus.Succeeded} />,
+  [MetStatus.Unmet]: <TaskBox status={TaskStatus.Failed} />,
+  [MetStatus.Pending]: <TaskBox status={TaskStatus.Pending} />,
 };
```
minnakt: I feel like this looks a little odd, the icons were more helpful:

Image

apps/spruce/src/components/GroupedTaskStatusBadge/index.tsx
```diff
@@ -58,7 +58,7 @@ export const GroupedTaskStatusBadge: React.FC<GroupedTaskStatusBadgeProps> = ({
         {statusCounts &&
           Object.entries(statusCounts).map(([taskStatus, taskCount]) => (
             <Row key={taskStatus}>
-              <TaskStatusIcon size={16} status={taskStatus} />
+              <TaskBox status={taskStatus as TaskStatus} />
```
minnakt: I feel like this looks odd too: 

Image

apps/spruce/src/components/VersionRestartModal/TaskStatusCheckbox.tsx
```diff
@@ -27,8 +28,12 @@ const CheckboxComponent: React.FC<TaskStatusCheckboxProps> = ({
     data-cy="task-status-checkbox"
     label={
       <StateItemWrapper>
-        <TaskStatusIcon status={status} />
-        {baseStatus ? <TaskStatusIcon status={baseStatus} /> : <EmptyCell />}
+        <TaskBox status={status as TaskStatus} />
```
minnakt: I like it but the squares would benefit from being a tad smaller 12~14px

apps/spruce/src/components/TaskStatusIcon/index.tsx
minnakt: Can we also delete the unused icons from `lib`? e.g. `OldKnownFailure`, `SystemFailure`, etc

minnakt (changes requested):

sophstad (changes requested):

apps/spruce/src/components/GroupedTaskStatusBadge/index.tsx
khelif96: I updated the met statuses since they do technically reflect different statuses that aren't exactly task statuses.
But these directly correlate to task statuses and I wonder if we should be consistent here otherwise we need to keep the task status icons and the task status squares and have 2 representations for the same state

khelif96: Maybe not lets do that

minnakt (changes requested):
just some small fixes

minnakt (approved):
everything looked in order when I checked

sophstad (approved):

PR #1431, opened 2026-03-11
DEVPROD-29280: Fix save button being disabled after toggling items

athammer:
DEVPROD-29280

### Description
When you toggled a setting and saved, if you unchecked that toggle, you couldn't save as the save button would be disabled. This is because we were using the initialData as the source of truth for what changed and comparing against it.
Screenshot 2026-03-11 at 3 22 01 PM

Changing initialData like that is a bit of a code smell imo but I think to fix this case more naturally would required a larger refactor compared to this simple change.

### Before
https://github.com/user-attachments/assets/44e591ae-bea3-4a74-a4e3-68a599e2e1ca

### After
https://github.com/user-attachments/assets/3dff2175-7991-4c79-a424-9a2c6dfbd0e1

### Testing
* Tested on local host
* Added tests

### DIFF
--- a/apps/spruce/src/pages/adminSettings/AdminSaveButton.tsx
+++ b/apps/spruce/src/pages/adminSettings/AdminSaveButton.tsx
@@ -46,6 +46,8 @@ export const AdminSaveButton: React.FC<AdminSaveButtonProps> = ({
     onError: (err) => {
       dispatchToast.error(`Error saving settings: ${err.message}`);
     },
+    awaitRefetchQueries: true,
+    refetchQueries: ["AdminSettings"],
   });
 
   const handleSave = () => {


sophstad (commented):
Did you try setting [`refetchQueries`](https://www.apollographql.com/docs/react/api/react/useMutation#usemutation-parameters-options-refetchqueries) on the save mutation? I think that may be less smelly

sophstad (commented):
what about [awaitRefetchQueries](https://www.apollographql.com/docs/react/api/react/useMutation#usemutation-parameters-options-awaitrefetchqueries) 👀

sophstad (approved):
I'd never seen it before either 😁

PR #1442, opened 2026-03-16
DEVPROD-29692: Default to exact filters on waterfall

sophstad:
DEVPROD-29692

### Description

Barely ~5% of waterfall string filters use exact matches, and I think many more could—and should, since they are so much faster.
- Default to exact match filtering
- Apply anchors `^` and `$` to filters to use equality matching. I think "exact" implies this so was a bit confusing that we did not do so already. Anchors allow these queries to use the task history index which is more performant. Using the regex option will still allow any filters that could be applied before.
- If a user toggles between regex/exact filter, save this and use it next time.
- Update tooltip text
- Include filter text in analytics
- Delete `TupleWithRegexConditional` as it's unused.
- Make Requesters filter slightly smaller, as it's [used way less](https://ui.honeycomb.io/mongodb-4b/environments/production/datasets/spruce/result/DM6MGRMX8WD) than the others, and generally with only 1 requester applied

### Testing

- Test regex/exact functionality in Cypress
- Test that new `TupleSelect` props work correctly


athammer (approved):
just a question on storyshots but other than that looks good!

PR #1449, opened 2026-03-17
DEVPROD-25996: Log line discoverability UI

sophstad:
DEVPROD-25996

### Description

This PR improves the discoverability of log line selection and sharing in Parsley by replacing the chain link icons with menu icons throughout the UI.

**Changes:**
- Replaced chain link (Link) and ArrowWithCircle icons with Ellipsis menu icons on all log lines
- Removed the share line toggle functionality (clicking link icon to set/unset share line)
- Updated "Copy share link" to use `selectedLineRange` query parameter instead of `shareLine`
- The menu now supports copying links for both single lines and line ranges
- Copy button uses the last used copy format (Jira/raw) from localStorage, consistent with DetailsMenu

**Backwards Compatibility:**
- The `shareLine` query parameter is still supported for old links, but can no longer be set via the UI
- All existing line selection behavior (click, shift+click for ranges) is preserved

### Screenshots

image

### Testing
- All unit tests passing (626 tests)
- Updated Cypress e2e tests for new UI behavior
- Updated snapshot tests
- TypeScript type checking passes
- ESLint passes (excluding expected GraphQL schema warnings)


apps/parsley/src/components/LogRow/BaseRow/index.tsx
```diff
@@ -82,30 +88,27 @@ const BaseRow: React.FC<BaseRowProps> = ({
...
-      sendEvent({ name: "Created new share line" });
-    }
-  }, [lineIndex, lineNumber, shared, scrollToLine, sendEvent, setShareLine]);
-
-  // Double clicking a line should add or remove the line from bookmarks.
-  const handleDoubleClick = useCallback(() => {
-    if (bookmarks.includes(lineNumber)) {
-      const newBookmarks = bookmarks.filter((b) => b !== lineNumber);
-      setBookmarks(newBookmarks);
-      sendEvent({ name: "Deleted bookmark" });
```
minnakt, line 102: Are we sure we want to remove the double click mechanism? I'm a little worried [based on usage](https://ui.honeycomb.io/mongodb-4b/environments/production/datasets/parsley/result/g47WbP7nuim) that people will be surprised that it's gone.

(Compared to [creation of share lines](https://ui.honeycomb.io/mongodb-4b/environments/production/datasets/parsley/result/HaJM53K38jo) which is much less)

apps/parsley/src/components/LogRow/BaseRow/index.tsx
```diff
@@ -82,30 +88,27 @@ const BaseRow: React.FC<BaseRowProps> = ({
...
-    if (shared) {
-      setShareLine(undefined);
-      sendEvent({ name: "Deleted share line" });
-    } else {
-      setShareLine(lineNumber);
-      scrollToLine(lineIndex);
-      sendEvent({ name: "Created new share line" });
-    }
-  }, [lineIndex, lineNumber, shared, scrollToLine, sendEvent, setShareLine]);
-
```
minnakt, line 96: It seems like we don't record any events for 
* `"Created new share line"`
* `"Deleted share line"`

now; can they be deleted or do they need to be added somewhere?

apps/parsley/src/components/LogRow/BaseRow/SharingMenu/index.tsx
```diff
@@ -159,9 +192,15 @@ const SharingMenu: React.FC = () => {
           glyph={<Icon glyph="Export" />}
           onClick={handleShareLinkToSelectedLines}
         >
-          Copy share link to selected {pluralize("line", lineCount)}
+          Copy link to {pluralize("line", lineCount)}
```
minnakt, line 195: I think this is missing a core functionality of `shareLine` where it jumps to the given line on load; users expect to see the shared line target at the top of their window

minnakt (changes requested):

apps/parsley/src/components/LogRow/BaseRow/index.tsx
sophstad: Good catch — I re-wired these events in the `MultiLineSelectContext` so they fire when a line is selected (`Created new share line`) and when the selection is cleared (`Deleted share line`).

apps/parsley/src/components/LogRow/BaseRow/SharingMenu/index.tsx
sophstad: Good point — added this in `LogPane` so the starting line of `selectedLineRange` is used for the initial scroll (taking precedence over `shareLine` / `failingLine`, matching the old jump-to behavior).

apps/parsley/src/components/LogRow/BaseRow/index.tsx
sophstad: Restored the double-click-to-bookmark behavior alongside the new menu option so existing muscle memory keeps working.

sophstad:
Replaced by #1658

PR #1459, opened 2026-03-19
DEVPROD-25194: Remove delete -> add Github App Credential Flow and instead use replace flow

athammer:
DEVPROD-25194

### Description
Removing delete and instead forcing a replace of Github "App Credentials" for project and repo level settings. This is to prevent people from removing Github App Credentials

### Screenshots
https://github.com/user-attachments/assets/5a5ec03e-7849-456b-b0b0-1fd66d491c04

### Testing
* Smoke tested
* Added tests

### DIFF
--- a/apps/spruce/src/pages/projectAndRepoSettings/shared/HeaderButtons.tsx
+++ b/apps/spruce/src/pages/projectAndRepoSettings/shared/HeaderButtons.tsx
@@ -130,17 +130,20 @@ export const HeaderButtons: React.FC<Props> = ({ id, projectType, tab }) => {
     });
   };
 
-  // Let users with without a Github App on their project page
-  // default to repo to use the repo Github settings for the project settings.
-  let hasExistingGithubApp = false;
+  // Prevent users from defaulting to repo if their project has a Github App
+  // but the repo does not, which would result in losing credentials entirely.
+  // If the repo has credentials, defaulting to repo is safe.
+  let disableDefaultToRepo = false;
   if (tab === ProjectSettingsTabRoutes.GithubAppSettings) {
     const appFormData = formData as AppSettingsFormState;
-    const githubAppId = appFormData?.appCredentials?.githubAppAuth?.appId ?? 0;
-    hasExistingGithubApp = githubAppId > 0;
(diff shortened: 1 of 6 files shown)


ablack12 (commented):
Thanks Aaron! I'm reviewing the behavior more than the code, but quick q about defaulting to repo -- does clicking "Default to Repo" at the page level still allow the app to be removed? (Ideally there'd be a separate way to default _just_ the app to repo, but i think this is so unlikely that even that would still be fine)

ablack12 (approved):
Looks great to me! I'm wondering if we can have defaulting to repo fail if no github app is set at the repo level and one is currently set at the project level. Just want to avoid being able to do this as much as possible. (Might need to be a backend change though? I can help advise there if so)

Approving this overall though, thanks for getting this out!

minnakt (changes requested):

apps/spruce/src/pages/projectAndRepoSettings/shared/tabs/GithubAppSettingsTab/AppSettingsTab.tsx
```diff
@@ -48,16 +49,19 @@ export const AppSettingsTab: React.FC<TabProps> = ({
         githubPermissionGroups,
         identifier,
         isAppDefined,
-        projectId,
+        isRepo,
+        projectId: projectId || repoId,
```
minnakt: not to be pedantic but can we call it `projectOrRepoId` 😁

apps/spruce/src/pages/projectAndRepoSettings/shared/HeaderButtons.tsx
```diff
@@ -129,7 +130,17 @@ export const HeaderButtons: React.FC<Props> = ({ id, projectType, tab }) => {
     });
   };
 
-  const canDefaultToRepo = !defaultToRepoDisabled.has(tab);
+  // Let users with without a Github App on their project page
```
minnakt: this seems fine for now but it does look out of place. Maybe we can think about how to restructure controlling this button in the future

minnakt (approved):

PR #1468, opened 2026-03-23
DEVPROD-29274: Remove migration script for cookies->localstorage preferences

athammer:
[DEVPROD-29274](https://jira.mongodb.org/browse/DEVPROD-29274)

### Description
Follow up from a PR where we moved preferences from cookies to localstorage. We had a migration script where we ported over these settings and now we're removing it after a week or so.

### Testing
* Tested in browser. Removing code so not much to test


minnakt (approved):
LGTM mod confirming the tests all pass

sophstad (commented):
nbd since it's still installed for Spruce but I think you can remove `js-cookie` and its types from Parsley's dependencies now

PR #1470, opened 2026-03-24
DEVPROD-32856: Upgrade TypeScript to v6

sophstad:
DEVPROD-32856

### Description

Upgrade TypeScript to v6. This is considered a [transition release](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html#preparing-for-typescript-70) for TypeScript 7, which is the Golang rewrite. It should be out in the next ~2 months and is _much_ faster, so this will allow us to upgrade immediately 🎉


khelif96 (approved):
🎉

PR #1480, opened 2026-03-25
DEVPROD-13241: Remove baseTaskStatuses

sophstad:
DEVPROD-13241

### Description

Removing this resolver will facilitate removing `DisplayStatus` calculations. This field is significantly slower than `Version.taskStatuses` because it filters out statuses that do not appear in the main version. We can just use `Version.baseVersion.taskStatuses` and apply that filtering on the client side.


minnakt (approved):
ok yay this actually didn't make sense to me at all in the first place

PR #1531, opened 2026-04-07
DEVPROD-29657: Remove "delete github app creds" query from the UI repo

athammer:
DEVPROD-29657

### Description
It's dead code after [this PR ](https://github.com/evergreen-ci/ui/pull/1459)and shouldn't be used and we should instead use the replace/set functionality to avoid people removing this.


minnakt (approved):
LGTM mod regenerating codegen types

PR #1565, opened 2026-04-16
DEVPROD-31460: Move Playwright config to `playwright-config` package

minnakt:
DEVPROD-31460

### Description
Moves Playwright config to a new package so it can be shared between apps.


.evergreen/shared.yml
```diff
@@ -189,7 +189,7 @@ functions:
 
         # Allow spec filtering for an intentional patch.
         if [[ "${playwright_spec}" != "" ]]; then
-          pnpm playwright:test "${playwright_spec}"
+          pnpm playwright:test ${playwright_spec}
```
khelif96, line 192: I think keeping the quotes is safer?

khelif96 (approved):

.evergreen/shared.yml
minnakt: it needs to remove quotes because of how the Playwright command will interpret quotes as regex

PR #1569, opened 2026-04-16
DEVPROD-31248: Improve usability of single task distro dropdown

minnakt:
DEVPROD-31248

### Description
Changes:
* Make general section open by default (requested by app engineers — collapsing the most used section is not helpful to them)
* Use a combobox and sort options alphabetically for single task distro dropdown in Admin Settings. 
* Added a notation for disabled projects. I didn't exclude them from the options because some of the ones in **production** are disabled projects, and I don't want people to get confused that they can't save the page. But we should clean up the list and exclude disabled projects at some point.
   * (Created [DEVPROD-31842](https://jira.mongodb.org/browse/DEVPROD-31842) for App to create an automation to remove these projects from single task distro section when disabled.) 

### Screenshots
Screenshot 2026-04-16 at 3 43 09 PM


apps/spruce/src/components/SpruceForm/Widgets/index.tsx
```diff
@@ -31,6 +32,7 @@ const widgets = {
   TextWidget: LeafyGreenTextInput,
   TextareaWidget: LeafyGreenTextArea,
   ToggleWidget: LeafyGreenToggle,
+  ComboboxWidget: LeafyGreenCombobox,
```
athammer: can we keep this list sorted alphabetically

athammer (approved):

PR #1572, opened 2026-04-17
DEVPROD-20921: Expose S3 Task Costs Field on Task Page

maddaak:
DEVPROD-20921

### Description
This PR is wire up and show the task cost on the left panel. Once the task is done we show the task cost.

### Screenshots
While Running:
image
image

After completion:
image

Task Endpoint: (this has the full precision value for cost)
image

### Testing
Tested on staging 

### Evergreen PR
[App PR](https://github.com/evergreen-ci/evergreen/pull/10131)

cc: @malikchaya2


khelif96 (changes requested):
I would prefer if you held off on adding these fields to the query until you were ready to use them .

maddaak:
> I would prefer if you held off on adding these fields to the query until you were ready to use them .



> I would prefer if you held off on adding these fields to the query until you were ready to use them .


This is the UI [Ticket](https://jira.mongodb.org/browse/DEVPROD-26469). 

Should i combine both into one PR?

apps/spruce/src/gql/queries/task.ts
```diff
@@ -134,6 +134,17 @@ export const TASK = gql`
         previousStepbackTaskId
       }
       tags
+      taskCost {
+        adjustedEBSStorageCost
```
malikchaya2, line 138: Can we add the new child patch cost here as well?

apps/spruce/src/pages/task/metadata/index.tsx
```diff
@@ -95,11 +95,29 @@ export const Metadata: React.FC<Props> = ({ error, loading, task }) => {
...
+    TaskStatus.SetupFailed,
+    TaskStatus.SystemFailed,
+    TaskStatus.SystemTimedOut,
+    TaskStatus.SystemUnresponsive,
+    TaskStatus.Aborted,
+    TaskStatus.KnownIssue,
+  ];
+  const costTooltip = completedStatuses.includes(displayStatus as TaskStatus)
+    ? "Final adjusted cost of this task."
+    : "Estimated cost based on execution so far. Updates as the task runs.";
```
malikchaya2: Can we instead wait and only show it once the task is done running?

malikchaya2 (approved):
Approved with two requests

apps/spruce/src/gql/queries/task.ts
maddaak: This will be added for the Version/patch page. Will address in that [PR](https://github.com/evergreen-ci/ui/pull/1603)

apps/spruce/src/pages/task/metadata/index.tsx
maddaak: Addressed!

apps/spruce/src/pages/task/metadata/index.tsx
```diff
@@ -327,6 +342,11 @@ export const Metadata: React.FC<Props> = ({ error, loading, task }) => {
         {testSelectionEnabledForProject && (
           <TestSelection testSelectionEnabled={testSelectionEnabled} />
         )}
+        {completedStatuses.includes(displayStatus as TaskStatus) && (
```
sophstad: You can just check that `finishTime` is defined as done below!

apps/spruce/src/pages/task/metadata/index.tsx
```diff
@@ -95,11 +95,26 @@ export const Metadata: React.FC<Props> = ({ error, loading, task }) => {
     spawnHostLink,
     startTime,
     tags,
+    taskCost,
     testSelectionEnabled,
     timeTaken,
     versionMetadata,
   } = task;
 
+  const totalCost = taskCost?.total ?? 0;
```
sophstad: I'd suggest only rendering the item below if `taskCost?.total` is defined, $0 could be confusing

sophstad (changes requested):

apps/spruce/src/pages/task/metadata/index.tsx
maddaak: Addressed!

apps/spruce/src/pages/task/metadata/index.tsx
maddaak: Addressed!

apps/spruce/src/pages/task/metadata/index.tsx
```diff
@@ -327,6 +328,11 @@ export const Metadata: React.FC<Props> = ({ error, loading, task }) => {
         {testSelectionEnabledForProject && (
           <TestSelection testSelectionEnabled={testSelectionEnabled} />
         )}
+        {finishTime && taskCost?.total != null && (
```
sophstad: [nit] This could also be `undefined`
```suggestion
        {finishTime && taskCost?.total && (
```

sophstad (approved):

apps/spruce/src/pages/task/metadata/index.tsx
maddaak: Addressed!

maddaak:
> I would prefer if you held off on adding these fields to the query until you were ready to use them .

Have made use of the values in the PR now. we show show the cost and the value in the task meta data panel. so PR is no longer just about exposing GraphQL values

apps/spruce/src/pages/task/metadata/index.tsx
sophstad: This seems unchanged still

sophstad (approved):
lgtm with requested change

apps/spruce/src/pages/task/metadata/index.tsx
maddaak: Sorry the file got missed from the commit. addressed!

PR #1573, opened 2026-04-17
DEVPROD-20922: Expose S3 Task Costs for GraphQL query

maddaak:
DEVPROD-20922

### Description
This PR is wiring up graphQL values to the UI that will be used to create a UI popup to show S3 Usage costs on the Versions page

The [Ticket](https://jira.mongodb.org/browse/DEVPROD-26470) where these values will be used

### Screenshots
image

### Testing
Tested this on playground and see the required task values

### Evergreen PR
The code is already in main

cc: @malikchaya2


khelif96 (changes requested):
It looks like you are just adding the fields to the query but we never actually use them.

maddaak:
> It looks like you are just adding the fields to the query but we never actually use them.



> It looks like you are just adding the fields to the query but we never actually use them.

We broke the PR into 2 parts. First adding the fields and then will implement the UI. The UI ticket --> The [Ticket](https://jira.mongodb.org/browse/DEVPROD-26470)

PR #1574, opened 2026-04-17
DEVPROD-31767 Add S3 storage cost admin fields for AWS account lists

malikchaya2:
DEVPROD-31767

### Description
This adds the fields added in https://github.com/evergreen-ci/evergreen/pull/10101 to the admin settings page. 

### Screenshots
image

### Testing
Added tests and tested against staging.


apps/spruce/src/pages/adminSettings/tabs/GeneralTab/OtherTab/schemaFields.ts
```diff
@@ -163,6 +163,20 @@ export const miscSettings = {
               title: "Default Max Artifact Expiration Days",
               minimum: 1,
             },
+            devprodOwnedAwsAccountIds: {
+              type: "array" as const,
+              title: "Devprod-owned AWS account IDs",
```
athammer: minor but seems like the other titles are title case, can we keep with that pattern?

apps/spruce/src/pages/adminSettings/tabs/GeneralTab/OtherTab/transformers.ts
```diff
@@ -273,11 +286,16 @@ export const formToGql = ((form: OtherFormState) => {
             miscSettings.cost.s3Cost.uploadCostDiscount ?? undefined,
         },
         storage: {
+          ...s3StoragePassthrough,
```
athammer: what's the reason for this addition? saw we added the new values below so wasn't sure if we were missing other values or something. if so can we add a TODO so we can add them explicitly at a later date?

athammer (approved):
added a few comments but nothing blocking

apps/spruce/src/pages/adminSettings/tabs/GeneralTab/OtherTab/transformers.ts
malikchaya2: It merges in the server’s storage (minus __typename) so partial “Other” saves don’t drop fields, and it avoids putting __typename on the mutation. But it's not strictly needed so I removed it to keep this PR more targeted.

PR #1585, opened 2026-04-20
DEVPROD-31460: Set up generate tasks for Playwright tests in Spruce

minnakt:
DEVPROD-31460

### Description
Migrate a couple of small tests and set up generate tasks to use for Playwright (not that useful right now since only a few tests are migrated). Sorry the PR is a little long but it's mostly set up.

I also really did not want to fix the `useAnnouncementToast` test so I just deleted it because we haven't used it for like 3 years.

### Follow up actions
I need to add `e2e_cypress` to Evergreen patch triggers when this is merged.


apps/spruce/playwright/helpers/index.ts
```diff
@@ -0,0 +1,135 @@
...
+
+/**
+ * Selects an option from a LeafyGreen select component
+ * @param page - The Playwright page object
+ * @param label - The label of the select input
+ * @param option - The option text or regex pattern to select
+ * @param options - Additional options for selecting the option
+ * @param options.exact - Whether to match the option text exactly (default: false)
+ */
+export async function selectLGOption(
```
sophstad: I was gonna suggest renaming this to `selectOption`, but maybe we can just wait it out and replace this with `selectOption` for the new Via design system 😆

apps/spruce/playwright/helpers/index.ts
```diff
@@ -0,0 +1,135 @@
...
+/**
+ * Validates the values in a date picker component
+ * @param page - The Playwright page object
+ * @param dataCy - The data-cy attribute value of the date picker
+ * @param opts - The expected date values
+ * @param opts.year - The expected year value
+ * @param opts.month - The expected month value
+ * @param opts.day - The expected day value
+ */
+export async function validateDatePickerDate(
```
sophstad, line 116: [nit] Could you colocate the datepicker utils in the file?

apps/spruce/playwright/fixtures.ts
```diff
@@ -0,0 +1,110 @@
...
+
+const bannerCookie = "This is an important notification";
+const hostMutations = ["ReprovisionToNew", "RestartJasper", "UpdateHostStatus"];
+
+type CustomFixtures = {
+  authenticatedPage: Page;
+};
+
+export const test = base.extend<CustomFixtures>({
+  authenticatedPage: async ({ page }, use) => {
```
sophstad, line 21: Basically echoing my comment [here](https://github.com/evergreen-ci/ui/pull/1588#pullrequestreview-4142798375); should we attempt using [`storageState`](https://playwright.dev/docs/auth#authenticate-with-api-request) for authentication?

packages/playwright-config/helpers/index.ts
```diff
@@ -12,6 +12,16 @@ export async function login(page: Page, user = users.admin): Promise<void> {
   });
 }
 
+/**
+ * Enters login credentials for the admin user
+ * @param page - The Playwright page object
+ */
+export async function enterLoginCredentials(page: Page): Promise<void> {
```
sophstad: Is this actually used anywhere?

apps/spruce/playwright/fixtures.ts
minnakt: I made [DEVPROD-31990](https://jira.mongodb.org/browse/DEVPROD-31990); I think it's ok to do as a follow up

apps/spruce/playwright/helpers/index.ts
minnakt: `selectOption` sgtm

packages/playwright-config/helpers/index.ts
minnakt: It's only used once in Cypress, so I will delete it and when migrating the specific test I will just write out the commands explicitly

minnakt:
changed some TSConfig path stuff let me know if you don't like it...

apps/spruce/playwright/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
+  "extends": "../../../tsconfig.base.json",
+  "compilerOptions": {
+    "paths": {
```
sophstad: TypeScript actually hates `paths` 😂 I think we should move away from them. There are some details in this PR... could try subpath imports if you want? idk :/ https://github.com/evergreen-ci/ui/pull/1343

minnakt: ok ty then nvm, I just reverted it

sophstad (approved):

PR #1588, opened 2026-04-20
DEVPROD-24935: Clean up Playwright tests in Parsley

minnakt:
DEVPROD-24935

### Description
Clean up Playwright tests in Parsley by using `page` (`authenticatedPage` is kinda long). This is just a find + replace PR

### DIFF
--- a/apps/parsley/playwright/fixtures.ts
+++ b/apps/parsley/playwright/fixtures.ts
@@ -3,10 +3,15 @@ import { execSync } from "child_process";
 import * as helpers from "./helpers";
 
 type CustomFixtures = {
+  unauthenticatedPage: Page;
   authenticatedPage: Page;
 };
 
 export const test = base.extend<CustomFixtures>({
+  unauthenticatedPage: async ({ page }, use) => {
+    await helpers.logout(page);
+    await use(page);
+  },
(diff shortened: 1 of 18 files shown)


sophstad (approved):
It seems like Parsley is _almost_ able to use a basic auth pattern like [this](https://playwright.dev/docs/auth#basic-shared-account-in-all-tests) (though it may be blocked on DEVPROD-30693). The ["Authenticate with API"](https://playwright.dev/docs/auth#authenticate-with-api-request) approach might also work for Spruce. Do you think we should consider these to avoid login on every file?

Changes look good though because either way just using `await page...` is better!

minnakt:
> It seems like Parsley is almost able to use a basic auth pattern like [this](https://playwright.dev/docs/auth#basic-shared-account-in-all-tests) (though it may be blocked on [DEVPROD-30693](https://jira.mongodb.org/browse/DEVPROD-30693)). The ["Authenticate with API"](https://playwright.dev/docs/auth#authenticate-with-api-request) approach might also work for Spruce. Do you think we should consider these to avoid login on every file?

I think it might be easier to implement this as a follow up, a separate ticket seems appropriate

PR #1589, opened 2026-04-20
DEVPROD-29657: Add task host overrides UI to distro provider settings

athammer:
DEVPROD-29657

### Description
Adding ability to add task host overrides in the distro setting page.

### Screenshots

https://github.com/user-attachments/assets/c77fdcb6-6036-40fc-960b-55d65f4c81e2

### Testing
* Smoke tested functionality

### DIFF
--- a/apps/spruce/src/components/SpruceForm/Widgets/LeafyGreenWidgets.tsx
+++ b/apps/spruce/src/components/SpruceForm/Widgets/LeafyGreenWidgets.tsx
@@ -197,7 +197,7 @@ export const LeafyGreenToggle: React.FC<SpruceWidgetProps> = ({
     elementWrapperCSS,
   } = options;
   return (
-    <ElementWrapper css={elementWrapperCSS}>
+    <ToggleElementWrapper css={elementWrapperCSS}>
       <ToggleWrapper>
         <Toggle
           aria-labelledby={`${id}-label`}
@@ -214,9 +214,18 @@ export const LeafyGreenToggle: React.FC<SpruceWidgetProps> = ({
       </ToggleWrapper>
       {descriptionNode ||
         (description && <Description>{description}</Description>)}
-    </ElementWrapper>
+    </ToggleElementWrapper>
   );
 };
+
+const ToggleElementWrapper = styled(ElementWrapper)`
+  &:last-child {
+    margin-bottom: 0;
+  }
+  > p {
+    margin-top: ${size.xxs};
+  }
+`;
 
 const ToggleWrapper = styled.div`
   align-items: center;
--- a/apps/spruce/src/components/SpruceForm/Widgets/LeafyGreenWidgets.tsx
+++ b/apps/spruce/src/components/SpruceForm/Widgets/LeafyGreenWidgets.tsx
@@ -197,7 +197,7 @@ export const LeafyGreenToggle: React.FC<SpruceWidgetProps> = ({
     elementWrapperCSS,
   } = options;
   return (
-    <ToggleElementWrapper css={elementWrapperCSS}>
+    <ElementWrapper css={elementWrapperCSS}>
       <ToggleWrapper>
         <Toggle
           aria-labelledby={`${id}-label`}
@@ -213,18 +213,13 @@ export const LeafyGreenToggle: React.FC<SpruceWidgetProps> = ({
         </Label>
       </ToggleWrapper>
       {descriptionNode ||
-        (description && <Description>{description}</Description>)}
-    </ToggleElementWrapper>
+        (description && <ToggleDescription>{description}</ToggleDescription>)}
+    </ElementWrapper>
   );
 };
 
-const ToggleElementWrapper = styled(ElementWrapper)`
-  &:last-child {
-    margin-bottom: 0;
-  }
-  > p {
-    margin-top: ${size.xxs};
-  }
+const ToggleDescription = styled(Description)`
+  margin-top: ${size.xxs};
 `;
 
 const ToggleWrapper = styled.div`
--- a/apps/spruce/src/pages/distroSettings/tabs/ProviderTab/getFormSchema.ts
+++ b/apps/spruce/src/pages/distroSettings/tabs/ProviderTab/getFormSchema.ts
@@ -7,6 +7,7 @@ import {
   ec2FleetProviderSettings,
   ec2OnDemandProviderSettings,
   ec2ProviderAccountField,
+  taskHostOverridesFields,
 } from "./schemaFields";
 import { textAreaCSS } from "./styles";
 
@@ -147,6 +148,11 @@ export const getFormSchema = ({
                   },
                 },
               },
+              taskHostOverrides: {
+                type: "object" as const,
+                title: "Task Host Overrides",
+                properties: taskHostOverridesFields.schema,
+              },
             },
           },
           {
@@ -178,6 +184,11 @@ export const getFormSchema = ({
                     ...ec2OnDemandProviderSettings.schema,
                   },
                 },
+              },
+              taskHostOverrides: {
+                type: "object" as const,
+                title: "Task Host Overrides",
+                properties: taskHostOverridesFields.schema,
               },
             },
           },
@@ -245,6 +256,13 @@ export const getFormSchema = ({
         },
         ...ec2OnDemandProviderSettings.uiSchema,
       },
+    },
+    taskHostOverrides: {
+      "ui:ObjectFieldTemplate": CardFieldTemplate,
+      "ui:data-cy": "task-host-overrides",
+      "ui:description":
+        "Override provider settings specifically for task hosts. Leave fields empty to use the default provider settings above.",
+      ...taskHostOverridesFields.uiSchema,
     },
   },
 });
--- a/apps/spruce/src/pages/distroSettings/tabs/ProviderTab/getFormSchema.ts
+++ b/apps/spruce/src/pages/distroSettings/tabs/ProviderTab/getFormSchema.ts
@@ -148,11 +148,7 @@ export const getFormSchema = ({
                   },
                 },
               },
-              taskHostOverrides: {
-                type: "object" as const,
-                title: "Task Host Overrides",
-                properties: taskHostOverridesFields.schema,
-              },
+              taskHostOverrides: taskHostOverridesFields.schema,
             },
           },
           {
@@ -185,11 +181,7 @@ export const getFormSchema = ({
                   },
                 },
               },
-              taskHostOverrides: {
-                type: "object" as const,
-                title: "Task Host Overrides",
-                properties: taskHostOverridesFields.schema,
-              },
+              taskHostOverrides: taskHostOverridesFields.schema,
             },
           },
         ],
@@ -261,7 +253,7 @@ export const getFormSchema = ({
       "ui:ObjectFieldTemplate": CardFieldTemplate,
       "ui:data-cy": "task-host-overrides",
       "ui:description":
-        "Override provider settings specifically for task hosts. Leave fields empty to use the default provider settings above.",
+        "When enabled, the values below replace the distro's provider settings for task hosts. Every field is required — empty values override the distro's settings rather than falling back to them.",
       ...taskHostOverridesFields.uiSchema,
     },
   },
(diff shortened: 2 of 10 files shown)


apps/spruce/src/pages/distroSettings/tabs/ProviderTab/schemaFields.ts
```diff
@@ -393,3 +399,103 @@ export const ec2ProviderAccountField = {
...
+                default: "",
+                minLength: 1,
+              },
+              iamInstanceProfileArn: {
+                type: "string" as const,
+                title: "IAM Instance Profile ARN",
+                default: "",
+                minLength: 1,
+                pattern: "^(a|ar|arn|arn:.*)?$",
+              },
```
minnakt: can we reuse preexisting items in this schema definition? For example, `iamInstanceProfileArn.schema` here?
```suggestion
              iamInstanceProfileArn: amInstanceProfileArn.schema,
```
Same for `securityGroups` and other fields

apps/spruce/src/pages/distroSettings/tabs/ProviderTab/schemaFields.ts
```diff
@@ -393,3 +399,103 @@ export const ec2ProviderAccountField = {
...
+  },
+  uiSchema: {
+    enableTaskHostOverrides: {
+      "ui:data-cy": "enable-task-host-overrides",
+      "ui:elementWrapperCSS": toggleSpacingCSS,
+    },
+    providerAccount: {
+      "ui:elementWrapperCSS": overrideFieldsLeadingSpaceCSS,
+    },
+    iamInstanceProfileArn: {
```
minnakt: same comment but with `instanceProfileARN.uiSchema`

apps/spruce/src/pages/distroSettings/tabs/ProviderTab/schemaFields.ts
```diff
@@ -393,3 +399,103 @@ export const ec2ProviderAccountField = {
...
+                default: false,
+              },
+            },
+          },
+        ],
+      },
+    },
+  },
+  uiSchema: {
+    enableTaskHostOverrides: {
```
minnakt, line 450: I wonder if using RadioBox would be more consistent with our other patterns on these types of forms; what do you think? (Is it ugly lol?)

apps/spruce/src/pages/distroSettings/tabs/ProviderTab/getFormSchema.ts
```diff
@@ -246,5 +249,12 @@ export const getFormSchema = ({
         ...ec2OnDemandProviderSettings.uiSchema,
       },
     },
+    taskHostOverrides: {
+      "ui:ObjectFieldTemplate": CardFieldTemplate,
+      "ui:data-cy": "task-host-overrides",
+      "ui:description":
+        "When enabled, the values below replace the distro's provider settings for task hosts. Empty values override the distro's settings rather than falling back to them. To remove the overrides, toggle off and save.",
+      ...taskHostOverridesFields.uiSchema,
```
minnakt: I think, based on our other patterns in project settings, that the description should go inside the card rather than outside

minnakt (changes requested):
it looks good I just have some style change requests

apps/spruce/src/pages/distroSettings/tabs/ProviderTab/schemaFields.ts
athammer: oh nice didn't know we had them. will use them. 

update: i spread them before the pattern so we can keep the patterns for these if that works

apps/spruce/src/pages/distroSettings/tabs/ProviderTab/schemaFields.ts
athammer: not to introduce a third thing but maybe a toggle would be better lol? this is an on or off setting so may be better ux
looking at it makes sense visually at least to me as well.
Screenshot 2026-05-11 at 5 17 29 PM

apps/spruce/src/pages/distroSettings/tabs/ProviderTab/getFormSchema.ts
athammer: done, i put the image above

apps/spruce/src/pages/distroSettings/tabs/ProviderTab/schemaFields.ts
```diff
@@ -393,3 +400,92 @@ export const ec2ProviderAccountField = {
...
+                type: "string" as const,
+                title: "Provider Account",
+                default: "",
+                minLength: 1,
+              },
+              iamInstanceProfileArn: {
+                ...instanceProfileARN.schema,
+                default: "",
+                minLength: 1,
+                pattern: "^(a|ar|arn|arn:.*)?$",
```
minnakt: I think you could probably add this pattern into `instanceProfileARN.schema` assuming it shares the same constraints

apps/spruce/src/pages/distroSettings/tabs/ProviderTab/schemaFields.ts
```diff
@@ -393,3 +400,92 @@ export const ec2ProviderAccountField = {
...
+                default: "",
+                minLength: 1,
+                pattern: "^(s|su|sub|subn|subne|subnet|subnet-.*)?$",
+              },
+              securityGroupIds: {
+                ...securityGroups.schema,
+                minItems: 1,
+                items: {
+                  ...securityGroups.schema.items,
+                  pattern: "^(s|sg|sg-.*)?$",
```
minnakt: same comment

apps/spruce/src/pages/distroSettings/tabs/ProviderTab/schemaFields.ts
```diff
@@ -393,3 +400,92 @@ export const ec2ProviderAccountField = {
...
+      "ui:elementWrapperCSS": toggleSpacingCSS,
+      "ui:widget": widgets.ToggleWidget,
+    },
+    providerAccount: {
+      "ui:elementWrapperCSS": overrideFieldsLeadingSpaceCSS,
+    },
+    iamInstanceProfileArn: {
+      ...instanceProfileARN.uiSchema,
+      "ui:placeholder":
+        "e.g. arn:aws:iam::123456789012:instance-profile/MyProfile",
```
minnakt: yeah I think you can add the placeholder to the `uiSchema` too

apps/spruce/src/pages/distroSettings/tabs/ProviderTab/schemaFields.ts
```diff
@@ -393,3 +400,92 @@ export const ec2ProviderAccountField = {
...
+    },
+    subnetId: {
+      "ui:placeholder": "e.g. subnet-xxxx",
+    },
+    securityGroupIds: {
+      ...securityGroups.uiSchema,
+      items: {
+        "ui:placeholder": "e.g. sg-xxxx",
+      },
```
minnakt: same comment

apps/spruce/src/pages/distroSettings/tabs/ProviderTab/schemaFields.ts
```diff
@@ -393,3 +400,92 @@ export const ec2ProviderAccountField = {
...
+          {
+            properties: {
+              enableTaskHostOverrides: {
+                enum: [true],
+              },
+              providerAccount: {
+                type: "string" as const,
+                title: "Provider Account",
+                default: "",
+                minLength: 1,
```
minnakt: can reuse `ec2ProviderAccountField`

apps/spruce/src/pages/distroSettings/tabs/ProviderTab/schemaFields.ts
```diff
@@ -393,3 +400,92 @@ export const ec2ProviderAccountField = {
...
+          },
+        ],
+      },
+    },
+  },
+  uiSchema: {
+    enableTaskHostOverrides: {
+      "ui:data-cy": "enable-task-host-overrides",
+      "ui:description":
+        "When enabled, the values below replace the distro's provider settings for task hosts. Empty values override the distro's settings rather than falling back to them. To remove the overrides, toggle off and save.",
```
minnakt, line 453: I have a question about this; if we are requiring things like `minLength: 1`, aren't the values not allowed to be empty?

apps/spruce/src/pages/distroSettings/tabs/ProviderTab/styles.ts
```diff
@@ -20,4 +20,23 @@ const indentCSS = css`
...
 
-export { textAreaCSS, mergeCheckboxCSS, indentCSS };
+const toggleSpacingCSS = css`
+  &:last-child {
+    margin-bottom: 0;
+  }
+  > p {
+    margin-top: ${size.xxs};
+  }
+`;
```
minnakt: rather than adjusting the label here, I'd prefer if we just updated the styling of the underlying toggle in `LeafyGreenWidgets.tsx` (the spacing change is good change anyways)

apps/spruce/src/pages/distroSettings/tabs/ProviderTab/styles.ts
```diff
@@ -20,4 +20,23 @@ const indentCSS = css`
...
+
+const overrideFieldsLeadingSpaceCSS = css`
+  margin-top: ${size.s};
+`;
```
minnakt: nit. I'd prefer to inline these in the actual schema since they aren't reused by other forms

minnakt (changes requested):

apps/spruce/src/pages/distroSettings/tabs/ProviderTab/schemaFields.ts
athammer: lemme see, was a bit worried about putting a blocker regex pattern as there's a decent amount of usage for some of these but it might be safe. i think regardless adding the placeholders should be good though.

apps/spruce/src/pages/distroSettings/tabs/ProviderTab/schemaFields.ts
athammer: oh good catch. added this as i was worried about people overwriting things unintentionally (empty values overwrite), but moved that responsibility back onto them through the disclaimer in the description. will remove these

apps/spruce/src/pages/distroSettings/tabs/ProviderTab/styles.ts
athammer: will do, was thinking this was a ui component library but seems more specialized for react json form so makes sense to move these to the form widget

apps/spruce/src/pages/distroSettings/tabs/ProviderTab/schemaFields.ts
```diff
@@ -393,3 +402,69 @@ export const ec2ProviderAccountField = {
...
+          {
+            properties: {
+              enableTaskHostOverrides: {
+                enum: [true],
+              },
+              providerAccount: ec2ProviderAccountField,
+              iamInstanceProfileArn: {
+                ...instanceProfileARN.schema,
+                default: "",
+              },
```
minnakt: I think it'll be safe to move the default value into the schema var as well
```suggestion
              iamInstanceProfileArn: instanceProfileARN.schema,
```

apps/spruce/src/components/SpruceForm/Widgets/LeafyGreenWidgets.tsx
```diff
@@ -214,10 +214,19 @@ export const LeafyGreenToggle: React.FC<SpruceWidgetProps> = ({
...
 
+const ToggleElementWrapper = styled(ElementWrapper)`
+  &:last-child {
+    margin-bottom: 0;
+  }
+  > p {
+    margin-top: ${size.xxs};
+  }
+`;
+
```
minnakt: Prefer keep using the base `ElementWrapper`.

I would just style the `Description` as

```suggestion
const ToggleDescription = styled(Description)`
  margin-top: ${size.xxs};
`;
```

Then, apply `elementWrapperCSS` as needed in the form.

minnakt (approved):
yippee! If you foresee chrome-overrides folder being a bother when committing, you can add it to the `.gitignore`

athammer:
just updated the pr with those suggestions! thank you for the reviews 🎉

PR #1596, opened 2026-04-22
DEVRPOD-31266 Support new log analyzer schema

khelif96:
DEVPROD-31266 

### Description
DEVPROD-31266 refactored the log analyzer and it now has a new schema. This PR ensures the UI now supports it and is usable. 

### Screenshots
image

### Sage PR
https://github.com/evergreen-ci/sage/pull/476


packages/fungi/src/MessageRenderer/ToolRenderer/utils.ts
```diff
@@ -5,35 +5,107 @@ export type ProgressUpdate = {
...
+  | "partial_failure"
+  | "unknown";
+export type FindingSeverity = "error" | "warning" | "info";
+
+export type MergedFindingError = {
+  line: number | null;
+  severity: FindingSeverity;
+  message: string;
+  evidence: string | null;
+};
```
athammer: is this only for errors? i see we can have normal info logs here.

packages/fungi/src/MessageRenderer/ToolRenderer/MergedFindingsView.tsx
```diff
@@ -0,0 +1,348 @@
...
+const ObservationsSection: React.FC<ObservationsSectionProps> = ({
+  observations,
+}) => {
+  if (observations.length === 0) return null;
+  return (
+    <Section>
+      <Subtitle>Observations</Subtitle>
+      <Bulleted>
+        {observations.map((o) => (
+          <li key={o}>
```
athammer, line 195: are these unique? if not we'll get a key collision. something like `key={${o}-${idx}}` would fix it though.

athammer (approved):
overall looks good! few cleanup things/questions but nothing blocking

packages/fungi/src/MessageRenderer/ToolRenderer/MergedFindingsView.tsx
khelif96: Yup they are! https://github.com/evergreen-ci/sage/blob/152eeac23705b9bdd3c0552accc4cd6666602aff/src/mastra/workflows/logAnalyzer/utils/index.ts#L195

PR #1604, opened 2026-04-24
DEVPROD-26470: S3 Costs Modal on Version Page to show breakdown

maddaak:
DEVPROD-26470

### Description
This PR add a costs detail button that then opens a modal to show the costs breakdown for a versin. We also are adding a honeycomb link that will take the user to honeycomb to see the values

### Screenshots
Patch Page Panel:
Created:
image

Running:
image

Completed:
image
Modal:
image

Version Panel:
Running:
image
 
Completed: 
image

### Testing
Testing in staging

### Evergreen PR
[App PR](https://github.com/evergreen-ci/evergreen/pull/10133)

CC: @malikchaya2


apps/spruce/src/pages/task/metadata/index.tsx
```diff
@@ -327,9 +350,45 @@ export const Metadata: React.FC<Props> = ({ error, loading, task }) => {
         {testSelectionEnabledForProject && (
           <TestSelection testSelectionEnabled={testSelectionEnabled} />
         )}
+        <MetadataItem
+          data-cy="task-metadata-cost"
+          tooltipDescription={costTooltip}
+        >
+          <MetadataLabel>Cost:</MetadataLabel> ${totalCost}
```
athammer: in this component it looks like we're treating totalCost as a number so 15.5 will render as $15.5 instead of $15.50. can we use `Intl.NumberFormat ` or another function to turn this into a money string for display

apps/spruce/src/pages/version/Metadata/index.tsx
```diff
@@ -52,11 +61,21 @@ export const Metadata: React.FC<MetadataProps> = ({ version }) => {
...
   } = version;
   const { sendEvent } = useVersionAnalytics(id);
+  const totalCost = cost?.total ?? 0;
+  const isVersionComplete = [
+    PatchStatus.Failed,
+    PatchStatus.Success,
+    PatchStatus.Aborted,
+  ].includes(status as PatchStatus);
+  const costTooltip = isVersionComplete
+    ? "Cost are for the cumulative cost of all tasks in this version."
```
athammer: bit of a nit, can we simplify the text here a bit this is a bit hard to read. "Cumulative cost of all tasks in this version"

apps/spruce/src/components/CostModal/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+            <th>Event</th>
+            <th>Cost</th>
+          </tr>
+        </thead>
+        <tbody>
+          {rows.map(({ cost, event }) => (
+            <tr key={event}>
+              <td>{event}</td>
+              <td>
+                {cost != null && cost > 0 ? `$${formatCost(cost)}` : "N/A"}
```
athammer: is zero not a valid cost here? just want to make sure we won't expect zero if so we'd show N/A which can be confusing

apps/spruce/src/components/CostModal/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+import { size } from "@evg-ui/lib/constants/tokens";
+import { DisplayModal } from "components/DisplayModal";
+
+export const formatCost = (cost: number): string => {
+  if (cost >= 0.01) return cost.toFixed(2);
+  return cost.toString();
+};
+
+const COST_DOC_URL =
+  "https://docs.devprod.prod.corp.mongodb.com/evergreen/FAQ/Cost-FAQ";
```
malikchaya2: same note as other PRs about making sure the link works before merging.

apps/spruce/src/components/CostModal/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+  cost: number | null | undefined;
+}
+
+interface CostModalProps {
+  /** Display name shown in the modal title, e.g. task display name */
+  name: string;
+  open: boolean;
+  onClose: () => void;
+  total?: number | null;
+  adjustedEC2Cost?: number | null;
```
malikchaya2: let's include child patch cost for patches

apps/spruce/src/components/CostModal/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+  adjustedS3LogPutCost,
+  adjustedS3LogStorageCost,
+  name,
+  onClose,
+  open,
+  total,
+}) => {
+  const rows: CostRow[] = [
+    { event: "Total", cost: total },
+    { event: "EC2 Instance", cost: adjustedEC2Cost },
```
malikchaya2:
```suggestion
    { event: "EC2", cost: adjustedEC2Cost },
```

apps/spruce/src/components/CostModal/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+  return (
+    <DisplayModal
+      data-cy="cost-modal"
+      open={open}
+      setOpen={onClose}
+      title={`Cost breakdown for ${name}`}
+    >
+      <StyledLink data-cy="cost-docs-link" href={COST_DOC_URL}>
+        View a breakdown of documented data tracked by Evergreen
```
malikchaya2:
```suggestion
        View a breakdown of cost tracked by Evergreen
```

apps/spruce/src/gql/queries/task.ts
```diff
@@ -134,6 +134,17 @@ export const TASK = gql`
         previousStepbackTaskId
       }
       tags
+      taskCost {
```
malikchaya2, line 137: is this needed here since this is for versions?

apps/spruce/src/gql/queries/version.ts
```diff
@@ -10,6 +10,16 @@ export const VERSION = gql`
       baseVersion {
         id
       }
+      cost {
+        adjustedEBSStorageCost
```
malikchaya2, line 14: let's also add this for patches

apps/spruce/src/pages/task/metadata/index.tsx
```diff
@@ -1,4 +1,6 @@
+import { useState } from "react";
```
malikchaya2, line 1: Are the changes in this file needed since this is for the version?

malikchaya2 (changes requested):

apps/spruce/src/pages/task/metadata/index.tsx
maddaak: We have a cost rounding function that we are using in evergreen-app for this the cost has a lot of decimal places. Handled this there! 
We also have a ticket in place for the trailing values --> https://jira.mongodb.org/browse/DEVPROD-32552

apps/spruce/src/pages/version/Metadata/index.tsx
maddaak: Reworded this!

apps/spruce/src/components/CostModal/index.tsx
maddaak: Addressed!

apps/spruce/src/components/CostModal/index.tsx
maddaak: Addressed

apps/spruce/src/components/CostModal/index.tsx
maddaak: Addressed

apps/spruce/src/components/CostModal/index.tsx
maddaak: Addressed!

apps/spruce/src/gql/queries/task.ts
maddaak: This branch was created on top of DEVPROD-26469 before it was merged into main. This should be addressed now!

apps/spruce/src/gql/queries/version.ts
maddaak: Addressed

The individual breakdown fields come from `version.cost`.  `patch.cost` only needs `childPatchesTotalCost` which is already fetched and shown as a separate line item in the modal.

apps/spruce/src/pages/task/metadata/index.tsx
maddaak: This branch was created on top of DEVPROD-26469 before it was merged into main. This should be addressed now!

apps/spruce/src/components/CostModal/index.tsx
maddaak: Individual fields are only non-zero when that cost category was actually incurred. So a zero value here means the category didn't apply (as in the that resource was not used). Using N/A for that

apps/spruce/src/pages/task/metadata/Metadata.test.tsx
```diff
@@ -83,6 +83,39 @@ describe("metadata", () => {
     await user.click(screen.getByDataCy("other-failing-commands-summary"));
     expect(screen.getByText("other failing command")).toBeVisible();
   });
+
+  it("HidesCostDetailsButtonWhenTaskIsRunning", () => {
```
maddaak, line 87: Missed a test case for this in the Task page. Adding it here.

malikchaya2 (approved):

athammer (approved):

PR #1616, opened 2026-04-28
DEVPROD-32605: Parallelize parsley e2e

sophstad:
DEVPROD-32605

### Description

Now that we're not applying any mutations via user settings...
- Parallelize Parsley e2e tests using Playwright's default workers config, which brings the Evergreen task duration down from ~8min to 4min 😍
- Move serial config/fixtures specifically into Spruce so new UIs (e.g. Sage) will default to parallel


apps/parsley/playwright/auth.setup.ts
```diff
@@ -0,0 +1,13 @@
+import { test as setup } from "@playwright/test";
+import { login } from "@evg-ui/playwright-config/helpers";
+
+const authFile = "bin/playwright/.auth/user.json";
+
+setup("authenticate", async ({ page }) => {
+  await login(page);
+  await page.goto("/");
+  await page.evaluate(() => {
+    localStorage.setItem("has-seen-searchbar-guide-cue-tab-complete", "true");
```
minnakt, line 10: If we can't import this, could we at least store it in a variable in this file?

minnakt (approved):
Lgtm

apps/parsley/playwright/auth.setup.ts
sophstad: `page.evaluate` runs this in a browser context so defining a variable wouldn't be so straightforward. I think it would look like this

```ts
await page.evaluate((key) => {
  localStorage.setItem(key, "true");
}, HAS_SEEN_SEARCHBAR_GUIDE_CUE);
```
but I think this is more confusing. Just gonna leave it if that's okay!

PR #1617, opened 2026-04-28
DEVPROD-31460: Migrate more tests to Playwright and improve checkbox interactions

minnakt:
DEVPROD-31460

### Description

Changes:
* I didn't like the checkbox interaction so I updated it.
* Cleaned up a bit in Parsley tests
* Migrated two more suites


packages/playwright-config/helpers/index.ts
```diff
@@ -55,21 +55,28 @@ export const validateToast = async (
   }
 };
 
+const clickHiddenElementByLabel = async (locator: Locator) => {
+  await expect(locator).not.toHaveAttribute("id", /undefined/);
```
sophstad, line 59: [nit] Is this necessary if we're using the id for the next operations?

sophstad (approved):
Cool! Any chance you'd be open to removing `eslint-plugin-cypress` in this PR? It unblocks the ESLint dependabot upgrade

waiting is no problem tho!

packages/playwright-config/helpers/index.ts
minnakt: yeah IDK why but the RJSF forms takes a moment before the IDs are loaded 😕

PR #1620, opened 2026-04-29
DEVPROD-31460: Migrate admin settings tests to Playwright

minnakt:
DEVPROD-31460

### Description
Migrates admin settings tests.


apps/spruce/playwright/helpers/index.ts
sophstad, line 1: I thought we enforce some of the syntax rules that were missed here, were they maybe missed on account of being in `playwright/` instead of `src/`? Perhaps the ESLint config could be updated

minnakt: I don't think we do, I just pasted this code verbatim (`async function` format) into a src file and it didn't autocorrect

sophstad (approved):

PR #1621, opened 2026-04-29
DEVPROD-31460: Migrate distro settings tests to Playwright

minnakt:
DEVPROD-31460

### Description
Migrates distro settings tests.


apps/spruce/playwright/tests/distroSettings/provider_section.spec.ts
```diff
@@ -0,0 +1,345 @@
...
+      const mergeUserDataCheckbox = page.getByRole("checkbox", {
+        name: "Merge with existing user data",
+      });
+      await clickCheckbox(mergeUserDataCheckbox);
+      await save(page);
+      await validateToast(page, "success", "Updated distro.", true);
+
+      await clickCheckbox(mergeUserDataCheckbox);
+      await save(page);
+      await validateToast(page, "success", "Updated distro.", true);
```
athammer: should this be duplicated by the block above? i see we're calling `clickCheckbox` three times which would leave this in checked state for the next test in this group unless there's some reset im missing

athammer (approved):
one question but other than that looks good

apps/spruce/playwright/tests/distroSettings/provider_section.spec.ts
minnakt: ty!

PR #1647, opened 2026-05-07
DEVPROD-32887: Use retryable assertions in Playwright tests

minnakt:
DEVPROD-32887

### Description
Things like `count()`, `innerText()`, `all()` are not retryable in Playwright, so we can replace these statements with `toHaveCount` and `toContainText`, or add additional guards. Also cleans up general test logic.


sophstad (approved):
This is basically flake prevention, right? Looks good!

PR #1635, opened 2026-05-06
DEVPROD-32861: Prevent Cloudfront caching for `commit.txt` 

minnakt:
DEVPROD-32861

### Description
`commit.txt` cannot be cached as it is used in the deploy steps.

PR #1648, opened 2026-05-08
DEVPROD-33032: Fix task tabs unnecessarily refreshing on return to tab

minnakt:
DEVPROD-33032

#### Issue
If you visit the task in the ticket, then switch to another tab, then come back, the entire page refreshes. This is not helpful because Apollo should at least continue showing the cached information rather than try to reload everything each time.

#### Solution
TBH I don't know why this happens. I suspected the change to `notifyOnNetworkStatusChange` in [this PR](https://github.com/evergreen-ci/ui/pull/1633), but even after I tried reverting, the issue persists. So it's not fully that. Possibly from when Apollo client was updated 3 months ago?

The `loading` state gets set to true every time we call `refetch` when we poll. So I just switched to using `networkStatus` which distinguishes between loading (first time) and loading (refetching).

### Testing
* Verified that going back and forth between tabs doesn't clear the trend charts, logs, etc


sophstad (commented):
Could this maybe be caused by #1526 alongside the new cache-control policy in #1635?

minnakt:
> Could this maybe be caused by #1526 alongside the new cache-control policy in #1635?

Reverting both does fix the issue on Logs/Tests but does not fix the issue for Trend Charts, which is probably the most important tab. I think blocking the task tab on data makes sense because that's what we do on [the version tab](https://github.com/evergreen-ci/ui/blob/654da9ad52f6cbdeb21240a7b20f86a03931d6cf/apps/spruce/src/pages/Version.tsx#L62-L64).

I changed the code to fix the entire page reloading & Trend Charts issue, but the Logs and Tests tab remain unfixed

sophstad (approved):

sophstad:
Thanks for looking into it! Do you think a followup ticket is necessary?

minnakt:
> Thanks for looking into it! Do you think a followup ticket is necessary?

I got some feedback that this is disruptive and we're gonna deploy it tomorrow, so I'm going to apply some ok-ish fixes but file a ticket to do a better fix 😅 [DEVPROD-33191](https://jira.mongodb.org/browse/DEVPROD-33191)

PR #1650, opened 2026-05-08
DEVPROD-28932: Show Honeycomb links for task history when base task isn't available

minnakt:
DEVPROD-28932

### Description
Since we can't show history for merge queue tasks, show a Honeycomb link instead.

### Screenshots
Screenshot 2026-05-08 at 5 26 00 PM


apps/spruce/src/pages/task/taskTabs/index.tsx
```diff
@@ -170,7 +177,23 @@ const useTabConfig = (
         name="History"
         {...walkthroughHistoryTabProps}
       >
-        <TaskHistory baseTaskId={baseTaskId} task={task} />
+        {requester !== Requester.GitHubMergeQueue ? (
+          <TaskHistory baseTaskId={baseTaskId} task={task} />
+        ) : (
+          <>
+            Evergreen cannot show history for GitHub merge queue tasks. See the{" "}
```
sophstad: Do you think it would be helpful to also link to the base task's history tab?

minnakt: You cooked me because it is better to block on `baseTaskId` 

also fixed the Honeycomb query for display tasks

minnakt: here's another one where history doesn't work `evergreen_ui_spruce_display_e2e_parallel_patch_3e16dc7dd416b7057b25a38d11eeb794059d8959_6a03a4f349d103000727256e_26_05_12_22_09_07` (for testing)

apps/spruce/src/pages/task/taskTabs/index.tsx
```diff
@@ -170,7 +176,26 @@ const useTabConfig = (
         name="History"
         {...walkthroughHistoryTabProps}
       >
-        <TaskHistory baseTaskId={baseTaskId} task={task} />
+        {baseTaskId ? (
+          <TaskHistory baseTaskId={baseTaskId} task={task} />
+        ) : (
+          <>
+            Evergreen cannot show history for this task because there is no
+            corresponding base task. Try viewing the{" "}
```
sophstad: "base task" isn't incredibly helpful for users but I can't think of a better option :/ one suggestion though is maybe if the task is in progress, we could indicate that task history may become available upon completion?

sophstad: Or further based on our convo, something like "Depending on the progression of the merge queue, task history may become available" if the task is in fact in a merge queue

minnakt: I don't really have a strong opinion, can you take a look at the copy and suggest additional changes if you have any?

sophstad (approved):
Your changes look great!

PR #1655, opened 2026-05-13
DEVPROD-33228: Upgrade pnpm to v11

minnakt:
[DEVPROD-33228](https://jira.mongodb.org/browse/DEVPROD-33228)

### Description
I think it's good to upgrade due to a number of security improvements made in pnpm 11, such as [`minimumReleaseAge`](https://pnpm.io/settings#minimumreleaseage) and [`blockExoticSubdeps`](https://pnpm.io/settings#blockexoticsubdeps).

Updated the build script to use `corepack`!

### DIFF
--- a/.evergreen/shared.yml
+++ b/.evergreen/shared.yml
@@ -119,18 +119,13 @@ functions:
         script: |
           ${PREPARE_SHELL}
 
-          # Install pnpm via standalone script.
-          # PNPM_VERSION should match packageManager field in package.json.
-          curl -fsSL https://get.pnpm.io/install.sh | env PNPM_VERSION=10.11.0 PNPM_HOME="$PNPM_HOME" SHELL="$(which bash)" bash -
+          # Download Node.js directly from official releases (with retries).
+          curl --output nodejs.tar.xz --retry 5 --retry-delay 10 "https://nodejs.org/dist/v${node_version}/node-v${node_version}-linux-x64.tar.xz"
+          tar xf nodejs.tar.xz
+          mv node-v* "$NODE_HOME" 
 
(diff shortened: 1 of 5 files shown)


.evergreen/shared.yml
```diff
@@ -119,18 +119,13 @@ functions:
         script: |
           ${PREPARE_SHELL}
 
-          # Install pnpm via standalone script.
-          # PNPM_VERSION should match packageManager field in package.json.
-          curl -fsSL https://get.pnpm.io/install.sh | env PNPM_VERSION=10.11.0 PNPM_HOME="$PNPM_HOME" SHELL="$(which bash)" bash -
+          # Download Node.js directly from official releases (with retries).
+          curl --output nodejs.tar.xz --retry 5 --retry-delay 10 "https://nodejs.org/dist/v${node_version}/node-v${node_version}-linux-x64.tar.xz"
```
athammer: can we add `--fail` so it fails on 4xx and 5xx and then subsequently retries? that should help reduce network flakes and also should make debugging & logging clearer

.evergreen/shared.yml
```diff
@@ -119,18 +119,13 @@ functions:
...
+          tar xf nodejs.tar.xz
+          mv node-v* "$NODE_HOME"
```
athammer: we should be able to reuse the node_version from above to prevent this from ever using the wrong node version in case there's two (which im assuming isn't possible?) but would be good in case there's ever an issue so we don't have to second guess ourselves.


```
mv "node-v${node_version}-linux-x64" "$NODE_HOME"
```

athammer (approved):
two small things but overall looks good. good idea upgrading!

PR #1658, opened 2026-05-14
DEVPROD-25996: Improve share menu UX

sophstad:
DEVPROD-25996

### Description

- Replace all link icons with ellipsis to open menu
- Preserve share link behavior, but only allow it to be set from the menu

### Screenshots

image

### Testing

- Update unit and playwright tests


apps/parsley/src/components/LogRow/BaseRow/index.tsx
```diff
@@ -82,18 +89,6 @@ const BaseRow: React.FC<BaseRowProps> = ({
...
   const shared = shareLine === lineNumber;
 
-  // Clicking link icon should set or unset the share line.
-  const handleClick = useCallback(() => {
-    if (shared) {
-      setShareLine(undefined);
-      sendEvent({ name: "Deleted share line" });
-    } else {
-      setShareLine(lineNumber);
-      scrollToLine(lineIndex);
```
athammer, line 92: doesn't seem like we use `scrollToLine` in this component anymore. can we remove it from the props?

athammer (approved):

athammer:
other than that prop cleanup looks good!

PR #1659, opened 2026-05-15
DEVPROD-33342: Add better messaging and UI around TSS action button

athammer:
DEVPROD-33342

### Description

Before the TSS action item would always show up and not give helpful information on why it would be there or disabled leading to confusing and not a great UX. Some of the copy is a bit length but also it shouldn't really show that much unless TSS is enabled.

**Behavior Table:**
  | Condition | Behavior | Message |
  |---|---|---|
  | Project's `testSelection.allowed` is false | Actions column hidden | — |
  | Project allows TSS, but task's `testSelectionEnabled` is false (non-patch requester, excluded build variant, or excluded task name) | Disabled dropdown item | "Test selection did not run for this task, so its tests cannot be quarantined. Test selection is only available on patch builds for build variants and tasks configured for it." |
  | Display task | Disabled dropdown item | "Quarantine status can only be managed from an execution task. Open an execution task from this display task to quarantine a test." |
  | Test is already quarantined | Enabled "Unquarantine" | "Allow this test to run in future task runs." |
  | Test is passing (and not quarantined) | Disabled "Quarantine" | "Passing tests cannot be quarantined." |
  | Test is failing (and not quarantined) | Enabled "Quarantine" | "Quarantine test so that it does not run in future task runs." |

### Screenshots
New message.
Screenshot 2026-05-15 at 4 04 54 PM

No longer showing action button.
Screenshot 2026-05-15 at 4 05 15 PM

### Testing
* Updated tests
* Smoke tested in staging

### DIFF
--- a/apps/spruce/src/pages/task/taskTabs/testsTable/ActionMenu/index.tsx
+++ b/apps/spruce/src/pages/task/taskTabs/testsTable/ActionMenu/index.tsx
@@ -85,8 +85,8 @@ export const ActionMenu: React.FC<ActionMenuProps> = ({ task, test }) => {
   } else if (task.displayOnly) {
     dropdownItems = [
       <DropdownItem key="display-task" disabled>
-        Quarantine status can only be managed from an execution task. Open an
-        execution task from this display task to quarantine a test.
+        Quarantine is not available on display tasks. Open one of this
+        task&apos;s execution tasks to quarantine a test.
       </DropdownItem>,
     ];
   } else if (test.isManuallyQuarantined) {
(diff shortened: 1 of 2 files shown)


apps/spruce/src/pages/task/taskTabs/testsTable/ActionMenu/index.tsx
```diff
@@ -77,14 +77,16 @@ export const ActionMenu: React.FC<ActionMenuProps> = ({ task, test }) => {
   if (!task.testSelectionEnabled) {
     dropdownItems = [
       <DropdownItem key="disabled" disabled>
-        Test selection is disabled for this task.
+        Test selection did not run for this task, so its tests cannot be
+        quarantined. Test selection is only available on patch builds for build
+        variants and tasks configured for it.
```
athammer, line 82: kinda lengthy but for now, most of these will be disabled due to this so being more descriptive is probably best

minnakt (approved):

PR #1663, opened 2026-05-19
DEVPROD-33402: Upgrade `vitest` to v4

minnakt:
DEVPROD-33402

### Description
Upgrades `vitest` to v4 in preparation for an upgrade for Vite. Also tried to consolidate the `setupTests` file in a single place.

Notable changes:
* `vi.restoreAllMocks` no longer resets the state of spies, so need to call other functions like `vi.resetAllMocks`
* mocks called with `new` keyword need to use new `function ()` keyword ([docs](https://vitest.dev/guide/migration.html#spyon-and-fn-support-constructors))


apps/spruce/src/gql/client/link/pausePollingLink.test.ts
```diff
@@ -1,7 +1,15 @@
 import { ApolloLink, execute, gql, ApolloClient } from "@apollo/client";
 import { Observable } from "@apollo/client/utilities";
 import { waitFor } from "@testing-library/react";
-import { describe, it, beforeEach, afterEach, vi, expect } from "vitest";
+import {
```
khelif96: Are these imports necessary in the first place?

khelif96 (approved):

PR #1665, opened 2026-05-19
DEVPROD-32552: Preserve trailing zeros on cost values on Task and Version Metadata Panel

maddaak:
DEVPROD-32552

### Description
This PR addresses a user reported issue where the Task/Patch/Version Cost value in the Metadata panel as well as the cost modal does not preserve trailing zero i.e if the cost is $0.10 the UI shows $0.1. 

Adding a custom formatter as for the cost we need higher precision than the `toFixedCost` function.
Ex: .1 --> .10
Ex: .000055 --> .000055

This function was added to apps/spruce/src/utils/numbers/index.ts and is for the cost value display in the task/version and modal places

### Screenshots

Task Page:
image
image
image

Patch/Version Page:
image
image
image

Patch Page with more Precision
image
image

### Testing
Tested in staging using this [Patch](https://spruce-local.corp.mongodb.com:8443/version/6a0c78bdb1b0c100077151f3/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)

### Evergreen PR
None

CC: @malikchaya2

### DIFF
--- a/apps/spruce/src/components/CostModal/CostModal.test.tsx
+++ b/apps/spruce/src/components/CostModal/CostModal.test.tsx
@@ -41,7 +41,7 @@ describe("CostModal", () => {
       />,
     );
     expect(screen.getByText("$0.52")).toBeInTheDocument();
-    expect(screen.getByText("$0.5")).toBeInTheDocument();
+    expect(screen.getByText("$0.50")).toBeInTheDocument();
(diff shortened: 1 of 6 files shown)


malikchaya2 (approved):

athammer (approved):
looks good. thank you for all the screen shots as well!

PR #1666, opened 2026-05-20
DEVPROD-31299: Upgrade Vite to v8

minnakt:
DEVPROD-31299

### Description
Upgrades Vite to v8:
* Get rid of `vite-tsconfig-paths` because there's now a `tsconfigPaths` option
* New automatic `codeSplitting` option is turned on by default


apps/parsley/vite.config.ts
```diff
@@ -92,13 +87,8 @@ const getProjectConfig = () => {
       }),
     ],
 
-    // Setting jsxImportSource to @emotion/react raises a warning in the console. This line silences
-    // the warning. (https://github.com/vitejs/vite/issues/8644)
-    esbuild: {
-      logOverride: { "this-is-undefined-in-esm": "silent" },
-    },
-
```
athammer, line 100: guessing this is just cleanup and we don't need this anymore?

athammer (approved):
looks good! if you haven't already might be good to throw on staging just to validate it as if it loads, it should be good to go.

apps/parsley/vite.config.ts
minnakt: yes there's no warning anymore I think

minnakt:
checked deploys on staging

PR #1667, opened 2026-05-20
 DEVPROD-33391: Wire up Sage UI deploy infrastructure

athammer:
DEVPROD-33391

### Description
Creating some of the configuration around infra for sage-ui in the evergreen ui repo.

### DIFF
--- a/packages/deploy-utils/src/utils/git/index.ts
+++ b/packages/deploy-utils/src/utils/git/index.ts
@@ -32,7 +32,7 @@ const getGitRoot = () => execTrim(`git rev-parse --show-toplevel`);
  * @param app - the app being deployed
  * @param fromCommit - the oldest commit in range to return
  * @param toCommit - optional hash marking the end of the range of commits. Defaults to HEAD.
- * @returns - a string of all commit messages between the first specified commit and last specified commit or HEAD. Commits are limited to those in the app's directory and all shared directories, as well as the root (the other app is excluded).
+ * @returns - a string of all commit messages between the first specified commit and last specified commit or HEAD. Commits are limited to those that touch the app's directory or the shared packages directory; commits scoped to other apps are excluded.
  */
 const getCommitMessages = (
   app: DeployableApp,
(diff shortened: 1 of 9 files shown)


packages/deploy-utils/src/utils/git/index.ts
```diff
@@ -42,13 +42,8 @@ const getCommitMessages = (
   const gitRoot = getGitRoot();
   const appDir = resolve(gitRoot, "apps", app);
   const packageDir = resolve(gitRoot, "packages");
-  const excludeDir = resolve(
-    gitRoot,
-    "apps",
-    app === "spruce" ? "parsley" : "spruce",
-  );
   const commitMessages = execTrim(
-    `git log ${fromCommit}..${toCommit} --oneline -- ${appDir} ${packageDir} '!${excludeDir}'`,
```
athammer, line 51: removed excludeDir. seemed like it wasn't working and appDir should already filter it so should be no behavior change 

can verify by running `diff <(git log --oneline -- apps/spruce) <(git log --oneline -- apps/spruce '!apps/parsley')`

packages/deploy-utils/src/utils/types.ts
```diff
@@ -10,12 +10,14 @@ export const isTargetEnvironment = (
...
+export type DeployableApp = (typeof DEPLOYABLE_APPS)[number];
 
 /**
  * isDeployable app asserts that a given string represents a deployable app
  * @param a - string to assert
  * @returns - typeguard for DeployableApp type
  */
 export const isDeployableApp = (a: string): a is DeployableApp =>
-  a === "parsley" || a === "spruce";
+  (DEPLOYABLE_APPS as readonly string[]).includes(a);
```
minnakt, line 23: possibly a silly question but could we just type it that way from the start?
```suggestion
export const DEPLOYABLE_APPS: readonly string[] = [
  "parsley",
  "sage-ui",
  "spruce",
];

export type DeployableApp = (typeof DEPLOYABLE_APPS)[number];

/**
* isDeployable app asserts that a given string represents a deployable app
* @param a - string to assert
* @returns - typeguard for DeployableApp type
*/
export const isDeployableApp = (a: string): a is DeployableApp =>
  DEPLOYABLE_APPS.includes(a);
```

minnakt (approved):

packages/deploy-utils/src/utils/types.ts
athammer: not at all! we'd lose the typings with `export const DEPLOYABLE_APPS: readonly string[]` as opposed to `as const` so we're kinda just fixing this by making the types looser. we could change `(DEPLOYABLE_APPS as readonly string[]).includes` to `(DEPLOYABLE_APPS as readonly string[]).some` which doesn't have the same type narrowing properties but because we use `DeployableApp` in a few other places it's nice to have that type being a bit more strict.

hopefully that answered your question! typescript is always confusing so not even sure what i said makes sense lol

PR #1673, opened 2026-05-21
DEVPROD-33401: Set up basic scaffolding for Sage UI

minnakt:
DEVPROD-33401

### Description
This PR includes basic set up for Sage UI. Difference from other apps:
* Using React 19
* Using Via components

I will add test setup in a different PR.

### Screenshots

Screenshot 2026-05-21 at 2 25 15 PM


khelif96 (commented):
Some questions I have are. 
Are the existing leafygreen components compatible with React 19? (For example if we want to reuse our table component) 
Is Via ready for production use?

apps/sage-ui/tsconfig.node.json
```diff
@@ -0,0 +1,20 @@
...
+    "lib": ["ESNext"],
+    "module": "ESNext",
+    "types": ["Node"],
+    "skipLibCheck": true,
+
+    /* Bundler mode */
+    "moduleResolution": "bundler",
+    "verbatimModuleSyntax": true,
+
+    /* Linting */
```
sophstad: [nit] I don't think JSON technically supports comments

apps/sage-ui/package.json
```diff
@@ -0,0 +1,37 @@
...
+    "eslint:staged": "STRICT=1 eslint --concurrency=auto",
+    "eslint:strict": "STRICT=1 eslint --concurrency=auto .",
+    "preview": "vite preview"
+  },
+  "dependencies": {
+    "@emotion/react": "11.14.0",
+    "@emotion/styled": "11.14.1",
+    "@evg-ui/lib": "workspace:*",
+    "@via-ds/components": "0.0.1-canary.0",
+    "@via-ds/icons": "0.0.2-canary.0",
```
sophstad, line 20: W.r.t. @khelif96's comment, I think that using Via here is a good idea. It could be a good opportunity to test it out and catch issues before we're working with a much larger surface area in Spruce. We may save ourselves some trouble down the line. I also think as an internal tool this is a perfect candidate for dogfooding (could be a good idea to reach out to Design Systems and let them know).

apps/sage-ui/public/favicon.svg
sophstad, line 1: Any opinions on renaming from `sage-ui` to `sage`? The ui part feels kinda redundant in this `ui` repo

apps/sage-ui/package.json
```diff
@@ -0,0 +1,37 @@
...
+  },
+  "devDependencies": {
+    "@evg-ui/eslint-config": "workspace:*",
+    "@evg-ui/lint-staged": "workspace:*",
+    "@types/node": "^24.12.3",
+    "@types/react": "^19.2.14",
+    "@types/react-dom": "^19.2.3",
+    "@vitejs/plugin-react": "^6.0.1",
+    "eslint": "^9.39.1",
+    "globals": "^17.6.0",
```
athammer: do we need globals here?

minnakt: It just came with the default Vite project but I can try uninstalling it

apps/sage-ui/tsconfig.node.json
minnakt: This came with the default Vite project install but I can delete it

apps/sage-ui/public/favicon.svg
minnakt: I think it's a little difficult to do it in this PR; I can do it in another one right after I merge it

minnakt:
> Some questions I have are. Are the existing leafygreen components compatible with React 19? (For example if we want to reuse our table component) Is Via ready for production use?

LeafyGreen can only support React 18, but I think it might be better for us to try starting a new project on React 19 with Via... a risk is potential conflicts with different versions of React installed

apps/sage-ui/vite.config.ts
```diff
@@ -0,0 +1,14 @@
+import react from "@vitejs/plugin-react";
+import { defineConfig } from "vite";
+
+const viteConfig = defineConfig({
+  server: {
+    port: 5493,
```
minnakt, line 6: No one noticed... I thought I was so clever... 🥹

apps/sage-ui/public/favicon.svg
sophstad: Sweet makes sense, figured it'd be good to do sooner than later

apps/sage-ui/vite.config.ts
sophstad: LOVEEEEEE

sophstad (approved):

athammer (approved):

apps/sage-ui/package.json
khelif96: Just followed up with lg and let them know they said we can share any feedback in #ask-via

PR #1669, opened 2026-05-21
DEVPROD-33572: Allow quarantining display task's execution tests

athammer:
DEVPROD-33572

### Description
Before we disallowed quaranting display task's tests. This will enable them to be quarantined.

### Screenshots
https://github.com/user-attachments/assets/d786aaed-0e0a-4808-adad-bc3ec261c1e1

https://github.com/user-attachments/assets/4cce0491-abd2-41f2-8406-a843dc9a2356

### Testing
* Added and updated UI tests 
* Updated storybook
* Smoke tested locally

### Evergreen PR:
https://github.com/evergreen-ci/evergreen/pull/10233

PR #1683, opened 2026-05-27
DEVPROD-33503: Allow quarantining passing tests and non-patch variants

athammer:
DEVPROD-33503

### Description
Work done:
  - **Quarantine passing tests**: drop the FE-side gate that required `test.status === Fail`. Users can now quarantine a flaky test from a run where it happened to pass for example
  - **Non-patch variants**:  removing "test selection is only available on patch builds" copy from the ActionMenu's disabled-state message. BE is also allowing non-patched variants to be quarantined as previously they were returning false. No other FE work needed

### Screenshots
https://github.com/user-attachments/assets/595858f1-ff1f-4daf-ba53-2c0de1755477

### Testing
* Smoke tested locally
* Added tests

### Evergreen PR
https://github.com/evergreen-ci/evergreen/pull/10240

### DIFF
--- a/apps/spruce/src/pages/task/taskTabs/testsTable/ActionMenu/ActionMenu.test.tsx
+++ b/apps/spruce/src/pages/task/taskTabs/testsTable/ActionMenu/ActionMenu.test.tsx
@@ -42,7 +42,7 @@ describe("action menu for tests table", () => {
     });
     expect(
       screen.getByText(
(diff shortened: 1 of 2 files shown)


khelif96 (approved):

athammer:
okay removed the code from https://github.com/evergreen-ci/ui/pull/1669 will double check nothing changed then merge

PR #1685, opened 2026-05-27
DEVPROD-33401: Set up Vitest and Playwright for Sage UI

minnakt:
DEVPROD-33401

### Description
Adds vitest and playwright

Sadly because `lib` is pinned to React 18, I can't reuse `test-utils` 😭😭


sophstad (commented):
> Sadly because lib is pinned to React 18, I can't reuse test-utils

I fear this is bad 😭 do you know if it applies to the entire lib package? If so, maybe we shouldn't go forward with React 19 as that could mean we enter Tech Debt City

minnakt:
> > Sadly because lib is pinned to React 18, I can't reuse test-utils
> 
> I fear this is bad 😭 do you know if it applies to the entire lib package? If so, maybe we shouldn't go forward with React 19 as that could mean we enter Tech Debt City

Yeah I'm pretty sure it's entirely not compatible due to the `lib` relying on React 18. We can downgrade to React 18 (shown in the current diff) or we can start a new package to start building out the components in Via and React 19 (shown in one of the commits)

athammer:
im thinking maybe we should just eat the cost here. if there was an opportunity to onboard to React 19 + via this is probably a better application to test run it on as the UI requirements are secondary and much smaller/incremental. also a bit worried if there's a push to upgrade to react 19 and Via, like if the old ui component stops receiving upgrades, we won't be caught in the middle or at the tail end of the project. it would be fun adding more storybooks as well lol

def want to know yall thoughts on this though

side-note: the pr itself looks good

sophstad:
That's fair -- but I'm also not sure whether making "new" components is going to be meaningful in any sense. It seems possible that upgrading to React 19 is just changing the version of the package used in `import x from "react"` with no other code changes required, in which case duplicating our code wouldn't be much of a win. But I think this is hard to predict exactly

minnakt:
Spoke with Aaron and I think we're leaning towards trying React 19 (especially since this is a entirely new package), so I updated the PR accordingly. It may end up not working out but I think it should be fairly easy to undo if we run into serious problems! 👀 

(I **did** figure out a way to get LeafyGreen components to render in React 19. But specifically for Vitest, it's not possible to reuse `lib/test_utils`. And I think we might rewrite the components as Via anyways during the course of the project if we need them)

athammer (approved):

.evergreen/scripts/generate-tasks.js
```diff
@@ -26,10 +26,11 @@ const targetsFromChangedFiles = (files) => {
       return;
     }
 
-    // If a change is made to a shared directory, test both apps.
+    // If a change is made to a shared directory, test all apps.
     if (packageDir !== APPS_DIR) {
       targets.add("spruce");
       targets.add("parsley");
+      targets.add("sage");
```
athammer: would changes to lib_new cause the other apps to deploy? is that a concern?

minnakt: it'll cause all tests for all apps to run currently. Let me see if I can fix it to only apply to sage

minnakt:
OK hopefully the generate tasks works correctly, kind of hard to test on this PR, but will double check after merging

PR #1689, opened 2026-05-29
DEVPROD-34043: Include user in Jira event creation log

athammer:
DEVPROD-34043

### Description
Adding user who created the Jira ticket to the event log.

### Screenshots
Screenshot 2026-05-29 at 1 57 57 PM

### Testing
* Story book tests

### DIFF
--- a/apps/spruce/src/pages/task/taskTabs/logs/logTypes/TaskEventLogLine.stories.tsx
+++ b/apps/spruce/src/pages/task/taskTabs/logs/logTypes/TaskEventLogLine.stories.tsx
@@ -69,14 +69,6 @@ export const TaskJiraAlertCreated = buildStory(
   },
 );
 
-export const TaskJiraAlertCreatedNoUser = buildStory(
-  TaskEventType.TaskJiraAlertCreated,
-  {
-    jiraIssue: "EVG-1234",
-    jiraLink: "https://jira.mongodb.org/browse/EVG-1234",
-  },
-);
-
 export const TaskDeactivated = buildStory(TaskEventType.TaskDeactivated, {
   userId: "mohamed.khelif",
 });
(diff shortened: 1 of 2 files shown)


apps/spruce/src/pages/task/taskTabs/logs/logTypes/TaskEventLogLine.stories.tsx
```diff
@@ -0,0 +1,112 @@
...
+  },
+);
+
+export const TaskJiraAlertCreatedNoUser = buildStory(
+  TaskEventType.TaskJiraAlertCreated,
+  {
+    jiraIssue: "EVG-1234",
+    jiraLink: "https://jira.mongodb.org/browse/EVG-1234",
+  },
+);
```
minnakt: this is a nit but I think this could also be a single story `TaskJiraAlertCreated` with an args toggle for `hasUser` false/true

apps/spruce/src/pages/task/taskTabs/logs/logTypes/TaskEventLogLine.tsx
```diff
@@ -77,6 +77,7 @@ export const TaskEventLogLine: React.FC<TaskEventLogEntry> = ({
           <StyledLink href={jiraLink ?? ""}>
             <strong>{jiraIssue}</strong>
           </StyledLink>
+          {userId && <> by {userId}</>}
```
minnakt: this can also just be a string I think

minnakt (approved):

apps/spruce/src/pages/task/taskTabs/logs/logTypes/TaskEventLogLine.stories.tsx
athammer: will remove. also for the toggle we can use userId string below and remove it to do this so should be good.

sorry for the huge image lol

Screenshot 2026-05-29 at 2 31 53 PM

apps/spruce/src/pages/task/taskTabs/logs/logTypes/TaskEventLogLine.tsx
athammer: good point fixed

PR #1692, opened 2026-06-01
DEVPROD-17132: Add variant display name to Variant History page

minnakt:
DEVPROD-17132

### Description
This is trying to add variant display name using the least-effort method possible. (I kinda assume the underlying query needs a partial rewrite eventually.)

### Screenshots
(Taking suggestions if anyone has them)

<kbd>
Screenshot 2026-06-01 at 4 13 00 PM</kbd>


apps/spruce/src/pages/variantHistory/index.tsx
```diff
@@ -76,6 +76,10 @@ const VariantHistoryContents: React.FC = () => {
       : skipToken,
   );
 
+  const variantDisplayName =
+    data?.mainlineCommits?.versions?.find((v) => v.version)?.version
+      ?.buildVariants?.[0]?.displayName ?? "";
```
athammer: only question i have is it worth checking the other indexes of `buildVariants`? 

only other thing is `?? ""` is technically not needed tmk as undefined and "" are both falsey but that's a nit.

athammer (approved):
had a question, but nothing blocking. looks good!

apps/spruce/src/pages/variantHistory/index.tsx
minnakt: there should just be 1 build variant at index 0 because of the filter! Will remove fallback

PR #1695, opened 2026-06-02
DEVPROD-33401: Set up Storybook for Sage UI

minnakt:
DEVPROD-33401

### Description
Sets up Storybook. 

I didn't create a Chromatic project yet since I don't think it will be a helpful signal until we have real components available.

### DIFF
--- a/apps/sage/.storybook/css.d.ts
+++ b/apps/sage/.storybook/css.d.ts
@@ -1,0 +1,1 @@
+declare module "*.css";
--- a/apps/sage/.storybook/css.d.ts
+++ b/apps/sage/.storybook/css.d.ts
@@ -1,1 +1,0 @@
-declare module "*.css";
--- a/apps/sage/package.json
+++ b/apps/sage/package.json
@@ -14,7 +14,9 @@
     "preview": "serve -s dist -p 5493",
     "playwright:test": "playwright test",
     "playwright:ui": "playwright test --ui",
-    "test": "vitest"
+    "test": "vitest",
+    "storybook": "storybook dev -p 6010",
+    "build-storybook": "storybook build"
   },
   "dependencies": {
     "@emotion/react": "11.14.0",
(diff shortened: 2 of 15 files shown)


apps/sage/package.json
```diff
@@ -14,7 +14,9 @@
     "preview": "serve -s dist -p 5493",
     "playwright:test": "playwright test",
     "playwright:ui": "playwright test --ui",
-    "test": "vitest"
+    "test": "vitest",
+    "storybook": "storybook dev -p 6010",
+    "build-storybook": "storybook build"
```
sophstad: [nit] Could we alphabetize these?

apps/sage/.storybook/css.d.ts
sophstad, line 1: Hmm we don't have this in Parsley or Spruce, any context on why it's necessary?

packages/lib_new/src/test_utils/types.ts
```diff
@@ -0,0 +1,45 @@
+import { ComponentType } from "react";
+import { StoryObj, Meta } from "@storybook/react-vite";
+import { DocumentNode, GraphQLError } from "graphql";
+
+/**
+ * Mock for ApolloProvider this allows you to mock GraphQL queries in tests.
+ */
+export type ApolloMock<Data, Variables> = {
```
sophstad: Do we need this still? Apollo [exports a `MockedResponse` type](https://www.apollographql.com/docs/react/development-testing/testing#dynamic-variables)

apps/sage/package.json
```diff
@@ -41,6 +47,7 @@
     "typescript": "~6.0.2",
     "typescript-eslint": "^8.59.2",
     "vite": "^8.0.12",
-    "vitest": "^4.1.6"
+    "vitest": "^4.1.6",
+    "storybook": "^10.1.10"
```
sophstad: Same, let's alphabetize

sophstad (approved):

apps/sage/.storybook/css.d.ts
minnakt: oops I thought it resolved a specific error but when I checked again it had no effect. So I will just delete it

PR #1698, opened 2026-06-03
DEVPROD-33404: Set up basic routes in Sage UI

minnakt:
DEVPROD-33404

### Description
This PR just sets up routing skeleton for Sage UI. (Will add error boundaries, observability in future PRs)

### Screenshots

Click on the placeholder links in the navbar to visit other pages:

<kbd>
Screenshot 2026-06-03 at 3 20 37 PM</kbd>


packages/lib/src/constants/tokens-via.ts
```diff
@@ -0,0 +1,66 @@
...
+  xxs: tokens.space[100].$value, // 4px
+  xs: tokens.space[200].$value, // 8px
+  s: tokens.space[400].$value, // 16px
+  m: tokens.space[600].$value, // 24px
+  l: tokens.space[800].$value, // 32px
+  xl: tokens.space[1600].$value, // 64px
+  xxl: tokens.space[1800].$value, // 72px
+} as const;
+
+export const palette = {
```
sophstad, line 13: Do you think it's necessary to re-export colors? we never did for LG

apps/sage/src/pages/AgentDetail/index.tsx
```diff
@@ -0,0 +1,6 @@
+import { useParams } from "react-router-dom";
+
+export const AgentDetailPage: React.FC = () => {
+  const { agentId } = useParams<{ agentId: string }>();
```
sophstad: I think if we use the [react router vite plugin](https://reactrouter.com/tutorials/quickstart#vite-config) we should be able to get [type safety ](https://reactrouter.com/how-to/route-module-type-safety)on params? Not sure if you're interested in trying this.

minnakt: I gave this a try before putting up this PR, and a second time after seeing your comment, but I think it might be genuinely not be great 😞 the framework mode requires a lot more setup. I couldn't figure out how to inject emotion and even Claude gave up on me lol

I think I prefer Tanstack's Router implementation if anything

packages/lib/src/constants/tokens-via.ts
minnakt: I thought it would be easier for an eventual migration since it looks like Via doesn't have a `palette` object! Everything is stored in a big object `tokens`. But also I can type them out if preferred (just leave a comment and I will change it). Just thought they were kinda wordy

apps/sage/src/pages/AgentDetail/index.tsx
sophstad: Tanstack is the new it girl 😅 even after her table caused us such pain...

Do you want to switch? I'd be interested, I've never liked react-router. Totally optional though!

packages/lib/src/constants/tokens-via.ts
sophstad: Sounds good to me, the `$value` bit is especially cumbersome

sophstad (approved):

PR #1705, opened 2026-06-09
DEVPROD-30922: Replace generator task with build variant path filtering 

minnakt:
DEVPROD-30922

### Description

Replaces `generate-tasks` with build variant path filtering. The `generate-spruce-e2e` will remain as-is.

Right before merging this PR, I will need to update the aliases on the Pull Request Testing project settings page. This setting will affect what gets scheduled in GitHub PRs (don't want to add it now because it will mess with others' GitHub PRs).

### Open Questions
Q: Do we want to run all tasks on mainline? Build variant path filtering will have unchanged variants as **unscheduled by default** on mainline. Here's an [example](https://spruce-staging.corp.mongodb.com/version/sandbox_ad1a44bdfd27d4eee33a9d37f892bae830a8286f/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&statuses=undispatched-umbrella,unscheduled,aborted,blocked&variant=%5Egit-env-test%24); this variant's path filter doesn't get triggered so it shows its tasks as unscheduled. However, we can add `activate: true` to the build variant to override that functionality and always activate all tasks on mainline.

A: For now we will mirror current functionality and activate all core tasks on mainline.


.evergreen/scripts/generate-tasks.js
```diff
@@ -1,79 +0,0 @@
...
-    if (IGNORED_FILE_EXTENSIONS.has(ext.toLowerCase())) {
-      return;
-    }
-
-    if (packageDir !== APPS_DIR) {
-      // Test all apps if a common package changes.
-      targets.add("spruce");
-      targets.add("parsley");
-      targets.add("sage");
-    }
```
athammer, line 34: now that we're filtering changes by path filtering, do we still want to redeploy everything if a root package gets changed?

athammer (approved):
everything looks good minus the comment i left

.evergreen/scripts/generate-tasks.js
minnakt: I think we can tweak this as time goes on and see what works best! 👀

minnakt:
I'll just add `activate: true` to match the current waterfall and we can re-evaluate if we don't like it

minnakt:
evergreen retry

PR #1706, opened 2026-06-09
DEVPROD-34692 Support searching for more commits when we run out on waterfall

khelif96:
DEVPROD-34692

### Description
A user was concerned that some sparsely ran tasks are not findable on the waterfall page since they run less frequently then our lookback limit. This PR adds a "Search more" button so users can manually trigger a secondary search.

### Screenshots

image

### DIFF
--- a/apps/spruce/src/pages/waterfall/EmptyState/index.tsx
+++ b/apps/spruce/src/pages/waterfall/EmptyState/index.tsx
@@ -1,11 +1,61 @@
+import { Button } from "@leafygreen-ui/button";
 import { BasicEmptyState } from "@leafygreen-ui/empty-state";
+import { useQueryParam } from "@evg-ui/lib/hooks";
+import { useWaterfallAnalytics } from "analytics";
 import { VERSION_SEARCH_LIMIT } from "../constants";
+import { Pagination, WaterfallFilterOptions } from "../types";
+import { usePaginationNavigation } from "../usePaginationNavigation";
 import { EmptyGraphic } from "./EmptyGraphic";
 
-export const EmptyState: React.FC = () => (
-  <BasicEmptyState
-    description={`Evergreen found no builds matching the applied filters in the ${VERSION_SEARCH_LIMIT} most recent commits.`}
-    graphic={<EmptyGraphic />}
-    title="No Results Found"
-  />
-);
+interface EmptyStateProps {
+  pagination: Pagination;
+}
+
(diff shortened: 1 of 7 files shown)


apps/spruce/src/pages/waterfall/usePaginationNavigation.ts
```diff
@@ -0,0 +1,52 @@
...
+      [WaterfallFilterOptions.Date]: undefined,
+      [WaterfallFilterOptions.MaxOrder]: nextPageOrder,
+      [WaterfallFilterOptions.MinOrder]: undefined,
+      [WaterfallFilterOptions.Revision]: undefined,
+    });
+  };
+
+  const goToPrevPage = () => {
+    setQueryParams({
+      ...queryParams,
```
sophstad: [nit] I believe `setQueryParams` support an updater function now, so this could look something like

```ts
setQueryParams((queryParams) => ({
  ...queryParams,
  // etc.
}));
```
That would make these functions easily memoizable I think

apps/spruce/src/pages/waterfall/EmptyState/index.tsx
```diff
@@ -1,11 +1,61 @@
...
+      />
+    );
+  }
+
+  return (
+    <BasicEmptyState
+      description={
+        searchIsWindowLimited
+          ? "Evergreen found no builds matching the applied filters in the searched commit history."
+          : "Evergreen found no builds matching the applied filters."
```
sophstad: I don't think this conditional is really necessary 🤷‍♀️

apps/spruce/src/pages/waterfall/EmptyState/index.tsx
```diff
@@ -1,11 +1,61 @@
...
+  // other filters search the project's entire history in one query.
+  const searchIsWindowLimited = tasks.length > 0 || statuses.length > 0;
+
+  if (searchIsWindowLimited && hasNextPage) {
+    return (
+      <BasicEmptyState
+        description={`Evergreen found no builds matching the applied filters in the ${VERSION_SEARCH_LIMIT} commits searched. Older commits may still contain matching builds.`}
+        graphic={<EmptyGraphic />}
+        primaryButton={
+          <Button
```
sophstad, line 34: I love this pattern as opposed to just updating the limit, I feel like it will successfully handle other edge cases that could come up 💟

sophstad (approved):

PR #1707, opened 2026-06-09
DEVPROD-13872: Support ANSI -> Jira transformation for copy to JIRA button in Parsley

khelif96:
DEVPROD-13872

### Description
Split from DEVPROD-13835: the SPM team's test failure diffs use red/green ANSI colors, but copying log lines with the "Copy selected contents" button pasted the raw ANSI escape bytes into Jira tickets.

This adds an `ansiToJiraColorMarkup` utility that translates ANSI foreground color codes (standard, bright, 256-color, and truecolor) into Jira `{color:#hex}text{color}` markup and strips all other escape sequences. The hex values mirror ansi_up's default palette, so the pasted colors match what Parsley renders. `getJiraFormat` now applies this per line, which covers both the sharing menu copy and the bookmarked-lines copy in the details menu. Raw copies are unchanged.

Each line is converted independently (matching how Parsley renders ANSI, with a fresh `AnsiUp` per row), and any color still active at the end of a line is closed so non-contiguous selections stay well-formed.

### Screenshots
N/A (changes copied text only)

### Testing
- New unit tests for `ansiToJiraColorMarkup` covering standard/bright/256-color/truecolor codes, resets, style-only codes, background colors, unterminated colors, and non-SGR escapes.
- New `getJiraFormat` test with ANSI-colored input.
- Note: Jira Server/DC renders `{color}` tags inside `{noformat}` blocks; verify with a comment preview on jira.mongodb.org when reviewing.

### Evergreen PR
N/A


athammer (approved):
sorry for the delay, looked this over already and forgot to finally stamp it. this all looks good

PR #1708, opened 2026-06-10
DEVPROD-34802: Reduce waterfall renders

sophstad:
DEVPROD-34802

### Description

So random but this is kinda followup to #1429... I just learned that `@emotion/css` and `@emotion/react` have different runtime behaviors, so the changes in that PR were not as impactful as desired. By using the vanilla version of Emotion, we can create one static class that's applied to all waterfall boxes.

Also reduce the impacts of pins on render, and apply the same Emotion changes to build rows!

### Testing

Improvements Claude measured by using [react-scan](https://react-scan.com) (great tool btw) on the waterfall page:

| Action | Before | After |
|---|---|---|
| Status filter: select 2 | 358ms | 278ms |
| Status filter: clear | 286ms | 148ms |
| Task filter: clear all | 395ms | 202ms |
| Pagination: next page | 370ms | 136ms |
| Pagination: next page (again) | 361ms | 148ms |
| Pagination: prev page | 363ms | 150ms |

### DIFF
--- a/apps/spruce/package.json
+++ b/apps/spruce/package.json
@@ -52,6 +52,7 @@
   },
   "dependencies": {
     "@apollo/client": "^4.1.3",
+    "@emotion/css": "11.13.5",
     "@emotion/react": "11.14.0",
     "@emotion/styled": "11.14.1",
     "@evg-ui/analytics-visualizer": "workspace:*",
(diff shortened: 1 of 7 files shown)


sophstad:
evergreen retry

minnakt (approved):
coool

PR #1711, opened 2026-06-12
DEVPROD-33404: Use TanStack Router in Sage UI

minnakt:
DEVPROD-33404

### Description
After discussion, trying out Tanstack Router in Sage UI may be a good idea. We've historically had issues upgrading React Router and the file-based approach in Tanstack is better than React Router's offering.

How it works:
* routes are defined in `src/routes`
* running `pnpm dev` regenerates `routeTree.gen.ts` which contains all of the generated types. This file should be committed ([docs](https://tanstack.com/router/latest/docs/faq#should-i-commit-my-routetreegents-file-into-git))


apps/sage/src/routeTree.gen.ts
```diff
@@ -0,0 +1,95 @@
+/* eslint-disable */
```
sophstad, line 1: Maybe you could explicitly ignore this in eslint.config.js and tsconfig.json? that may be more efficient/less prone to regeneration changes

apps/sage/src/pages/AgentDetail/index.tsx
```diff
@@ -1,6 +1,8 @@
-import { useParams } from "react-router-dom";
+import { getRouteApi } from "@tanstack/react-router";
+
+const routeApi = getRouteApi("/agents/$agentId/");
```
sophstad: Should we import values from `src/constants/routes.ts` for these?

minnakt: it expects a plain string literal I think, the server would error when I tried to replace it. These are still strongly typed; if you try to use a route like `/random`, it will say that it's not valid

apps/sage/src/routeTree.gen.ts
sophstad: [nit] Should this be removed now?

apps/sage/src/routes/agents/$agentId/runs/$runId.tsx
sophstad, line 1: It seems like one of the benefits of file-based routing (at least from what I can tell in the [Vite example](https://github.com/TanStack/router/tree/main/examples/react/quickstart-file-based)) is colocating the routes with the pages themselves. Would configuring the [`routesDirectory` option](https://tanstack.com/router/latest/docs/installation/with-vite#configuration) potentially be a better option for us?

apps/sage/src/constants/routes.ts
sophstad, line 1: Do you think the `routes` object is potentially vestigial with type-safe routing? I know historically we like using constants instead of typed strings, but it feels awk to use these in some places (like the navbar) and not others.

minnakt: yeah we don't really need it for the actual route definitions. We might need it for Honeycomb `routeConfig` though but I delete it for now 🤔

apps/sage/src/routeTree.gen.ts
minnakt: it's auto-generated! Even if I delete it it comes back 😁

apps/sage/src/routes/agents/$agentId/runs/$runId.tsx
sophstad: I think this works! I suppose [virtual file routing](https://tanstack.com/router/latest/docs/routing/virtual-file-routes) would be closest to what we use with react-router, but I'm not opposed to switching here. (Maybe that will be a better option for Spruce though.)

Also just wanted to call out that using [flat routes](https://tanstack.com/router/latest/docs/routing/file-based-routing#flat-routes) may be helpful at some point if the nesting gets super deep! They can be combined with this approach easily.

apps/sage/src/components/NavBar/index.tsx
```diff
@@ -1,19 +1,19 @@
...
   <NavContainer>
-    <PrimaryLink to={routes.home}>Home</PrimaryLink>
-    <PrimaryLink to={getAgentDetailRoute("sage-bot")}>Agent Info</PrimaryLink>
-    <PrimaryLink to={getAgentRunsRoute("sage-bot", "abcdefg")}>
+    <NavLink to="/">Home</NavLink>
+    <NavLink params={{ agentId: "sage-bot" }} to="/agents/$agentId">
+      Agent Info
+    </NavLink>
+    <NavLink
+      params={{ agentId: "sage-bot", runId: "abcdefg" }}
```
sophstad, line 12: [nit] Is this just test stuff?

apps/sage/tsconfig.json
```diff
@@ -11,6 +11,7 @@
   "exclude": [
     "node_modules",
     "vite.config.ts",
+    "routeTree.gen.ts"
```
sophstad: Wait I can't tell if this is right or if it defeats the point... feel free to remove since Tanstack includes the `@ts-nocheck` tag

sophstad (approved):

apps/sage/src/components/NavBar/index.tsx
minnakt: Yes we'll replace it with real values eventually! (this shouldn't be in the `NavBar` at all lolol)

minnakt:
failures also failing on `main`; merging

PR #1713, opened 2026-06-15
DEVPROD-33404: Set up error boundary in Sage UI

minnakt:
DEVPROD-33404

### Description
Set up error boundary in Sage UI.

Adds Sentry and Honeycomb configuration. I'd like to test it in staging, but the ticket for setting up different environments is still stuck in waiting.


apps/parsley/src/App.tsx
athammer, line 1: does it make sense to add a story book for this work? if not can you add a screen shot of the error boundary in the PR? could reuse the test component possibly?

packages/lib/src/components-via/ErrorFallback/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+const Center = styled.div`
+  display: flex;
+  height: 100%;
+  justify-content: center;
+  align-items: center;
+`;
+
+const Paragraph = styled.div`
+  position: absolute;
```
athammer, line 32: is this meant to be absolute? guessing so but just wanted to sanity check

apps/sage/src/App.tsx
```diff
@@ -20,7 +22,11 @@ const Layout = () => (
 
 const router = createBrowserRouter([
   {
-    element: <Layout />,
+    element: (
+      <ErrorBoundary FallbackComponent={ErrorFallback} homeURL="/">
+        <Layout />{" "}
```
athammer: this {" "} needed?

packages/lib/src/components-via/ErrorFallback/index.tsx
minnakt: yes `absolute`! To overlap the image rather than put it side-by-side

apps/sage/src/App.tsx
minnakt: Not needed sorry is a mistake when I saved the file with a space in it

apps/parsley/src/App.tsx
minnakt: adding!

athammer (approved):
looks good! i also approved the storybook change

minnakt:
failing tests are failing on main; merging

PR #1714, opened 2026-06-15
DEVPROD-35061: Add analytics for diff viewer

sophstad:
DEVPROD-35061

### Description
I was curious about usage of the diff buttons on a patch's changes tab and noticed we didn't have analytics events.

### DIFF
--- a/apps/spruce/src/components/CodeChanges/Table.tsx
+++ b/apps/spruce/src/components/CodeChanges/Table.tsx
@@ -4,6 +4,7 @@ import {
   BaseTable,
   LGColumnDef,
 } from "@evg-ui/lib/components/Table";
+import { useVersionAnalytics } from "analytics";
 import { getFileDiffRoute } from "constants/routes";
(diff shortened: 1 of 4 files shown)


apps/spruce/src/components/CodeChanges/Table.tsx
```diff
@@ -20,8 +21,9 @@ export const Table: React.FC<TableProps> = ({
   moduleIndex,
   patchId,
 }) => {
+  const { sendEvent } = useVersionAnalytics(patchId);
   const table = useLeafyGreenTable<FileDiffsFragment>({
-    columns: getColumns(patchId, moduleIndex, disableDiffLinks),
+    columns: getColumns(patchId, moduleIndex, disableDiffLinks, sendEvent),
```
minnakt: small nit. Can consider turning these into named arguments since now there's quite a lot

minnakt (approved):

PR #1715, opened 2026-06-15
DEVPROD-29223: Prevent values from persisting between distro changes

athammer:
DEVPROD-29223

### Description
Using key based on distro to remount the UI to prevent inputs from not remounting due to being identical between each distro. Pretty common issue in React

### Screenshots

Before
https://github.com/user-attachments/assets/4d9596b0-f18e-4d52-a97d-d39211fa52fd

After
https://github.com/user-attachments/assets/6c513d3e-fdcb-42e3-bee2-3571f7149236

### Testing
* Verified and smoke tested
* Added tests


apps/spruce/src/pages/distroSettings/index.tsx
```diff
@@ -63,7 +63,8 @@ const DistroSettings: React.FC = () => {
   const imageId = data?.distro?.imageId ?? "";
 
   return (
-    <DistroSettingsProvider>
+    // Remount on distro switch so one distro's unsaved edits can't leak into another (DEVPROD-29223).
```
minnakt: nit. I think you'd be good to remove this comment
```suggestion
```

Assuming the test would capture if someone accidentally deleted this

minnakt (approved):

PR #1718, opened 2026-06-16
DEVPROD-35138: Allow display tasks to have event logs

athammer:
DEVPROD-35138

### Description
Display tasks don't have logs, so when we get event logs such as when someone files a ticket at the display task level we are not able to see them. This exposes event logs only for display tasks.

### Screenshots
Screenshot 2026-06-16 at 12 17 15 PM

### Testing
* Added tests
* Tested on prod with Chrome overrides (screen shot)


sophstad (approved):
Nice quick win 😎

PR #1721, opened 2026-06-17
DEVPROD-35234: Set minReleaseAge to 3 days

athammer:
DEVPROD-35234

### Description
Increasing this to three days to give us more time to respond to security incidents.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "repository": "https://github.com/evergreen-ci/ui",
   "private": true,
   "type": "module",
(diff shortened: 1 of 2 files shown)


minnakt (approved):
I think we should also bump `pnpm` versions here if possible

athammer:
good idea. bumped pnpm

minnakt:
Oops sorry you're probably running into a case I didn't consider from implementing build variant path filtering. We can file a ticket to fix; for now, I just scheduled some additional tasks

PR #1728, opened 2026-06-18
DEVPROD-35061: Always include diff format in analytics

sophstad:
DEVPROD-35061

### Description

Followup to #1714 - always include `diff.format` since file diffs are HTML-based.


minnakt (approved):
sgtm

PR #1734, opened 2026-06-22
DEVPROD-35136: Update app URLs to use `VITE_` prefix

minnakt:
DEVPROD-35136

### Description
Part 1: Update URLs to use `VITE_` environment variables.

I will change the remaining `REACT_APP_` environment variables in a different PR because I think it is safer this way.

### Testing
* Deployed apps on staging to check if things were working


apps/sage/.env-cmdrc.json
sophstad, line 1: [nit] Could you alphabetize these vars?

apps/spruce/.env-cmdrc.json
sophstad, line 1: Same here

apps/parsley/.env-cmdrc.json
sophstad, line 1: Here too :D

sophstad (approved):

apps/sage/.env-cmdrc.json
minnakt: yessirrr i will do it

PR #1737, opened 2026-06-23
DEVPROD-34715: Add more metrics for Spruce golden path dashboard

athammer:
DEVPROD-34715

### Description

Adding more metrics so we can create our golden path dashboard for Spruce. Most we already had, filling in the gaps.

Doc: https://docs.google.com/document/d/1JjDwRV6CnQ-g_7bb7srQqfs9oe2g8Hl41sIHtHj3fIw/edit?tab=t.0#heading=h.6t9nsdzj4hw

Board: https://ui.honeycomb.io/mongodb-4b/environments/production/board/oueLptWDmxC/Spruce-Golden-Path-Metrics-DEVPROD-34715

### Testing
* Added some quick sanity checks
* Verified nothing broke locally


packages/lib/src/analytics/hooks/usePageVisibilityAnalytics.ts
```diff
@@ -171,6 +185,46 @@ export const usePageVisibilityAnalytics = ({
     };
   }, [enabled, sendEvent, minDurationMs]);
 
+  // Emit active (foreground) time per route on navigation away/unmount. Route
+  // name is set explicitly; the processor would otherwise stamp the new route.
+  useEffect(() => {
```
athammer, line 192: some of this is duplicated by `apps/spruce/src/hooks/usePageVisibility` but we're specifically looking for a duration and not a boolean so keeping this in

packages/lib/src/utils/observability/AttributeStore/index.ts
```diff
@@ -18,10 +18,7 @@ class OpenTelemetryAttributeStore implements AttributeStore {
   }
 
   removeGlobalAttribute(key: string): void {
-    const value = this.attributesMap.get(key);
```
athammer, line 21: existing falsy bug

packages/lib/src/analytics/hooks/usePageVisibilityAnalytics.ts
athammer: im 50/50 on making this a reusable hook, this is specifically for tracking page duration not any generic duration so didn't extract for that reason but happy to do so if we think we'd reuse it

apps/parsley/src/main.tsx
```diff
@@ -16,13 +16,8 @@ import {
...
 } from "utils/environmentVariables";
 import App from "./App";
-import routes, { slugs } from "./constants/routes";
+import { observabilityRouteConfig } from "./constants/routes";
 
-const routeConfig = {
-  ...routes,
-  // Override the testLogs route to include the groupID parameter so that we can easily identify routes with groupID slugs in Honeycomb.
-  testLogs: `${routes.testLogs}/:${slugs.groupID}?`,
-};
```
athammer, line 26: moving to consts above

apps/spruce/src/constants/routes.ts
```diff
@@ -177,6 +177,19 @@ export const routes = {
...
+export const observabilityRouteConfig = {
+  ...routes,
+  projectSettings: `${routes.projectSettings}/:${slugs.tab}?`,
+  image: `${routes.image}/:${slugs.tab}?`,
+  distroSettings: `${routes.distroSettings}/:${slugs.tab}?`,
+  preferences: `${routes.preferences}/:${slugs.tab}?`,
+  spawn: `${routes.spawn}/:${slugs.tab}?`,
+  jobLogs: `${routes.jobLogs}/:${slugs.taskId}/:${slugs.execution}/:${slugs.groupId}`,
+  patchRedirect: redirectRoutes.patch,
+};
```
athammer, line 191: we need to do something similar to parsley here. from the parsley codebase " // Override the testLogs route to include the groupID parameter so that we can easily identify routes with groupID slugs in Honeycomb."

apps/spruce/src/components/TasksTable/TaskLink.tsx
```diff
@@ -16,11 +17,12 @@ export const TaskLink: React.FC<TaskLinkProps> = ({
   execution,
   onClick = () => {},
   showTaskExecutionLabel,
+  status,
   taskId,
   taskName,
 }) => (
   <StyledRouterLink
-    onClick={() => onClick(taskId)}
+    onClick={() => onClick(taskId, status)}
```
minnakt: since we already have `taskId` / `status` in the parent components, I was wondering if it might be cleaner to pass a plain `onClick={onClick}` function here and have the parent handle what parameters get called?

apps/spruce/src/components/TasksTable/Columns.tsx
```diff
@@ -50,13 +50,14 @@ export const getColumnsTemplate = ({
     cell: ({
       getValue,
       row: {
-        original: { execution, id },
+        original: { displayStatus, execution, id },
       },
     }): React.JSX.Element => (
       <TaskLink
         execution={execution}
         onClick={onClickTaskLink}
```
minnakt: (Continued from last comment) e.g. `onClick={() => onClickTaskLink(id, displayStatus)}`

packages/lib/src/analytics/hooks/usePageVisibilityAnalytics.test.tsx
```diff
@@ -392,4 +397,92 @@ describe("usePageVisibilityAnalytics", () => {
...
+
+    act(() => {
+      router.navigate("/other");
+    });
+
+    expect(mockSendEvent).toHaveBeenCalledWith(
+      expect.objectContaining({
+        name: "System Event route active",
+        "page.route_name": "home",
+        "visibility.duration_ms": expect.any(Number),
```
minnakt: I guess it doesn't really matter but this can be 200 right?
```suggestion
        "visibility.duration_ms": 200,
```

packages/lib/src/analytics/hooks/usePageVisibilityAnalytics.test.tsx
```diff
@@ -392,4 +397,92 @@ describe("usePageVisibilityAnalytics", () => {
...
+      vi.advanceTimersByTime(300);
+    });
+
+    unmount();
+
+    expect(mockSendEvent).toHaveBeenCalledWith(
+      expect.objectContaining({
+        name: "System Event route active",
+        "page.route_name": "home",
+        "visibility.duration_ms": expect.any(Number),
```
minnakt: same comment but 300

apps/spruce/src/pages/Version.tsx
```diff
@@ -57,6 +57,20 @@ export const VersionPage: React.FC = () => {
...
+  // page views be split by type.
+  useEffect(() => {
+    const { AttributeStore } = window;
+    const isPatch = versionData?.version?.isPatch;
+    if (!AttributeStore || isPatch === undefined) {
+      return;
+    }
+    AttributeStore.setGlobalAttribute("version.is_patch", isPatch);
+    return () => {
+      AttributeStore.removeGlobalAttribute("version.is_patch");
```
minnakt, line 70: `useTaskAnalytics`, `useAnnotationAnalytics`, `useVersionAnalytics` already has `is_patch` on them but are there other analytics being called from the version route?

packages/lib/src/analytics/hooks/usePageVisibilityAnalytics.ts
```diff
@@ -171,6 +185,54 @@ export const usePageVisibilityAnalytics = ({
     };
   }, [enabled, sendEvent, minDurationMs]);
 
+  // Flush foreground time for the current route when it is left. Emitting on
+  // "hidden" (not just on cleanup) captures tab close/refresh/bfcache, where
+  // React cleanup never runs. Route name is set explicitly because on navigation
+  // the span processor would otherwise stamp the destination route.
+  useEffect(() => {
```
minnakt, line 192: Does this hook handle if a `pathname` changes (navigation change) but the document continues to be visible? (Is it supposed to handle that)

packages/lib/src/analytics/hooks/usePageVisibilityAnalytics.ts
```diff
@@ -171,6 +185,54 @@ export const usePageVisibilityAnalytics = ({
...
+    if (!enabled || !pathname || !routeConfig) {
+      return;
+    }
+
+    const routeName =
+      calculateRouteName(pathname, routeConfig)?.name ?? pathname;
+
+    let activeMs = 0;
+    let resumeTime: number | null =
+      document.visibilityState === "visible" ? Date.now() : null;
```
minnakt: I guess it doesn't suuuper matter but `performance.now` could be a better way to measure duration metrics (applies to other usage in this hook too)

minnakt (changes requested):

apps/spruce/src/pages/Version.tsx
athammer: this was mostly an attempt to tag all of the related metrics we don't have much control over such as browser performance metrics, so this is just a good way to make sure all of those metrics get tagged. i also think graphql queries would be tagged with this, which would be nice for filtering

packages/lib/src/analytics/hooks/usePageVisibilityAnalytics.ts
athammer: yup to both! have a sanity check test for that here
Screenshot 2026-07-02 at 10 07 51 AM

packages/lib/src/analytics/hooks/usePageVisibilityAnalytics.ts
athammer: good point, will update. edit: done

apps/spruce/src/components/TasksTable/TaskLink.tsx
athammer: good idea, cleaned this up. gonna resolved this and the "visibility.duration_ms" changes as i fixed them 1:1

apps/spruce/src/pages/Version.tsx
minnakt: I wonder if it would work to do this in `useAnalyticsAttributes.ts`, where we already set global attributes, by checking the current pathname? I think that could be a good place to put it if it can work with your implementation; the analytics call here seems somewhat out of place

minnakt (approved):
Overall LGTM (see my final comment above)

apps/spruce/src/pages/Version.tsx
athammer: maybe we can make this into a generic generic handler.

something like 
```
type GlobalAttributes = {
  "user.id": string;
  "version.is_patch": boolean;
}
export const useGlobalAttribute = <K extends keyof GlobalAttributes>(
  key: K,
  value: GlobalAttributes[K] | undefined,
) => {
  useEffect(() => {
    const { AttributeStore } = window;
    if (!AttributeStore || value === undefined) return;
    AttributeStore.setGlobalAttribute(key, value);
    return () => AttributeStore.removeGlobalAttribute(key);
  }, [key, value]);
};
```

bit worried about expanding useAnalyticsAttributes as we'd create a bit of a grab bag of global attributes in useEffects we'd need to add throughout the app as we add more. going to make a ticket for this work though if that's good

athammer: ticket: https://jira.mongodb.org/browse/DEVPROD-36626 - going to merge but will watch this comment in case you want to change anything with that ticket or anything

PR #1738, opened 2026-06-23
DEVPROD-34716: Add Parsley golden path metrics

athammer:
DEVPROD-34716

### Description
Adding more metrics so we can create our golden path dashboard for Parsley. Most we already had, filling in the gaps.

Doc: https://docs.google.com/document/d/1JjDwRV6CnQ-g_7bb7srQqfs9oe2g8Hl41sIHtHj3fIw/edit?tab=t.0#heading=h.6t9nsdzj4hw

Board: https://ui.honeycomb.io/mongodb-4b/environments/production/board/6LzKCQgiWQj/Parsley-Golden-Path-Metrics-DEVPROD-34716

### DIFF
--- a/apps/parsley/src/analytics/logWindow/useLogWindowAnalytics.ts
+++ b/apps/parsley/src/analytics/logWindow/useLogWindowAnalytics.ts
@@ -66,11 +66,9 @@ type Action =
       "settings.jump_to_failing_line.enabled": boolean;
     }
   | {
-      // Fired once per log view on initial load. Flags indicate which precision
-      // params the URL carried, so inbound shared-link rate is measurable.
+      // is_shared_link reflects the ShareLine param, which only the share-link action sets.
       name: "System Event opened log view";
       "share.is_shared_link": boolean;
-      "share.has_share_line": boolean;
       "share.has_filters": boolean;
       "share.has_highlights": boolean;
       "share.has_bookmarks": boolean;
--- a/apps/parsley/src/analytics/logWindow/useLogWindowAnalytics.ts
+++ b/apps/parsley/src/analytics/logWindow/useLogWindowAnalytics.ts
@@ -64,6 +64,24 @@ type Action =
       name: "Viewed log with sections and jump to failing line";
       "settings.sections.enabled": boolean;
       "settings.jump_to_failing_line.enabled": boolean;
+    }
+  | {
+      // Fired once per log view on initial load. Flags indicate which precision
+      // params the URL carried, so inbound shared-link rate is measurable.
+      name: "System Event opened log view";
+      "share.is_shared_link": boolean;
+      "share.has_share_line": boolean;
+      "share.has_filters": boolean;
+      "share.has_highlights": boolean;
+      "share.has_bookmarks": boolean;
+      "share.has_selected_line_range": boolean;
+    }
+  | {
+      // Fired once per viewed project when its ProjectFilters query resolves.
+      name: "System Event loaded project filters";
+      "project.id": string;
+      "project.has_parsley_filters": boolean;
+      "project.parsley_filters_count": number;
     };
 
 export const useLogWindowAnalytics = () =>
--- a/apps/parsley/src/components/Chatbot/index.tsx
+++ b/apps/parsley/src/components/Chatbot/index.tsx
@@ -126,19 +126,13 @@ export const Chatbot: React.FC<{ children: React.ReactNode }> = ({
               openSectionAndScrollToLine(line);
             }
           }}
-          onSendMessage={(message, { conversationId, messageIndex }) => {
+          onSendMessage={(_message, { conversationId, messageIndex }) => {
             if (messageIndex === 0) {
               sendEvent({
                 "conversation.id": conversationId,
                 name: "Created Parsley AI session",
               });
             }
-            sendEvent({
-              "conversation.id": conversationId,
-              message,
-              "message.index": messageIndex,
-              name: "Interacted with Parsley AI",
-            });
           }}
           transformMessage={(message, { pendingChips: chips }) => {
             let transformed = message;
--- a/apps/parsley/src/components/Chatbot/index.tsx
+++ b/apps/parsley/src/components/Chatbot/index.tsx
@@ -140,13 +140,6 @@ export const Chatbot: React.FC<{ children: React.ReactNode }> = ({
               name: "Interacted with Parsley AI",
             });
           }}
-          onToolResult={({ isError, toolName }) => {
-            sendEvent({
-              name: "System Event AI tool result",
-              "tool.is_error": isError,
-              "tool.name": toolName,
-            });
-          }}
           transformMessage={(message, { pendingChips: chips }) => {
             let transformed = message;
             if (chips && chips.length > 0) {
--- a/apps/parsley/src/components/Chatbot/index.tsx
+++ b/apps/parsley/src/components/Chatbot/index.tsx
@@ -126,8 +126,26 @@ export const Chatbot: React.FC<{ children: React.ReactNode }> = ({
               openSectionAndScrollToLine(line);
             }
           }}
-          onSendMessage={(message) => {
-            sendEvent({ message, name: "Interacted with Parsley AI" });
+          onSendMessage={(message, { conversationId, messageIndex }) => {
+            if (messageIndex === 0) {
+              sendEvent({
+                "conversation.id": conversationId,
+                name: "Created Parsley AI session",
+              });
+            }
+            sendEvent({
+              "conversation.id": conversationId,
+              message,
+              "message.index": messageIndex,
+              name: "Interacted with Parsley AI",
+            });
+          }}
+          onToolResult={({ isError, toolName }) => {
+            sendEvent({
+              name: "System Event AI tool result",
+              "tool.is_error": isError,
+              "tool.name": toolName,
+            });
           }}
           transformMessage={(message, { pendingChips: chips }) => {
             let transformed = message;
--- a/apps/parsley/src/components/LogPane/index.tsx
+++ b/apps/parsley/src/components/LogPane/index.tsx
@@ -87,7 +87,8 @@ const LogPane: React.FC<LogPaneProps> = ({ rowCount, rowRenderer }) => {
           "settings.jump_to_failing_line.enabled": jumpToFailingLineEnabled,
           "settings.sections.enabled": sectionsEnabled,
         });
-        const hasShareLine = queryParams[QueryParams.ShareLine] !== undefined;
+        // ShareLine is only set by the share-link action, so it marks an inbound share.
+        const isSharedLink = queryParams[QueryParams.ShareLine] !== undefined;
         const hasFilters = queryParams[QueryParams.Filters] !== undefined;
         const hasHighlights = queryParams[QueryParams.Highlights] !== undefined;
         const hasBookmarks = queryParams[QueryParams.Bookmarks] !== undefined;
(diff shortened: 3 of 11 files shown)


packages/fungi/src/MessageRenderer/ToolRenderer/index.tsx
```diff
@@ -31,15 +31,34 @@ const toolStateToLabelCopy = (
...
 export const ToolRenderer: React.FC<ToolRendererProps> = ({
   onLinkClick,
+  onToolResult,
   progress,
   ...tool
 }) => {
   const [isExpanded, setIsExpanded] = useState(false);
+
+  // Report each tool call once when it reaches a terminal state, so tool
+  // error rate can be measured.
```
sophstad: We already have tool usage and their error rates for Parsley AI measured by sage. I think it would be a lot more meaningful/less duplicative to link these as distributed traces, if that seems doable.

Example: https://ui.honeycomb.io/mongodb-4b/environments/production/datasets/sage/result/rA2RCRCudnK/trace/7eRjfKXfo3q?fields[]=s_name&fields[]=s_serviceName&span=d56d74f3699e96ae

apps/parsley/src/components/LogPane/index.tsx
```diff
@@ -86,6 +87,26 @@ const LogPane: React.FC<LogPaneProps> = ({ rowCount, rowRenderer }) => {
...
+          "share.has_filters": hasFilters,
+          "share.has_highlights": hasHighlights,
+          "share.has_selected_line_range": hasSelectedLineRange,
+          "share.has_share_line": hasShareLine,
+          "share.is_shared_link":
+            hasShareLine ||
+            hasFilters ||
+            hasHighlights ||
+            hasBookmarks ||
+            hasSelectedLineRange,
```
sophstad: I'm not sure if this is totally accurate since we apply those params within a user's own session (not just on share) 🤔 what's the thinking here?

apps/parsley/src/components/Chatbot/index.tsx
```diff
@@ -126,8 +126,26 @@ export const Chatbot: React.FC<{ children: React.ReactNode }> = ({
...
+                "conversation.id": conversationId,
+                name: "Created Parsley AI session",
+              });
+            }
+            sendEvent({
+              "conversation.id": conversationId,
+              message,
+              "message.index": messageIndex,
+              name: "Interacted with Parsley AI",
+            });
```
sophstad: This also feels like information that may be duplicated by the Sage dataset, not entirely sure though.

athammer: tmk this does duplicate, at least the message and conversation id (threadId) fields, but at least for these two events I'm watching for open -> send conversions (toggle parsley AI panel -> interact with parlsey AI) which I don't think the sage data's set has a great way to do that.

I can drop conversation.id and message from `Interacted with Parsley AI` though as that's duplicated, but they're mostly for convenience

packages/fungi/src/MessageRenderer/ToolRenderer/index.tsx
athammer: good idea was too focused on what the UI reported. got rid of this

apps/parsley/src/components/Search/index.tsx
```diff
@@ -63,6 +64,17 @@ const Search: React.FC = () => {
   const { project } = data || {};
   const { parsleyFilters } = project || {};
 
+  useEffect(() => {
+    if (!project) return;
+    sendEvent({
+      name: "System Event loaded project filters",
```
sophstad, line 70: Is this basically to get a denominator for the "Used project filters" event? This information is pretty obtainable via the database, but I can see the use in having it available in Honeycomb. Maybe it'd be preferable to add these attributes to the "System Event opened log view" trace? I feel like this isn't much of an "event" on its own. But lmk if there was other thinking here!

apps/parsley/src/components/Chatbot/index.tsx
sophstad: That makes sense to me! I do think "Interacted with Parsley AI" could be linked to sage's POST traces, what do you think 👀  I did something similar the other week to link the coding agent to the express app in https://github.com/evergreen-ci/sage/pull/683.

apps/parsley/src/components/LogPane/index.tsx
athammer: just removed this, was wrong assumption on url filter states indicating a shared url

apps/parsley/src/components/Search/index.tsx
athammer: nope spot on here, just making it easier to get it in honeycomb and being used an the denominator. doing my best to keep everything in honeycomb so we can use that as the source of truth for our golden path dashboard

apps/parsley/src/components/Chatbot/index.tsx
athammer: Image

sweet. works perfectly!

apps/parsley/src/analytics/logWindow/useLogWindowAnalytics.ts
```diff
@@ -64,6 +64,22 @@ type Action =
...
       "settings.jump_to_failing_line.enabled": boolean;
+    }
+  | {
+      // is_shared_link reflects the ShareLine param, which only the share-link action sets.
+      name: "System Event opened log view";
+      "share.is_shared_link": boolean;
+      "share.has_filters": boolean;
+      "share.has_highlights": boolean;
+      "share.has_bookmarks": boolean;
+      "share.has_selected_line_range": boolean;
```
sophstad, line 75: what do you think about reusing some attributes here? e.g. `share.has_bookmarks` => `bookmark.expression`, `share.has_filters` => `filter.expression`, etc.

apps/parsley/src/components/Chatbot/index.tsx
```diff
@@ -126,8 +126,13 @@ export const Chatbot: React.FC<{ children: React.ReactNode }> = ({
...
             }
           }}
-          onSendMessage={(message) => {
-            sendEvent({ message, name: "Interacted with Parsley AI" });
+          onSendMessage={(_message, { conversationId, messageIndex }) => {
+            if (messageIndex === 0) {
+              sendEvent({
+                "conversation.id": conversationId,
+                name: "Created Parsley AI session",
+              });
```
sophstad, line 134: The distributed traces like [this](https://ui.honeycomb.io/mongodb-4b/environments/staging/datasets/parsley/result/aDdU9N1EZjW/trace/fgWectn7jv8?fields[]=s_name&fields[]=s_serviceName&fields[]=c_mastra.metadata.step.id&span=739566a7e77c780e) are awesome, but I noticed this span doesn't appear there. Is there any way to make it so?

apps/parsley/src/components/LogPane/index.tsx
```diff
@@ -86,6 +87,21 @@ const LogPane: React.FC<LogPaneProps> = ({ rowCount, rowRenderer }) => {
           "settings.jump_to_failing_line.enabled": jumpToFailingLineEnabled,
           "settings.sections.enabled": sectionsEnabled,
         });
+        // ShareLine is only set by the share-link action, so it marks an inbound share.
+        const isSharedLink = queryParams[QueryParams.ShareLine] !== undefined;
+        const hasFilters = queryParams[QueryParams.Filters] !== undefined;
+        const hasHighlights = queryParams[QueryParams.Highlights] !== undefined;
+        const hasBookmarks = queryParams[QueryParams.Bookmarks] !== undefined;
+        const hasSelectedLineRange =
+          queryParams[QueryParams.SelectedLineRange] !== undefined;
```
sophstad, line 96: [nit] I think these are also available via `preferences`, not sure which is preferable

apps/parsley/src/components/Search/index.tsx
sophstad: After thinking about it a bit more, why not just add a `project.parsley_filters` attribute to the "Used project filters" event? Do you think there is a reason to make this information available as a per-load metric? I also feel like our traces in parsley are not well-linked (often orphans), so this may be more difficult to obtain than we'd like in a board.

apps/parsley/src/hooks/useLogDownloader/index.ts
```diff
@@ -110,6 +110,12 @@ const useLogDownloader = ({
               title: "Log not fully downloaded",
             });
           }
+          sendEvent({
+            duration: Date.now() - timeStart,
+            "log.size": getFileSize(),
+            "log.type": logType,
+            name: "System Event log downloaded",
+          });
```
sophstad, line 118: [nit] any reason for this relocation?

packages/fungi/src/ChatFeed/index.tsx
```diff
@@ -65,7 +74,10 @@ export const ChatFeed: React.FC<ChatFeedProps> = ({
   });
 
   const handleSend = (message: string) => {
-    onSendMessage?.(message);
+    onSendMessage?.(message, {
+      conversationId,
```
sophstad, line 78: Nice addition, DEVPROD-34110 will make this work better on the Mastra side 👍

packages/lib/src/utils/observability/honeycomb.ts
```diff
@@ -86,7 +93,10 @@ const initializeHoneycomb = ({
             }
           },
           // Allow connecting frontend & backend traces.
-          propagateTraceHeaderCorsUrls: [new RegExp(backendURL || "")],
+          propagateTraceHeaderCorsUrls: [
+            new RegExp(backendURL || ""),
+            ...tracePropagationURLs.map((url) => new RegExp(url)),
```
sophstad, line 98: Any reason not to just add the Sage URL here?

apps/parsley/src/hooks/useLogDownloader/index.ts
athammer: .finally runs after a .catch so moved it out of .finally to make this successful metric a bit more accurate

packages/lib/src/utils/observability/honeycomb.ts
athammer: mostly to let us expand this later if we ever have anything else we want to trace, was pretty low lift so figured it would be good to make this a bit more generic

apps/parsley/src/components/Search/index.tsx
athammer: this is a bit of a standalone metric that i think won't be too useful to connect to any traces. it's really only needed (at least from what i can see) to calculate the denominator for our conversion metrics as we'll be using this as a COUNT so sadly I don't think we can expand the "Used project filters" metric here as that would be our numerator

apps/parsley/src/components/LogPane/index.tsx
athammer: Image

i think we were missing a few things such as filters and highlights so we'll probably need to use the query params for this case

apps/parsley/src/analytics/logWindow/useLogWindowAnalytics.ts
athammer: i could go either way honestly, leaning towards keeping it the same for the sole reason of keeping this low cardinality approach by using booleans so it's easy query for and make graphs for.

apps/parsley/src/components/Chatbot/index.tsx
athammer: yup can add this! will do that in the next commit

athammer: ah so i think we'd need to start the trace in fungi and add @opentelemetry/api to it in `fungi/src/ChatFeed/index.tsx` as that's where the actual request is? tried making this smaller/isolated to not blow up the pr with this and wasn't able to easily. i can create a new ticket for it though if you think it's worth while

sophstad: I think a ticket would be good to have!

sophstad (approved):

athammer:
oh oops, created the ticket here https://jira.mongodb.org/browse/DEVPROD-36566

PR #1739, opened 2026-06-24
DEVPROD-35136: Migrate remaining `REACT_APP_` variables to use `VITE_` prefix

minnakt:
DEVPROD-35136

### Description
Migrates remaining `REACT_APP_` usage and uninstalls `vite-plugin-env-compatible`.

### Testing
* On staging, verify that `<head>` tags are still populated correctly
* On staging, verify that Sentry reports with correct environment and app version


apps/parsley/vite.config.ts
```diff
@@ -43,15 +44,12 @@ const getProjectConfig = () => {
         // Only Typescript files should use fast refresh.
         include: ["**/*.tsx", "**/*.ts"],
       }),
-      envCompatible({
-        prefix: "REACT_APP_",
-      }),
       injectVariablesInHTML({
```
sophstad: Just a stackoverflow answer but it seems like this may be doable natively now 👀  https://stackoverflow.com/a/77160404

minnakt: ohh ok, let me try some deploys and double check

minnakt:
Prefer not to remove `NODE_ENV` in this PR because it introduces additional risk. It should happen in [DEVPROD-11638](https://jira.mongodb.org/browse/DEVPROD-11638).

apps/spruce/config/injectVariablesInHTML.ts
```diff
@@ -1,45 +0,0 @@
-import { replaceInFileSync } from "replace-in-file";
```
sophstad, line 1: Can we remove this dependency now?

athammer (approved):

apps/spruce/config/injectVariablesInHTML.ts
minnakt: ahh ty!

sophstad (approved):

PR #1720, opened 2026-06-16
DEVPROD-31053: Organize task metadata by common theme

minnakt:
DEVPROD-31053

### Description
Create new components
* `GeneralSection`
* `TimelineSection`
* `ExecutionSection`
* `LinksSection`

and moves each existing setting into one of these categories.

Everything is just copy-pasted into a new section with the exception of [`apps/spruce/src/pages/task/metadata/TimelineSection/Timeline/index.tsx‎`](https://github.com/evergreen-ci/ui/pull/1720/changes#diff-1936a276b15d78576888f32d3aed9a3deaf519c10d39c302eec2d8318a6f72f5), in which time fields are formatted to look more like a timeline.

### Screenshots

<kbd>
Screenshot 2026-06-16 at 2 20 41 PM</kbd>

PR #1745, opened 2026-06-24
DEVPROD-36089: Organize version metadata sidebar

ZackarySantana:
DEVPROD-36089

(similar to #1720)

### Description

This extends the task metadata sidebar organization to the version metadata sidebar.

- Extracts the task timeline styling into a shared `MetadataTimeline` component.
- Reorganizes version metadata into themed sections: General, Timeline, Execution, and External Links.
- Adds the same timeline treatment for version submitted/started/finished timestamps.
- Adds version metadata test and Storybook coverage for the new layout.

#### Version

v2: 
Screenshot 2026-06-25 at 12 37 22 PM

v1:
Screenshot 2026-06-24 at 5 45 28 PM

#### Task

v2:
Screenshot 2026-06-25 at 12 37 41 PM

v1:
Screenshot 2026-06-24 at 5 50 35 PM

### Testing

- Ran pre-commit checks.
- Added/updated unit coverage for version metadata sections and timeline rendering.


apps/spruce/src/pages/task/metadata/BuildVariant/index.tsx
minnakt, line 1: I think changing this card is overkill / unnecessary, and we could leave it as-is (the link shown is also not an external link)

apps/spruce/src/pages/version/Metadata/GeneralSection/index.tsx
```diff
@@ -0,0 +1,220 @@
...
+    owner,
+    repo,
+  } = projectMetadata || {};
+  const hasOwnerAndRepo = !!owner && !!repo;
+
+  const isGithubMergePatch = requester === Requester.GitHubMergeQueue;
+  const isGitHubPullRequest = requester === Requester.GitHubPR;
+
+  return (
+    <MetadataSection title="General">
```
minnakt: nit. we can keep this in a component but let's remove the `"General"` section since we opted not to use section headers for general info

minnakt (changes requested):
Would you mind merging main and updating the usages of `MetadataItem`? It changed a lot from this version

apps/spruce/src/pages/version/Metadata/GeneralSection/index.tsx
ZackarySantana: I removed this!

apps/spruce/src/pages/task/metadata/BuildVariant/index.tsx
ZackarySantana: That makes sense to me, I reverted it!

ZackarySantana: I added new screenshots to the PR desctiption

apps/spruce/src/pages/version/Metadata/ExecutionSection/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+  const { sendEvent } = useVersionAnalytics(id);
+
+  const totalCost = isPatch ? patch?.cost?.total : cost?.total;
+  const isVersionComplete = !!finishTime;
+  const hasChildPatches = (patch?.childPatches?.length ?? 0) > 0;
+  const costTooltip = getCostTooltip(isVersionComplete, hasChildPatches);
+  const hasCost = !!startTime && totalCost != null && totalCost > 0;
+  const hasParameters = parameters.length > 0;
+
+  if (!hasCost && !hasParameters && !costModalOpen) return null;
```
minnakt:
```suggestion
```

`MetadataSection` should be able to inherently handle the case of no children; if it's not working, probably the child conditional && statements need to be adjusted

apps/spruce/src/components/MetadataCard/MetadataTimeline.tsx
```diff
@@ -0,0 +1,108 @@
...
+  dataCy?: string;
+  label: string;
+  timestamp?: Date | null;
+}
+
+interface MetadataTimelineProps {
+  entries: MetadataTimelineEntry[];
+}
+
+export const MetadataTimeline: React.FC<MetadataTimelineProps> = ({
```
minnakt: Is the intention to dedupe this with Task metadata's version of `Timeline`?

apps/spruce/src/components/MetadataCard/MetadataTimeline.tsx
```diff
@@ -0,0 +1,108 @@
...
+      ))}
+    </TimelineContainer>
+  );
+};
+
+const DOT_SIZE = 10;
+const ROW_GAP = 12;
+const LABEL_WIDTH = 60;
+
+const TimelineContainer = styled.ul`
```
minnakt: a couple of the styles changed here as well so just be careful when merging definitions

apps/spruce/src/pages/version/Metadata/GeneralSection/index.tsx
```diff
@@ -0,0 +1,209 @@
...
+    owner,
+    repo,
+  } = projectMetadata || {};
+  const hasOwnerAndRepo = !!owner && !!repo;
+
+  const isGithubMergePatch = requester === Requester.GitHubMergeQueue;
+  const isGitHubPullRequest = requester === Requester.GitHubPR;
+
+  return (
+    <MetadataSection>
```
minnakt, line 62: This change makes sense; can we update the `GeneralSection/index.tsx` in task metadata to be wrapped in a `MetadataSection` too? (to be consistent between implementations)

minnakt (changes requested):

apps/spruce/src/pages/version/Metadata/ExecutionSection/index.tsx
ZackarySantana: That makes sense. I missed that `MetadataSection` returns null when there's no children! I adjusted this

apps/spruce/src/components/MetadataCard/MetadataTimeline.tsx
ZackarySantana: It was but I see with your commit that it won't be as clean so I moved it, lmk if that's what you meant!

apps/spruce/src/components/MetadataCard/MetadataTimeline.tsx
ZackarySantana: Oops.. I tried sharing them now to prevent something like this happening in the future. I appreciate any feedback on it! 🙏

apps/spruce/src/pages/version/Metadata/GeneralSection/index.tsx
ZackarySantana: Sgtm. The less fragments we have the better 🚀

apps/spruce/src/pages/version/Metadata/TimelineSection/Timeline/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export const Timeline: React.FC<TimelineProps> = ({ entries }) => {
+  const visibleEntries = entries.filter(
+    (entry): entry is TimelineEntry & { timestamp: Date } => !!entry.timestamp,
+  );
+
+  if (visibleEntries.length === 0) return null;
+
+  return (
+    <MetadataTimelineContainer>
```
minnakt: Similar to the `Timeline` for task metadata, I think you should just do the same thing here; instead of abstracting out to entries, just list every field manually. It would be better to just be consistent between implementations, since it's odd that they don't use the same abstraction

minnakt (approved):
great job (lgtm with addressing comment)

apps/spruce/src/pages/version/Metadata/TimelineSection/Timeline/index.tsx
ZackarySantana: That sounds good, I switched it to accept a version like task's timeline accepts the task!

PR #1749, opened 2026-06-25
DEVPROD-36017: Use Patch/version instead of versionFull

sophstad:
DEVPROD-36017

### Description

Replace all `Patch/versionFull` queries with `Patch/version` to use the non-REST based type.


minnakt (approved):
yaaay

PR #1752, opened 2026-06-26
DEVPROD-36206: Consolidate cost metadata item

minnakt:
DEVPROD-36206

### Description
I think it's helpful to co-locate the `CostModal` with the corresponding item.

Also I ended up disabling snapshots for the ETA timer because I should have realized a dynamic element won't work with that 😅


apps/spruce/src/pages/version/Metadata/ExecutionSection/index.tsx
```diff
@@ -16,79 +13,26 @@ interface ExecutionSectionProps {
...
   return (
     <MetadataSection title="Execution">
-      {hasCost ? (
-        <MetadataItem
-          data-cy="version-metadata-cost"
-          label="Cost"
-          tooltipDescription={costTooltip}
-        >
-          ${formatCost(totalCost)}
-          {cost != null && isVersionComplete && (
```
athammer, line 48: was this removed in the migration? just checking if that was intentional

apps/spruce/src/pages/task/metadata/ExecutionSection/index.tsx
```diff
@@ -93,34 +88,14 @@ export const ExecutionSection: React.FC<ExecutionSectionProps> = ({ task }) => {
       {testSelectionEnabledForProject && (
         <TestSelection testSelectionEnabled={testSelectionEnabled} />
       )}
-      {finishTime && taskCost?.total != null && (
```
athammer, line 96: same with this check, do we not need finishTime anymore?

apps/spruce/src/pages/version/Metadata/ExecutionSection/index.tsx
```diff
@@ -16,79 +13,26 @@ interface ExecutionSectionProps {
...
-    startTime,
-  } = version;
+  const { cost, id, isPatch, parameters, patch } = version;
   const { sendEvent } = useVersionAnalytics(id);
 
-  const totalCost = isPatch ? patch?.cost?.total : cost?.total;
-  const isVersionComplete = !!finishTime;
-  const hasChildPatches = (patch?.childPatches?.length ?? 0) > 0;
-  const costTooltip = getCostTooltip(isVersionComplete, hasChildPatches);
-  const hasCost = !!startTime && totalCost != null && totalCost > 0;
```
athammer, line 36: sorry one more question lol. `startTime` was dropped from hasCost, just checking to see if that was also needed

athammer (approved):
few questions on behavior, but if they were all intentional looks good

apps/spruce/src/pages/version/Metadata/ExecutionSection/index.tsx
minnakt: the `finishTime` is checked inside the component now!

apps/spruce/src/pages/version/Metadata/ExecutionSection/index.tsx
minnakt: I think `startTime` was not necessary; we can show the cost if it exists

PR #1757, opened 2026-07-01
DEVPROD-36468: Add translateProjectConcurrencyLimit to Runners settings

malikchaya2:
DEVPROD-36468

### Description
Adds a new `translateProjectConcurrencyLimit` field to the Scheduler settings in the Runners tab of admin settings, allowing configuration of the max number of project configs translated concurrently (0 = unlimited).

### Screenshots
image

### Testing
Updated transformer unit tests to cover the new field.

### Evergreen PR
https://github.com/evergreen-ci/evergreen/pull/10429


apps/spruce/src/pages/adminSettings/tabs/GeneralTab/RunnersTab/schemaFields.ts
```diff
@@ -273,6 +273,12 @@ export const scheduler = {
       minimum: 0,
       maximum: 100,
     },
+    translateProjectConcurrencyLimit: {
```
athammer, line 276: no maximum needed here? not a blocker tho

athammer (approved):

apps/spruce/src/pages/adminSettings/tabs/GeneralTab/RunnersTab/schemaFields.ts
malikchaya2: I don't think a maximum is needed. We don't currently have a limit on it and it would be okay to leave it unlimited or set it really high when things are stable if we want to.

PR #1760, opened 2026-07-02
DEVPROD-35285: Create `SageProvider` with authenticated API client

minnakt:
DEVPROD-35285

### Description
Create `SageProvider` that handles auth and creation of a Sage client. Kinda just a best guess of what this should be; note you need to have the Sage backend running in order to login to the application.

Not as big as it looks; 300 lines are just tests.

### Testing
* Added unit tests


apps/sage/package.json
```diff
@@ -7,7 +7,7 @@
     "build": "tsc -b && vite build",
     "build:local": "env-cmd -e local pnpm build",
     "check-types": "tsc -p tsconfig.json --noEmit",
-    "dev": "vite",
+    "dev": "env-cmd -e devLocal pnpm vite",
```
sophstad, line 10: If you're interested in using [Vite env vars & modes](https://vite.dev/guide/env-and-mode.html#html-env-replacement), I think that could potentially be less confusing than our current setup 😅 totally up to you though!

.evergreen/scripts/generate-parallel-e2e-tasks.js
```diff
@@ -144,7 +144,7 @@ const generateParallelPlaywrightTasks = (bv) => {
   const bvTasks = e2eTasks.map(({ name }) => ({ name }));
   const displayTasks = [
     {
-      name: Tasks.E2EParallel,
+      name: E2E_PARALLEL,
```
sophstad: [nit] If you wanna remove the var that works for me
```suggestion
      name: "e2e_parallel",
```

apps/sage/src/api/sageClient.ts
```diff
@@ -0,0 +1,115 @@
...
+export type ApiResult<T> = ApiSuccess<T> | ApiError;
+
+const DEFAULT_TIMEOUT_MS = 30_000;
+
+export class SageClient {
+  private readonly baseURL: string;
+
+  constructor(
+    private readonly logout?: () => void,
+    baseURL: string = sageAPIURL,
```
sophstad: I think switching the order here would allow removing a lot of the `undefined` in the test file (or use an object)

apps/sage/src/api/sageClient.ts
```diff
@@ -0,0 +1,115 @@
...
+    );
+
+    try {
+      const response = await fetch(url, {
+        method: "GET",
+        ...options,
+        credentials: "include",
+        headers: {
+          ...options?.headers,
+        },
```
sophstad: [nit] Unnecessary spread
```suggestion
        headers: options?.headers,
```

apps/sage/src/context/SageProvider/index.tsx
```diff
@@ -0,0 +1,65 @@
...
+  }
+  return ctx;
+};
+
+export const SageProvider: React.FC<{ children: React.ReactNode }> = ({
+  children,
+}) => {
+  const [isAuthenticated, setIsAuthenticated] = useState(false);
+  const [hasCheckedAuth, setHasCheckedAuth] = useState(false);
+  const [authError, setAuthError] = useState<ApiError | null>(null);
```
sophstad: [nit] Could combine into one auth state if you think this provider's state may grow a lot. Not necessary though

apps/sage/package.json
minnakt: I think it'd be best covered by another ticket, if the new team wants to do it

apps/sage/src/context/SageProvider/index.tsx
minnakt: I think I did what you meant, but let me know if it's not right

sophstad (approved):

PR #1761, opened 2026-07-06
DEVPROD-36765: Fix TypeError by ensuring logBucket lifecycleLastSyncedAt is Date object

malikchaya2:
DEVPROD-36765

### Description
The GraphQL Time scalar is mapped to Date in codegen, so TypeScript thought lifecycleLastSyncedAt was a Date. But that mapping is compile-time only. At runtime Apollo hands back the raw JSON string, so .toISOString() threw TypeError whenever the field was actually populated in the DB. When it was null/undefined, optional chaining short-circuited and hid the bug.

Fixed by wrapping all three lifecycleLastSyncedAt values in new Date(...) before calling .toISOString() — for logBucket,
  logBucketLongRetention, and logBucketFailedTasks.

### Testing
Added a case in transformers.test.ts that feeds a populated ISO-string value and asserts it converts correctly.


minnakt (approved):

minnakt:
tweaked the title a bit but you can tweak it again

PR #1769, opened 2026-07-07
DEVPROD-34789: Surface TSS quarantine skipped tests in Spruce for Versions/Builds

athammer:
[DEVPROD-34789](https://jira.mongodb.org/browse/DEVPROD-34789)

### Description
Adding TSS skipped test information to the version page similar to how we have it on the task page (see PR below). 

- Added TSS skipped count on the Version's metadata UI
- Added skipped TSS tests modal where you can search, link to, and download the skipped tests.

First PR was [here](https://github.com/evergreen-ci/ui/pull/1787)

### Screenshots
Screenshot 2026-08-05 at 10 11 15 AM
Screenshot 2026-08-05 at 10 11 33 AM

### Testing
* Unit tests
* Smoke tested locally


athammer:
think the ci is going to fail until the first pr is merged, but this is good to be reviewed (will be blocked until the first pr is merged)

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/VersionSkippedTestsModal.tsx
```diff
@@ -0,0 +1,226 @@
...
+      );
+    }
+  };
+
+  if (!loading && !detailsAvailable) {
+    return null;
+  }
+
+  return (
+    <SkippedTestsModal
```
athammer, line 154: reusing this from the other pr

apps/spruce/src/gql/queries/version-quarantined-tasks.ts
```diff
@@ -0,0 +1,19 @@
+import { gql } from "@apollo/client";
+
+export const VERSION_QUARANTINED_TASKS = gql`
+  query VersionQuarantinedTasks($versionId: String!) {
```
sophstad, line 4: Can we add these fields to the VersionTasks query instead of creating a new one? A primary concern is we don't seem to be applying the same filters as in `VersionTasks` so this wouldn't necessarily return the tasks that are being shown in the table with the given filters.

apps/spruce/src/analytics/version/useVersionAnalytics.ts
```diff
@@ -36,6 +36,9 @@ type Action =
       "task.status": string;
       column: string;
     }
+  | { name: "Clicked version skipped tests details button" }
+  | { name: "Viewed version skipped tests modal" }
```
sophstad, line 40: Aren't these effectively the same?

athammer: pretty much, added two separate events as it lets us have a cheap way to track issues around failing to open the modal

apps/spruce/src/gql/queries/version-quarantined-tasks.ts
athammer: this is suppose to be a version wide rollup so people can see what tasks were quarantined at the version level so no filters on this. 

for using the VersionTask query, i think we can but we'd need to fetch all the tasks  and I was worried about the performance implications of fetching all that data, especially if that query grows over time as some people [have a lot of tasks ](https://spruce.corp.mongodb.com/version/6a564b58fdcb610007b5ad4a/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC)which caused our last TSS outage. so was coding a bit defensively here to keep this tightly bounded and remove risk around people adding anything to VersionTask.

lemme know if you disagree though, im def happy to try with the VersionTask query and I can set up some monitoring/alert. the actual change is pretty small

sophstad: That is a great point, totally missed this. I'm wondering if it would make more sense to have a field on Version that aggregates the total count in order to avoid the looping DB operations. I haven't been able to generate a trace on this operation but I imagine there's a lot going on. It does take ~800ms on that version you linked and ultimately sums to 0 😆 could perform this more expensive operation only once the modal is opened. What do you think?

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+};
+
+export const SkippedTestsMetadata: React.FC<Props> = ({ versionId }) => {
+  const { sendEvent } = useVersionAnalytics(versionId);
+  const [detailsOpen, setDetailsOpen] = useState(false);
+
+  const { data, error, loading, refetch } = useQuery<
+    VersionQuarantinedTasksQuery,
+    VersionQuarantinedTasksQueryVariables
+  >(VERSION_QUARANTINED_TASKS, { variables: { versionId } });
```
sophstad: Should we have a `skip` here so we don't query for projects that don't have TSS enabled? I know at one point you mentioned handling the edge case of a project that previously had TSS enabled but no longer does, but I think preventing this query from firing on _every_ version is worth that tradeoff.

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+        (task) => task.quarantinedTestsSkippedCount > 0,
+      ),
+    [data],
+  );
+  const totalCount = skippedTestTasks.reduce(
+    (sum, task) => sum + task.quarantinedTestsSkippedCount,
+    0,
+  );
+
+  if (loading) {
```
sophstad, line 50: This loading state looks a little awkward for projects that aren't using TSS

athammer:
ticket for the follow up Version query https://jira.mongodb.org/browse/DEVPROD-41302

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/index.tsx
athammer: gated this with `testSelectionEnabled` now so should never reach this for projects not using tss

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/index.tsx
athammer: added this. skipped if tss is off

apps/spruce/src/gql/queries/version-quarantined-tasks.ts
athammer: its been a while so responded before i forget lol, but we talked about this week and will do this work in the follow [up ticket](https://github.com/evergreen-ci/ui/pull/1769#issuecomment-5257927911)

apps/spruce/src/analytics/version/useVersionAnalytics.ts
sophstad: imo not highly necessary but your call :)

apps/spruce/src/components/SkippedTestsModal/index.tsx
```diff
@@ -20,6 +20,7 @@ import { DisplayModal } from "components/DisplayModal";
 
 interface SkippedTestsModalProps<T extends LGRowData> {
   columns: LGColumnDef<T>[];
+  dataTestIdPrefix?: string;
```
sophstad: I feel like this prop may not be necessary since the version vs task context can be obtained otherwise in the test (via page in Playwright or mock data in Vitest). Up to you though!

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+    variables: { versionId },
+  });
+
+  const skippedTestTasks = useMemo(
+    () =>
+      (data?.version.tasks.data ?? []).filter(
+        (task) => task.quarantinedTestsSkippedCount > 0,
+      ),
+    [data],
+  );
```
sophstad: [nit] I don't think memoizing results from Apollo hooks is necessary because they're stateful

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/VersionSkippedTestsModal.tsx
```diff
@@ -0,0 +1,226 @@
...
+  versionId: string;
+}
+
+export const VersionSkippedTestsModal: React.FC<
+  VersionSkippedTestsModalProps
+> = ({ open, setOpen, skippedTestTasks, totalCount, versionId }) => {
+  const dispatchToast = useToastContext();
+  const { sendEvent } = useVersionAnalytics(versionId);
+
+  const taskIds = useMemo(
```
sophstad, line 53: [nit] Could use `taskById.keys()` instead of creating this array

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/VersionSkippedTestsModal.tsx
```diff
@@ -0,0 +1,226 @@
...
+  >(TASK_QUARANTINED_TESTS_SAMPLE);
+
+  const samples = samplesData?.version.taskQuarantinedTestsSample;
+  const detailsAvailable =
+    samples != null && samplesMatchTasks(samples, taskById);
+
+  useEffect(() => {
+    if (loading) {
+      return;
+    }
```
sophstad, line 90: Why should loading be a dependency of this hook at all?

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/VersionSkippedTestsModal.tsx
```diff
@@ -0,0 +1,226 @@
...
+    setOpen(false);
+    if (error) {
+      dispatchToast.error(
+        "There was an error loading the skipped test details.",
+      );
+      return;
+    }
+    dispatchToast.warning(
+      "Skipped test details are not available for this version.",
+    );
```
sophstad: There's a lot of logic here and unfortunately an incomplete dependency array. Maybe a simple hook to show an error toast message would be more appropriate? don't think we need to handle open state either

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/VersionSkippedTestsModal.tsx
```diff
@@ -0,0 +1,226 @@
...
+      setOpen={setOpen}
+      subtitle={`${totalCount} ${pluralize("test", totalCount)} ${
+        totalCount === 1 ? "was" : "were"
+      } skipped by TSS when this version's tasks ran. This snapshot may differ from what TSS would skip now.`}
+      totalCount={totalCount}
+    />
+  );
+};
+
+const samplesMatchTasks = (
```
sophstad: This seems like maybe overkill for the client to validate, assume we'd want it done on the server side. Was there a use case you had in mind?

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/VersionSkippedTestsModal.tsx
```diff
@@ -0,0 +1,226 @@
...
+    });
+    try {
+      const { data: fullData } = await fetchFullList({
+        variables: { versionId, taskIds, limit: FULL_LIST_LIMIT },
+      });
+      const fullSamples = fullData?.version.taskQuarantinedTestsSample;
+      if (!fullSamples || !samplesMatchTasks(fullSamples, taskById)) {
+        throw new Error(
+          "incomplete or mismatched skipped test samples returned",
+        );
```
sophstad: [nit] Instead of this I think you could examine the `error` property returned by `useLazyQuery` and reuse the error toast

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/VersionSkippedTestsModal.tsx
```diff
@@ -0,0 +1,226 @@
...
+    () =>
+      (detailsAvailable ? samples : []).flatMap((sample) => {
+        const task = taskById.get(sample.taskId);
+        return sample.quarantinedTests.map(({ displayTestName, testName }) => ({
+          buildVariantDisplayName: task?.buildVariantDisplayName ?? "",
+          execution: sample.execution,
+          taskDisplayName: task?.displayName ?? sample.taskId,
+          taskId: sample.taskId,
+          testName: displayTestName || testName,
+        }));
```
sophstad: [nit] Instead of constructing this object you could use the `accessorKey` property of the table to get the nested fields. Not necessary though!

sophstad (commented):
Really sorry for the delay!!! Just a few more thoughts

athammer:
all good! was ooo and busy with security tickets that took priority anyways. thanks for the review!

apps/spruce/src/components/SkippedTestsModal/index.tsx
athammer: fixed, went ahead and removed it to use hardcoded strings there

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/index.tsx
athammer: fixed, remove the usememo here

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/VersionSkippedTestsModal.tsx
athammer: we need the array for the graphql query. i think the iterator .keys() would return would fail there. we could wrap it in an Array() though but mostly the same thing

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/VersionSkippedTestsModal.tsx
athammer: agree, going to rethink this useEffect, it's kinda a pain and complicated. will think of a few ways to simplify this

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/VersionSkippedTestsModal.tsx
athammer: good question, lemme think about this. we need loading here so we don't get the inbetween state where error and detailsAvailable is undefined (or false) while loading, but not sure we need it in the dep arrays

athammer: i dont think we need it in the dep arrays, but also probably doesn't hurt to have it in the dep arrays so just going to keep it to so we can keep the linter happy

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/VersionSkippedTestsModal.tsx
athammer: tried extracting this to a hook but it needs so much data ingested it became a mess. going to simplify it a bit, but will need to keep setOpen(false) here so we don't leave the modal opened while it is errored/missing data

update: done tried simplifying this a bit

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/VersionSkippedTestsModal.tsx
athammer: Adding the execution to the taskQuarantinedTestsSample query would help fix this, didn't think about the execution mismatch issue though sadly. However, we can warn users that if a task restarts while on the Version page that this may be inaccurate. We can simplify a lot of this logic as well.

update: added a small warning message that it may be inaccurate in the middle of task restarts

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/VersionSkippedTestsModal.tsx
athammer: lemme know if im misunderstanding something, but i think it would make more sense to keep this non-reactive instead of using something reactive such as useLazyQuery so we can avoid a useffect or something

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/VersionSkippedTestsModal.tsx
athammer: do agree its kinda a mess regardless tho lol. will see what i can do here

athammer: gonna extract this out, its a lot of plumbing that doesn't interact with anything so a good candidate for a local util function imo. edit: done

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/VersionSkippedTestsModal.tsx
athammer: Screenshot 2026-08-21 at 2 25 45 PM

here's the UI for this now.

athammer: made this much quieter 
Screenshot 2026-08-24 at 11 06 48 AM

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  return (
+    <>
+      <MetadataItem as="div">
+        <MetadataContent data-testid="version-skipped-tests-metadata">
+          <SummaryRow>
+            <MetadataLabel>Tests skipped by TSS:</MetadataLabel>
+            <Count data-testid="version-skipped-tests-metadata-count">
+              {totalCount} {pluralize("test", totalCount)}
```
sophstad: [nit] Including the word tests here seems slightly redundant since it's in the label, and removing it would allow the button to be shown on the same line

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/VersionSkippedTestsModal.tsx
```diff
@@ -170,21 +156,36 @@ export const VersionSkippedTestsModal: React.FC<
...
+        <Callout
+          data-testid="version-skipped-tests-restart-warning"
+          title="Restarted tasks can affect these details"
+          variant="warning"
+        >
+          If a task restarts after this page loads, displayed details and
+          downloaded JSON may not match the skipped test count. Refresh the page
+          before relying on them.
+        </Callout>
+      }
```
sophstad: I kinda feel like this is an implicit pattern across Spruce, calling it out almost seems more confusing. It also doesn't seem highly likely that a task would be started in the background by an external process while this modal is open, but before the page is updated via polling.

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/VersionSkippedTestsModal.tsx
```diff
@@ -126,10 +115,8 @@ export const VersionSkippedTestsModal: React.FC<
         variables: { versionId, taskIds, limit: FULL_LIST_LIMIT },
       });
       const fullSamples = fullData?.version.taskQuarantinedTestsSample;
-      if (!fullSamples || !samplesMatchTasks(fullSamples, taskById)) {
-        throw new Error(
-          "incomplete or mismatched skipped test samples returned",
-        );
+      if (!fullSamples) {
+        throw new Error("no skipped test samples returned");
```
sophstad: Is this actually an error? I think instead we'd want to check if the error object on line 114 is defined (this also relates to the `useLazyQuery` comment)

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/VersionSkippedTestsModal.tsx
```diff
@@ -78,43 +83,27 @@ export const VersionSkippedTestsModal: React.FC<
...
-    if (detailsAvailable) {
-      sendEvent({ name: "Viewed version skipped tests modal" });
-      return;
-    }
-    setOpen(false);
-    if (error) {
-      dispatchToast.error(
-        "There was an error loading the skipped test details.",
-      );
+    if (!detailsAvailable) {
```
sophstad, line 98: Same comment, I don't think this actually is an error

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/VersionSkippedTestsModal.tsx
sophstad: It seems like `useLazyQuery` is already being used, no?

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/index.tsx
athammer: good point will see if i can squish this into one line

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/VersionSkippedTestsModal.tsx
athammer: i dont really have any major concerns with removing this especially if you think it will be more confusing to users. will remove!

update: removed

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/index.tsx
athammer: Screenshot 2026-08-24 at 3 52 54 PM

kept the (i) just because we ran into the confusion between around what TSS will skip in follow up runs (answer is we won't really know 😅 )

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/VersionSkippedTestsModal.tsx
athammer: oh yup that's right was too focused on the await part when looking at this. okay makes sense to me, will update

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/VersionSkippedTestsModal.tsx
athammer: this would be separate from the error as our contract is `taskQuarantinedTestsSample?: Maybe<Array<TaskQuarantinedTestsSample>>;` so under normal circumstances this shouldn't happen, looking at the backend code should only happen if we call the API with no task IDs. so this should be mostly just a type narrow so we can access the array

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/VersionSkippedTestsModal.tsx
athammer: removed the error for this, but this only happens if we have no [tasks ids](https://github.com/evergreen-ci/evergreen/blob/main/graphql/version_resolver.go#L317-L319) in the resolver. so this only hits if something really weird is going on. once we get to this point it should always be populated due to the loading and error check above. 

this is more of a completeness issue to cover all paths than what would most likely happen in production. happy to remove it if you think it makes the code read worse though as at least for this one, i don't think its strictly needed for anything

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+  if (!testSelectionEnabled) {
+    return null;
+  }
+
+  if (loading) {
+    return (
+      <MetadataItem as="div">
+        <SummaryRow data-testid="version-skipped-tests-metadata-loading">
+          <MetadataLabel>Tests skipped by TSS:</MetadataLabel>
+          <Skeleton size={SkeletonSize.Small} />
```
sophstad: Super super nit but I'm not sure about using this skeleton because it actually creates more layout shift on account of being on a separate line

Image

sophstad (approved):
yay! Sorry for the extended back and forth, nice job pushing this through 🎉 

I think pulling in main should fix the broken tasks

athammer:
all good! appreciate the reviews, the work is pretty weird but visible due to the nature of TSS so happy to nail this down in the pr review

apps/spruce/src/pages/version/Metadata/SkippedTestsMetadata/index.tsx
athammer: shrunk it to keep it under one line still, good catch.

athammer: Screenshot 2026-08-25 at 1 17 38 PM

PR #1775, opened 2026-07-08
DEVPROD-30382: Reintroduce caching to Project

sophstad:
DEVPROD-30382

### Description

Caching was disabled on the `Project` types because cache normalization broke the rendering of project settings event logs, which use this type. However, it'd be much more helpful to cache projects throughout the app and skip caching for event logs entirely.

- Remove `keyArgs: false` configuration for `Project` cache
- Use `no-cache` fetch policy for settings events
  - Implement "fetchMore with queries that set a `no-cache` fetch policy" as per the [Apollo docs](https://www.apollographql.com/docs/react/pagination/core-api#using-fetchmore-with-queries-that-set-a-no-cache-fetch-policy)
  - Update `useEvents` hook to use simpler params
  - Add reusable "merge" operation so we can delete the cache policies

### Testing

- Update unit tests
- Project dropdown is now persisted when navigating across the site in dev mode


apps/spruce/src/pages/adminSettings/tabs/EventLogsTab/useAdminEvents.ts
```diff
@@ -10,19 +11,33 @@ import { ADMIN_EVENT_LOG } from "gql/queries";
...
+        updateQuery: getEventsUpdateQuery<"adminEvents", AdminEventsQuery>(
+          "adminEvents",
+        ),
+      }),
+    [fetchMoreBase],
+  );
+
   const events = useMemo(
     () => data?.adminEvents?.eventLogEntries ?? [],
     [data],
```
minnakt: nit. is `useMemo` necessary? We are inconsistent about its usage between these hooks

apps/spruce/src/pages/projectAndRepoSettings/shared/tabs/EventLogTab/EventLogTab.tsx
minnakt: nit. while we are refactoring here, I think it could make sense to expose these `handleFetchMore` functions from their respective hooks, so that all of the logic lives in the hook

apps/spruce/src/pages/image/tabs/EventLogTab/useImageEvents.ts
```diff
@@ -14,29 +14,58 @@ export const useImageEvents = (
...
+                ...fetchMoreResult.image.events,
+                eventLogEntries: [
+                  ...(previousData.image?.events.eventLogEntries ?? []),
+                  ...fetchMoreResult.image.events.eventLogEntries,
+                ],
+              },
+            },
+          };
+        },
+      }),
```
minnakt: nit. I think it could be reasonable to extract this out to a function at the top of file

```
const imageEventsUpdateQuery = ( // stuff...
```

minnakt (approved):

apps/spruce/src/pages/adminSettings/tabs/EventLogsTab/useAdminEvents.ts
sophstad: It's not 💯

PR #1777, opened 2026-07-08
DEVPROD-37127: Upgrade to TypeScript 7

sophstad:
DEVPROD-37127

### Description

- Upgrade TypeScript to Go-based version
  - Because 7.0 does not ship with an API, some dependencies like `typescript-eslint` and our own analytics package still require v6. I followed the instructions from [Running Side-by-Side with TypeScript 6.0](https://devblogs.microsoft.com/typescript/announcing-typescript-7-0/#running-side-by-side-with-typescript-6.0) found in the release announcement.
- Also remove pnpm overrides that are out of date (removing them didn't change any dependencies in `pnpm-lock.yaml`)
- Upgrade Storybook to support TS 7

### Testing
- This change shaves ~30s off Spruce's `tsc` execution time on CI
- Locally, `pnpm check-types` on Spruce was reduced from ~20s to about 6s 😎


apps/parsley/package.json
```diff
@@ -127,7 +127,8 @@
     "serve": "^14.2.5",
     "storybook": "^10.1.10",
     "storybook-addon-apollo-client": "^10.0.0",
-    "typescript": "~6.0.2",
+    "typescript": "npm:@typescript/typescript6@~6.0.2",
+    "typescript-7": "npm:typescript@~7.0.2",
```
khelif96, line 131: Lame that we have to do this. 
Do we need the `npm:....` filler?

khelif96 (approved):

minnakt (approved):

apps/parsley/package.json
sophstad: Yeah, this is how a package alias is defined across npm, pnpm, and Yarn, so it's necessary to have these to install multiple versions.

And +1, lint is so slow that it would be really nice to leverage the improvements there as well. Soon 🤞

PR #1778, opened 2026-07-08
DEVPROD-36433: Disable spawnable checkbox for single task distros

minnakt:
DEVPROD-36433

### Description
Two changes:
* Make the spawnable checkbox disabled if a distro is a single task distro.
* If a distro is being changed to be a single task distro, set its spawnable field to false.

Distro creation is not an issue because the distro will be `spawnable: false` by default ([source](https://github.com/evergreen-ci/evergreen/blob/02c79860dc8f9c1fdb867bdb6f46d01fc10796ca/rest/data/distro.go#L125-L149)).

### Before Merge
<s>Before merging, I will run a migration script to fix all existing distros with this error. I will use the ops channel to ask for LGTM.</s> **Done**.

### Testing
- Unit and e2e tests


sophstad (approved):

minnakt:
will merge the codegen fix on top

PR #1780, opened 2026-07-08
DEVPROD-15014: Use stylistic eslint plugin

khelif96:
### Description
ESLint deprecated its core stylistic rules in favor of the [@stylistic](https://eslint.style/) plugin. Since Prettier handles formatting in this repo, the only remaining deprecated stylistic rule was `spaced-comment`. This PR adds `@stylistic/eslint-plugin` to `@evg-ui/eslint-config` and replaces `spaced-comment` with `@stylistic/spaced-comment` (including the GraphQL virtual-docs override), removing the deprecation TODO.

### Testing
- `pnpm eslint:strict` passes across all workspaces.
- Verified the rule still fires by adding a `//bad comment` to a Spruce file and confirming ESLint reports `@stylistic/spaced-comment`.

### DIFF
--- a/packages/eslint-config/eslint.config.js
+++ b/packages/eslint-config/eslint.config.js
@@ -1,5 +1,6 @@
 import * as emotionPlugin from "@emotion/eslint-plugin";
 import { fixupPluginRules } from "@eslint/compat";
+import stylisticPlugin from "@stylistic/eslint-plugin";
 import eslint from "@eslint/js";
 import graphqlPlugin from "@graphql-eslint/eslint-plugin";
 import { defineConfig } from "eslint/config";
@@ -65,6 +66,7 @@ const eslintConfig = {
   name: "@eslint/js/rules",
   plugins: {
     "@eslint/js": eslint,
+    "@stylistic": stylisticPlugin,
   },
   files: ["**/*.js?(x)", "**/*.ts?(x)"],
   rules: {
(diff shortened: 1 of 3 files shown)


packages/eslint-config/eslint.config.js
```diff
@@ -65,6 +66,7 @@ const eslintConfig = {
   name: "@eslint/js/rules",
   plugins: {
     "@eslint/js": eslint,
+    "@stylistic": stylisticPlugin,
   },
```
minnakt, line 69: I think this belongs in its own section because when you run `pnpm eslint --inspect-config`, it's grouped under the ESLint rules which doesn't entirely match

packages/eslint-config/eslint.config.js
minnakt, line 63: Also can consider applying `stylisticPlugin.configs.recommended.rules` if it doesn't throw a bunch of errors in our files

packages/eslint-config/eslint.config.js
```diff
@@ -369,7 +372,7 @@ const graphQLConfig = {
     ],
     "@graphql-eslint/no-deprecated": ERROR,
     "@graphql-eslint/selection-set-depth": [ERROR, { maxDepth: 8 }],
-    "spaced-comment": OFF,
+    "@stylistic/spaced-comment": OFF,
```
minnakt: I wonder if it can be removed, maybe they made some improvements to the stylistic config
```suggestion
```

minnakt (changes requested):

minnakt (approved):

PR #1787, opened 2026-07-13
DEVPROD-34789: Add task-level TSS quarantine-skipped tests UI

athammer:
DEVPROD-34789

### Description
Adding visibility to skipped tests due to TSS on the task level.

- Added TSS skipped count on the Task's metadata UI
- Added skipped TSS tests modal where you can search and download the skipped tests.
- Added unit tests and storybooks

Part two that adds this to the version level is here: https://github.com/evergreen-ci/ui/pull/1769

- [x] Won't merge till the BE PR populating these values is merged + deployed

### Screenshots
From storybook

Inlined on Task metadata section
Screenshot 2026-08-04 at 9 46 44 AM

Modal when you click on the above text.
Screenshot 2026-08-04 at 9 47 01 AM

Video (ignore the end was fake data).
https://github.com/user-attachments/assets/4b8e3fac-6179-4fea-bda7-7ca9ecf81e60

### Testing
* Unit tests
* Story tests/local smoke tests


apps/spruce/src/pages/task/metadata/ExecutionSection/QuarantinedTestsSkipped/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+    return null;
+  }
+
+  return (
+    <MetadataItem as="div" label="Quarantined test skips">
+      <BadgeWrapper data-cy="quarantined-test-skips">
+        <InfoSprinkle baseFontSize={BaseFontSize.Body1}>
+          Tests skipped on this execution because they were quarantined in TSS
+          when it ran. This is an execution-time snapshot, not the live
+          quarantine state.
```
athammer: this is a confusing concept so wanted to call it out here. this is _not_ the quarantined state but the tests TSS skipped due to most likely being quarantined

apps/spruce/src/components/SkippedTestsModal/SkippedTestsModal.stories.tsx
```diff
@@ -0,0 +1,60 @@
...
+    <QuarantinedTestsModal
+      columns={columns}
+      dataCyPrefix="story-quarantined-tests"
+      getSearchText={({ testName }) => testName}
+      onClickDownload={() => {}}
+      open
+      rows={getRows(8)}
+      searchPlaceholder="Search test names"
+      setOpen={() => {}}
+      subtitle="8 tests were skipped by TSS when this task ran. This snapshot may differ from what TSS would skip now."
```
athammer, line 39: this is a confusing concept so wanted to call it out here. this is not the quarantined state but the tests TSS skipped due to _most likely_ being quarantined and may be change in the future. i dont think its worth going into detail here and better to treat it as a black box

apps/spruce/src/pages/task/metadata/ExecutionSection/SkippedTestsMetadata/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  return (
+    <MetadataItem as="div" label="Tests skipped by TSS">
+      <BadgeWrapper data-cy="skipped-tests-metadata">
+        <InfoSprinkle baseFontSize={BaseFontSize.Body1}>
+          Tests skipped by TSS when this execution ran. This snapshot may differ
+          from what TSS would skip now.
+          {execution !== latestExecution &&
+            " Test names are only available for the latest execution."}
+        </InfoSprinkle>
```
sophstad: [nit] I think showing the info sprinkle after the badge makes more sense

apps/spruce/src/pages/task/metadata/ExecutionSection/index.tsx
```diff
@@ -88,6 +90,13 @@ export const ExecutionSection: React.FC<ExecutionSectionProps> = ({ task }) => {
       {testSelectionEnabledForProject && (
         <TestSelection testSelectionEnabled={testSelectionEnabled} />
       )}
+      <SkippedTestsMetadata
```
sophstad, line 93: [nit] It would seem a little clearer to me to conditionally render this component like we do above

```suggestion
      {testSelectionEnabledForProject && <SkippedTestsMetadata
```

apps/spruce/src/pages/task/metadata/ExecutionSection/SkippedTestsMetadata/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+  // actually ran on this task.
+  if (count === 0 && !testSelectionEnabled) {
+    return null;
+  }
+
+  const detailsAvailable = count > 0 && execution === latestExecution;
+  const badge = (
+    <Badge
+      data-cy="skipped-tests-metadata-badge"
+      variant={count === 0 ? BadgeVariant.Green : BadgeVariant.Yellow}
```
sophstad: Could you share more insight into why this is a badge? Does having some tests skipped imply investigation is required? That's what the yellow badge suggests, imo

apps/spruce/src/gql/queries/task.ts
```diff
@@ -125,6 +125,7 @@ export const TASK = gql`
           allowed
         }
       }
+      quarantinedTestsSkippedCount
```
sophstad, line 128: On the page you sent me via Slack, this is returned as 0 🤔 should it be working yet?

apps/spruce/src/pages/task/taskTabs/testsTable/TestsTable.tsx
```diff
@@ -178,34 +179,37 @@ const TestsTable: React.FC<TestsTableProps> = ({ task }) => {
...
-          filteredCount={filteredTestCount}
-          label="tests"
-          // @ts-expect-error: FIXME. This comment was added by an automated script.
-          limit={limitNum}
-          onClear={clearQueryParams}
-          onPageSizeChange={() => {
-            sendEvent({ name: "Changed page size" });
-          }}
+    <>
+      <SkippedTests task={task} />
```
sophstad: Similarly, I'd only render this component if the count > 0. Then we don't have to worry about all of that modal mount stuff

apps/spruce/src/pages/task/taskTabs/testsTable/SkippedTests/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+export const SkippedTests: React.FC<SkippedTestsProps> = ({ task }) => {
+  const { execution, id: taskId, quarantinedTestsSkippedCount: count } = task;
+  const versionId = task.versionMetadata.id;
+  const dispatchToast = useToastContext();
+  const { sendEvent } = useTaskAnalytics();
+  const [showModal, setShowModal] = useState(false);
+  const [queryParams, setQueryParams] = useQueryParams();
+
+  const shouldAutoOpen = queryParams[QueryParams.SkippedTests] === true;
```
sophstad: What's the thinking behind supporting this auto-open modal? It seems somewhat intrusive, but where would it be linked from?

apps/spruce/src/pages/task/metadata/ExecutionSection/SkippedTestsMetadata/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+            onClick={() =>
+              sendEvent({
+                name: "Clicked skipped tests metadata link",
+              })
+            }
+            to={getTaskRoute(taskId, {
+              execution,
+              tab: TaskTab.Tests,
+              [QueryParams.SkippedTests]: true,
+            })}
```
sophstad: Using a query param causes the table to re-enter the loading state, any way we could open the modal more statefully? I think it could be included as a component here

apps/spruce/src/pages/task/metadata/ExecutionSection/SkippedTestsMetadata/index.tsx
sophstad: Ah I see, the badge is the button to open the modal. Maybe text count + a details button like we do for cost would be clearer? 

Image

sophstad (changes requested):

apps/spruce/src/pages/task/metadata/ExecutionSection/index.tsx
athammer: think we should keep this just for the fact this is a historical snapshot, so if someone turns off TSS we'd still want people to see what tests were skipped during their test execution. it's a bit of an edge case though as if TSS is off this should pretty much never show

apps/spruce/src/gql/queries/task.ts
athammer: for now sadly, but the [BE](https://github.com/evergreen-ci/evergreen/pull/10573) work should be merged soon! once that is we should start being able to test this and merge this pr. (have been using mock data for now)

apps/spruce/src/pages/task/taskTabs/testsTable/SkippedTests/index.tsx
athammer: the idea was basically if you clicked the tss badge you'd be sent to the test tab with this query param which would then open it up. moved this to local state though and removed that behavior as it might be too annoying tbh. now it just opens up the modal wherever you are.

apps/spruce/src/pages/task/metadata/ExecutionSection/SkippedTestsMetadata/index.tsx
athammer: solved in combination with the above fix as well. opens up on the page without redirecting tabs now

apps/spruce/src/pages/task/metadata/ExecutionSection/SkippedTestsMetadata/index.tsx
athammer: good idea, fixed this

apps/spruce/src/pages/task/metadata/ExecutionSection/SkippedTestsMetadata/index.tsx
athammer: updated

apps/spruce/src/pages/task/taskTabs/testsTable/TestsTable.tsx
athammer: okay [fixed](https://github.com/evergreen-ci/ui/pull/1787/changes#diff-4e82466e6669639c7041d47bfe14ad359312927d12b75ceb085e75d74c1bbce6R69-R76). did move this though as it was in TestTable and making the skippedtest modal less dependent on the test table tab

athammer:
will update the screenshots when i get back on tuesday

athammer:
updated the screen shots. here's the new ui changes. much less in your face which is nice.
image

apps/spruce/src/pages/task/metadata/ExecutionSection/index.tsx
sophstad: That makes sense!

apps/spruce/src/components/SkippedTestsModal/SkippedTestsModal.stories.tsx
sophstad: Is this the reason for "0 tests skipped" in the sidebar, but a test in the middle indicates it was skipped? ([src](https://spruce.corp.mongodb.com/task/mongodb_mongo_master_enterprise_amazon_linux2023_arm64_all_feature_flags_jsCore_patch_5785c96e76654498d13ca0835d8b93c5c57304b0_6a284a3e0c100b000786b709_26_06_09_17_16_01/tests?execution=0&sorts=STATUS%3AASC))
Image

apps/spruce/src/components/SkippedTestsModal/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+          Download JSON
+        </Button>
+      </HeaderRow>
+      {!loading && rows.length < totalCount && (
+        <Disclaimer data-cy={`${dataCyPrefix}-truncation-note`}>
+          Showing the first {rows.length} of {totalCount} tests. Download the
+          JSON for all available stored tests.
+        </Disclaimer>
+      )}
+      <OverflowContainer>
```
sophstad, line 97: [nit] I encounter a kinda weird double-scrollbar in the modal, would be nice if the table didn't make the modal overflow itself (lmk if that makes sense)

apps/spruce/src/gql/queries/task.ts
sophstad: Just flagging this doesn't seem true 😢  (see screenshot above)

packages/lib/src/utils/request/index.ts
```diff
@@ -22,6 +22,38 @@ export const getUserStagingHeader = (): {
...
+  }, 1000);
+};
+
+/**
+ * downloadObjectAsJson serializes a value to pretty-printed JSON and triggers a
+ * browser download of it as a .json file.
+ * @param data - the value to serialize
+ * @param filename - the name of the file to save
+ */
+export const downloadObjectAsJson = (data: unknown, filename: string) => {
```
sophstad, line 50: [nit] Maybe could combine these into one downloader function?

sophstad (approved):
Still wasn't able to test against prod to validate the download button/modal, so lmk if there's any way to do that! Looks pretty much good though

athammer:
will reverify now that the BE changes are in. it doesn't backfill anything though so that example version should still report 0 test skipped. but if we reran it, it should work but will verify and post a screenshot.

okay nvm think there's also aggressive caching here so going to [follow up](https://mongodb.slack.com/archives/C0A799RFXJS/p1785960963281119) with TSS so people don't get confused why tests are not skipped. will still verify and test once the cache is busted though, dont think the caching part blocks the pr

athammer:
okay tested this on staging to confirm with real data instead of mocked data. can see the test isn't run because of TSS and shows up in the TSS modal as skipped. With resmoke we'd actually see the skipped test in the UI like in the patch you linked.

reminds me once we move off of project using resmoke we'll need to do more work on the evergreen to show skipped tests. that resmoke environment actually has a second "is tss enabled" gate for themselves which makes this all the more complicated so even though it's enabled on our side, it's disabled on their side 😭 

[ticket](https://jira.mongodb.org/browse/DEVPROD-40975)

https://github.com/user-attachments/assets/6224b108-0d8d-41d2-aafa-b9ab07ffcad0

apps/spruce/src/components/SkippedTestsModal/index.tsx
athammer: wasn't sure if i was able to replicate so might be way off but added a max-height which should help with this. if it doesn't lemme know and can file a ticket for it

packages/lib/src/utils/request/index.ts
athammer: saveBlobToFile was extracted/reused from `downloadFile`, `downloadObjectAsJson` is really just a subset of `downloadFile` as the full function is a bit more opinionated so couldn't use it sadly (takes a url not a blob).

apps/spruce/src/components/SkippedTestsModal/index.tsx
athammer: here's the behavior, the background kinda jiggles a bit but not double scrollbar, verified in safari as well so hopefully its fixed.
https://github.com/user-attachments/assets/4909c861-358d-432b-a356-1a48d10af157

athammer:
getting codegen failures, going to try to fix in this pr

PR #1795, opened 2026-07-16
DEVPROD-32813: Reorganize admin settings

sophstad:
DEVPROD-32813

### Description

- Link straight to the feature flags page from "admin settings" in the navbar. I opted to keep the general settings page as the default though.
- Do some reorganizing: no need to fetch admin settings data from the admin settings root. Instead, it can be handled by the page component.

### Testing

Existing tests passing


apps/spruce/src/pages/adminSettings/tabs/GeneralTab/GeneralTab.tsx
```diff
@@ -9,33 +20,73 @@ import { ProvidersTab } from "./ProvidersTab/ProvidersTab";
...
+    [adminSettings],
+  );
+
+  useEffect(() => {
+    if (tabData) {
+      setInitialData(tabData);
+    }
+  }, [setInitialData, tabData]);
+
+  useScrollToAnchor();
```
athammer: does this need to be run after `loading` is done and/or after it's done and if there's tab data?

athammer (approved):

apps/spruce/src/pages/adminSettings/tabs/GeneralTab/GeneralTab.tsx
sophstad: fixed 🫡

PR #1796, opened 2026-07-16
DEVPROD-29652: Remove `@emotion/eslint-plugin`

minnakt:
DEVPROD-29652

### Description
I don't think we've seen a lot of benefits from this plugin. 

It hasn't been updated for 2 years and was never updated to support ESLint 9.

### DIFF
--- a/apps/spruce/src/components/TaskBox/index.tsx
+++ b/apps/spruce/src/components/TaskBox/index.tsx
@@ -1,6 +1,4 @@
 import { forwardRef } from "react";
-// Using non-React Emotion generates a static class, avoiding runtime performance impacts on pages like the waterfall.
-// eslint-disable-next-line @emotion/no-vanilla
 import { css as classNameCss } from "@emotion/css";
 import { css } from "@emotion/react";
 import styled from "@emotion/styled";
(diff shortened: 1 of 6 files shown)


apps/spruce/src/components/TaskBox/index.tsx
```diff
@@ -1,6 +1,4 @@
 import { forwardRef } from "react";
-// Using non-React Emotion generates a static class, avoiding runtime performance impacts on pages like the waterfall.
```
sophstad, line 2: [nit] I think the first line is helpful context since we don't do this anywhere else

apps/spruce/src/pages/waterfall/BuildRow.tsx
```diff
@@ -1,6 +1,4 @@
 import { memo, useCallback, useEffect, useMemo, useRef, useState } from "react";
-// Using non-React Emotion generates a static class, avoiding runtime performance impacts on pages like the waterfall.
```
sophstad, line 2: same as above

sophstad (approved):

apps/spruce/src/components/TaskBox/index.tsx
minnakt: I'll add back these comments

PR #1798, opened 2026-07-16
DEVPROD-29652: Share `sort-imports` ESLint rule across all projects

minnakt:
DEVPROD-29652

### Description
As part of upgrading ESLint, I'm trying to unify Spruce and Parsley ESLint configs and stop having random ESLint rules in different projects. All projects should have the same, consistent rules.

This PR makes `sort-imports` shared across projects instead of Parsley specific.

Every fix was generated via `pnpm eslint:fix`


athammer (approved):
looks like a noop behavior change to me? couldn't find anything other than that so assuming the lint is the only change this all looks good to me.

packages/eslint-config/eslint.config.js
```diff
@@ -119,6 +119,7 @@ const eslintConfig = {
     ],
     "prefer-regex-literals": [ERROR, { disallowRedundantWrapping: true }],
     "prefer-template": ERROR,
+    "sort-imports": [ERROR, { ignoreDeclarationSort: true }],
```
sophstad, line 122: Why is this rule applied?

minnakt: It tells the linter to ignore checking the order of import lines and only validate the alphabetical sorting of individual members within the import brackets

If we turned it off, it would conflict with `eslint-plugin-import`

sophstad (approved):

PR #1799, opened 2026-07-16
PLATFORMS-30919 / PLATFORMS-30918: Container isolation distro settings and task badge

wratner:
### Description

Two UI tasks for GOAL-279 (per-task Docker container isolation), rounding out Phase 0 alongside the backend work on `wratner/evergreen`.

**PLATFORMS-30919: Distro configuration settings for container isolation**
- Adds a "Container Isolation" card to the distro Host tab exposing `enabled`, `image` (ECR URI), and `requireIsolation`, backed by `BootstrapSettings.containerIsolation` in GraphQL.
- Client-side validation mirrors the backend's `Distro.ValidateBootstrapSettings`: `image` required when `enabled`, exec user required when `enabled` (reusing the existing top-level `execUser` field rather than duplicating it), and `requireIsolation` rejected when `enabled` is false.
- Card is hidden for non-Linux architectures, matching the existing `resourceLimits` convention in the same file.
- Adds `rawErrors` support to the shared `LeafyGreenCheckBox` widget so checkbox-backed fields can surface inline validation errors — previously any error attached to a checkbox field silently blocked Save with no visible explanation.

**PLATFORMS-30918: Surface when a task ran in a container**
- Adds a blue "Container" badge to the Spruce task metadata page (Host Information card) and a smaller secondary badge to Parsley's task breadcrumb, driven by the new `Task.executionPlatform` GraphQL field.
- Verified Parsley's log-rendering pipeline has no platform-dependent branching (no regression risk for container-task logs).

### Screenshots
See linked staging verification in ticket comments.

### Testing
- `pnpm check-types` and `pnpm eslint:strict` clean in both `apps/spruce` and `apps/parsley`.
- Unit tests added/updated for HostTab validation, the new checkbox widget error path, Spruce task metadata badge, and Parsley breadcrumb badge.
- Full end-to-end verification on staging against the corresponding evergreen backend branch (`platforms-30919-distro-container-isolation` on `wratner/evergreen`), including all three client-side validation paths, successful save/round-trip, and the container badge rendering against a real container-run task.

### Evergreen PR
Corresponding backend branch: `wratner/evergreen@platforms-30919-distro-container-isolation` (PR pending).

### DIFF
--- a/apps/parsley/src/components/SubHeader/EvergreenTaskSubHeader.test.tsx
+++ b/apps/parsley/src/components/SubHeader/EvergreenTaskSubHeader.test.tsx
@@ -4,7 +4,10 @@ import {
   screen,
   waitFor,
 } from "@evg-ui/lib/test_utils";
+import { ApolloMock } from "@evg-ui/lib/test_utils/types";
 import { LogTypes } from "constants/enums";
+import { TaskQuery, TaskQueryVariables } from "gql/generated/types";
+import { GET_TASK } from "gql/queries";
 import { evergreenTaskMock } from "test_data/task";
 import { EvergreenTaskSubHeader } from "./EvergreenTaskSubHeader";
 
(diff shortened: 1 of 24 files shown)


wratner:
Screenshot 2026-07-08 at 2 37 23 PM
Screenshot 2026-07-08 at 2 35 05 PM

athammer:
thanks for the behavior change screenshots! will take a look

apps/parsley/src/components/SubHeader/EvergreenTaskSubHeader.test.tsx
```diff
@@ -32,4 +39,82 @@ describe("evergreen task subheader", () => {
...
+    await waitFor(() => {
+      expect(screen.getByText("check_codegen")).toBeInTheDocument();
+    });
+    expect(
+      screen.queryByTestId("task-execution-platform-badge"),
+    ).not.toBeInTheDocument();
+  });
+
+  it("renders a Container badge for a task that ran in a container", async () => {
+    const containerTaskMock: ApolloMock<TaskQuery, TaskQueryVariables> = {
```
athammer: should be able to reuse evergreenTaskMock and spread the execution to avoid writing the full task inline

apps/spruce/src/pages/distroSettings/tabs/HostTab/schemaFields.tsx
```diff
@@ -708,3 +743,25 @@ export const sshConfig = {
...
+    enabled: containerIsolationEnabled.schema,
+    image: containerIsolationImage.schema,
+    memoryMb: { type: "integer" as const },
+    requireIsolation: containerIsolationRequireIsolation.schema,
+  },
+  uiSchema: (architecture: Arch) => ({
+    "ui:ObjectFieldTemplate": CardFieldTemplate,
+    // Container isolation is only supported on Linux.
+    ...(!linuxArchitectures.includes(architecture) && {
+      "ui:widget": "hidden",
```
athammer, line 759: this hides the values in the UI but I don't think it actually removes the values we store/send to the API. if that's the case we might need to do something like this in the transformer if it's an issue https://github.com/evergreen-ci/ui/blob/8eb60eca2ed9a4073a535e101e00604db5a3400a/apps/spruce/src/pages/distroSettings/tabs/ProviderTab/transformers.ts#L64-L107

athammer (commented):
just one [blocker](https://github.com/evergreen-ci/ui/pull/1799/changes#r3908401165) and one nit. overall looks good!

apps/spruce/src/pages/distroSettings/tabs/HostTab/schemaFields.tsx
wratner: Confirmed that hiding the section does not remove its form values. `formToGql` now resets the complete container-isolation block to disabled defaults when the selected architecture is non-Linux, with transformer coverage for stale enabled values. Linux submissions continue preserving `cpus` and `memoryMb` because those are supported persisted backend settings.

apps/parsley/src/components/SubHeader/EvergreenTaskSubHeader.test.tsx
wratner: Updated the container case to derive from `evergreenTaskMock` and override only the request task ID, task ID, and `executionPlatform`. This removes the duplicated task payload.

apps/spruce/src/pages/distroSettings/tabs/HostTab/HostTab.tsx
```diff
@@ -48,5 +50,25 @@ const validate = ((formData, errors) => {
     );
   }
 
+  // Container isolation requires an exec user to scope between-task process
+  // cleanup inside the container's PID namespace.
+  if (containerIsolation?.enabled && !sshConfig?.execUser) {
```
athammer: one last thing i just noticed, we're hiding the linux form state in the ui and clearing it during the request/transfomration, but we'll probably need to do the same thing during validation as well so the save button doesn't get disabled if they switch from linux -> windows or something

so something like this 
```
  import { linuxArchitectures } from "./constants";

  const {
    containerIsolation,
    setup: { arch, bootstrapMethod, communicationMethod },
    sshConfig,
  } = formData;

  const supportsContainerIsolation = linuxArchitectures.includes(arch);

  if (supportsContainerIsolation) {
    // Existing three checks
  }
```

athammer (approved):
one comment we  need to fix but after that it's good to merge. thank you!

apps/spruce/src/pages/distroSettings/tabs/HostTab/HostTab.tsx
wratner: Updated validation to gate all three container-isolation checks on a supported Linux architecture. Added coverage for switching to Windows while stale invalid isolation values remain in form state; those values no longer disable Save and are still cleared by `formToGql` before submission.

PR #1800, opened 2026-07-16
DEVPROD-35243: Add previous task testing feature

sophstad:
DEVPROD-35243

### Description

Add view with additional column linking to last completed task.

### Screenshots

image

### Testing

I was planning to add tests when we decide on one option but can do now instead!

### Evergreen PR

https://github.com/evergreen-ci/evergreen/pull/10508


minnakt (approved):
LGTM - I added some proposed changes which you can revert if you don't like them

PR #1801, opened 2026-07-17
DEVPROD-37400: Replace pride with regular evergreen logo

sophstad:
DEVPROD-37400

### Description

Last year we replaced the pride logo on September 30 so I guess this is an improvement (or not...)

We previously used the LG icon implementation

```ts
<Icon glyph="EvergreenLogo" size={36} />
```

but it seems like `size` support for icons created via `createIconComponent` is broken. 

### Screenshots

image


minnakt (approved):
we should make a summer logo 🏝️

PR #1807, opened 2026-07-20
DEVPROD-23206: Replace custom pagination component with LG component for tables

minnakt:
DEVPROD-23206

### Description
This change only applies to **pagination on task/test tables**. 

Created a temporary component `Table/Pagination/index.tsx`. I also want to get rid of the `Pagination` and `PageSizeSelector` components (which are always used together) in `lib`, but that would make the PR too big. So I am saving that as a followup.

### Screenshots
Note: the Pagination component only has a count for what's being shown (`filteredCount`). So I listed the total count on the side.

Screenshot 2026-07-20 at 3 03 01 PM

### Testing
* Updated tests and added tests for the new Pagination component


apps/spruce/playwright/tests/version/version_timing.spec.ts
```diff
@@ -23,12 +23,16 @@ test.describe("Version Timing Tab without a variant selected", () => {
   });
 
   test("has disabled pagination functionality", async ({ page }) => {
-    await expect(
-      page.locator("button[aria-labelledby='page-size-select']"),
-    ).toHaveAttribute("aria-disabled", "true");
```
athammer, line 28: any reason we're deleting this test? is disabled not used anymore?

packages/lib/src/components/Table/Pagination/index.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const numTotalItems =
+    countLimit !== undefined && totalResults >= countLimit
+      ? undefined
+      : totalResults;
+
+  return (
+    <StyledPagination
+      currentPage={currentPage + 1}
+      data-cy="pagination"
+      itemsPerPage={pageSize ?? PAGE_SIZES[0]}
```
athammer: do we want `??` here as that would give us 0 which might cause some issues downstream. kinda a nit as idk why this would happen but think we just want an plain ol `||`

minnakt: ty!

apps/spruce/playwright/tests/version/version_timing.spec.ts
minnakt: I kept the test but LeafyGreen doesn't disable the page size select in their component so there's nothing to check for this assertion!

minnakt:
evergreen retry

PR #1811, opened 2026-07-21
DEVPROD-26793: Reduce Parsley statefulness

sophstad:
DEVPROD-26793

### Description

Remove some derived states by converting to `useMemo` or plain variable


athammer (approved):
nice cleanup!

PR #1815, opened 2026-07-22
DEVPROD-31461: Make lib backwards compatible with `data-testid`

minnakt:
DEVPROD-31461

### Description
Facilitates `data-cy` -> `data-testid` migration for Spruce by supporting both attributes in `lib` temporarily


athammer (approved):
looks good!

PR #1819, opened 2026-07-23
DEVPROD-27824: Fix waterfall scroll performance

sophstad:
DEVPROD-27824

### Description

We have a comment that says
> Safari performance of the waterfall chokes if using overflow-y: scroll, so we need the page to scroll instead.

...but unfortunately `overflow-y: scroll` was still being applied due to the presence of `overflow-x: hidden` 😭 

I also replaced each variant/version combination with a small grid instead of using `float: left`, which is supposed to help with performance. This enabled us to fix an occasional bug caused by an incorrectly-measured row.

### Testing

Deployed to [beta](https://spruce-beta.corp.mongodb.com/project/mongodb-mongo-master/waterfall)


minnakt (approved):
nice the grid is clean!

PR #1821, opened 2026-07-24
DEVPROD-31461: Migrate `data-cy` -> `data-testid` attributes in Spruce

minnakt:
DEVPROD-31461

### Description
Replaces `data-cy` with `data-testid` in Spruce. Sorry it's so big 🧍‍♂️

I will need to make 1 more cleanup PR after this.


apps/spruce/src/pages/task/metadata/ExecutionSection/SkippedTestsMetadata/index.tsx
```diff
@@ -40,10 +40,10 @@ export const SkippedTestsMetadata: React.FC<Props> = ({
   return (
     <>
       <MetadataItem as="div">
-        <MetadataContent data-cy="skipped-tests-metadata">
```
athammer, line 43: can we add `data-testid="skipped-tests-metadata">`? i think we just remvoed it. i just merged this in so guessing it got lost

minnakt: fixed

PR #1824, opened 2026-07-27
DEVPROD-27824: Use VersionLite for waterfall

sophstad:
DEVPROD-27824

### Description

Use non-REST based `versions` instead of `flattenedVersions` for waterfall entries


athammer (approved):
looks good!

PR #1827, opened 2026-07-30
DEVPROD-36862: Use webhook secret correctly

sophstad:
DEVPROD-36862

### Description

Use a state to track a single secret value across the save process (i.e. clicking save and the confirmation modal). Previously calling `generateWebhookSecret()` twice yielded a different display value than was actually saved.

### Testing

- Tested against staging db
- Added unit test

### DIFF
--- a/apps/spruce/src/pages/projectAndRepoSettings/shared/HeaderButtons.tsx
+++ b/apps/spruce/src/pages/projectAndRepoSettings/shared/HeaderButtons.tsx
@@ -42,13 +42,17 @@ const defaultToRepoDisabled: Set<WritableProjectSettingsType> = new Set([
   ProjectSettingsTabRoutes.GithubPermissionGroups,
 ]);
 
-interface Props {
+interface Props<T extends WritableProjectSettingsType> {
   id: string;
   projectType: ProjectType;
-  tab: WritableProjectSettingsType;
+  tab: T;
 }
 
-export const HeaderButtons: React.FC<Props> = ({ id, projectType, tab }) => {
+export const HeaderButtons = <T extends WritableProjectSettingsType>({
+  id,
+  projectType,
+  tab,
+}: Props<T>) => {
   const { sendEvent } = useProjectSettingsAnalytics();
   const dispatchToast = useToastContext();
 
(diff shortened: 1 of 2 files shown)


apps/spruce/src/pages/projectAndRepoSettings/shared/HeaderButtons.tsx
```diff
@@ -111,10 +118,7 @@ export const HeaderButtons: React.FC<Props> = ({ id, projectType, tab }) => {
     refetchQueries: ["RepoSettings", "ViewableProjectRefs"],
   });
 
-  const performSave = () => {
-    // @ts-expect-error: FIXME. This comment was added by an automated script.
```
athammer, line 115: wooo ts error removed lol

apps/spruce/src/pages/projectAndRepoSettings/shared/HeaderButtons.tsx
```diff
@@ -60,6 +64,9 @@ export const HeaderButtons: React.FC<Props> = ({ id, projectType, tab }) => {
 
   const [defaultModalOpen, setDefaultModalOpen] = useState(false);
   const [saveModalOpen, setSaveModalOpen] = useState(false);
+  const [savePayload, setSavePayload] = useState<
+    ProjectSettingsInput | RepoSettingsInput | null
+  >(null);
```
athammer, line 69: this is a simple state so I don't consider this a blocker, but do we need this extra state here? can we just derive this from `formToGql(formData, isRepo, id);` or something similar? 

if so and it makes it cleaner would suggest that, if not that's my only concern so everything else looks good!

athammer (approved):
this looks good. small question on if we can simplify state but not a blocker if we can't.

apps/spruce/src/pages/projectAndRepoSettings/shared/HeaderButtons.tsx
sophstad: Alas that would reintroduce the regenerated secret issue; we need to call it once to show on the modal for users to access, and omitting this separate state would mean that it would be generated again in the mutation when `formToGql` is called.

PR #1828, opened 2026-07-30
DEVPROD-27824: Cache improvements

sophstad:
DEVPROD-27284

### Description

A couple Apollo caching updates I've been meaning to make:
- Set `executionSelectionSet` max size of 150,000 (from default 50,000). Since server projects hit 60,000 upon first load, we don't want to have to begin cache eviction on initial load.
- Configure version limit for the waterfall page's cache. For server projects, limit to 2 pages (10 active versions) to remain under the 150k limit. Other projects get a conservative 6 pages (mms is ~20,000 entries per page so this should be no problem).
- Clear waterfall queries from the cache on project switching and when navigating away from the page. They are just so large, the data is never used by other queries on the site, and users should hit updated data instead of the cache next time they go to the waterfall.
- Split out caching read/merge functions into more understandable utils
- Add instrumentation for the waterfall performance board

### DIFF
--- a/apps/spruce/index.html
+++ b/apps/spruce/index.html
@@ -19,6 +19,11 @@
     You need to enable JavaScript to run this app.
   </noscript>
   <div id="root" style="height: 100%"></div>
+  <script>
+    globalThis[Symbol.for("apollo.cacheSize")] = {
+      "inMemoryCache.executeSelectionSet": 150000,
+    };
+  </script>
   <script type="module" src="/src/index.tsx"></script>
   <!--
         Load LeafyGreen's monospace font dependency via Google Fonts' CDN
--- a/apps/spruce/src/pages/waterfall/EmptyState/index.tsx
+++ b/apps/spruce/src/pages/waterfall/EmptyState/index.tsx
@@ -9,12 +9,16 @@ import { EmptyGraphic } from "./EmptyGraphic";
 
 interface EmptyStateProps {
   pagination: Pagination;
+  projectIdentifier: string;
 }
 
-export const EmptyState: React.FC<EmptyStateProps> = ({ pagination }) => {
+export const EmptyState: React.FC<EmptyStateProps> = ({
+  pagination,
+  projectIdentifier,
+}) => {
   const { sendEvent } = useWaterfallAnalytics();
   const { goToNextPage, hasNextPage, isNavigatingToPage } =
-    usePaginationNavigation(pagination);
+    usePaginationNavigation(pagination, projectIdentifier);
 
   const [tasks] = useQueryParam<string[]>(WaterfallFilterOptions.Task, []);
   const [statuses] = useQueryParam<string[]>(
--- a/apps/spruce/src/pages/waterfall/PaginationButtons/index.tsx
+++ b/apps/spruce/src/pages/waterfall/PaginationButtons/index.tsx
@@ -8,10 +8,12 @@ import { usePaginationNavigation } from "../usePaginationNavigation";
 
 interface PaginationButtonsProps {
   pagination: Pagination | undefined;
+  projectIdentifier: string;
 }
 
 export const PaginationButtons: React.FC<PaginationButtonsProps> = ({
   pagination,
+  projectIdentifier,
 }) => {
   const { sendEvent } = useWaterfallAnalytics();
   const {
@@ -20,7 +22,7 @@ export const PaginationButtons: React.FC<PaginationButtonsProps> = ({
     hasNextPage,
     hasPrevPage,
     isNavigatingToPage,
-  } = usePaginationNavigation(pagination);
+  } = usePaginationNavigation(pagination, projectIdentifier);
 
   const onNextClick = () => {
     sendEvent({ name: "Changed page", direction: "next" });
--- a/apps/spruce/src/pages/waterfall/WaterfallFilters.tsx
+++ b/apps/spruce/src/pages/waterfall/WaterfallFilters.tsx
@@ -1,4 +1,5 @@
 import { useCallback } from "react";
+import { useApolloClient } from "@apollo/client/react";
 import styled from "@emotion/styled";
 import { useQueryParam, useQueryParams } from "@evg-ui/lib/hooks";
 import { useWaterfallAnalytics } from "analytics";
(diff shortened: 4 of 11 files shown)


apps/spruce/index.html
```diff
@@ -19,6 +19,11 @@
     You need to enable JavaScript to run this app.
   </noscript>
   <div id="root" style="height: 100%"></div>
+  <script>
+    globalThis[Symbol.for("apollo.cacheSize")] = {
+      "inMemoryCache.executeSelectionSet": 150000,
+    };
```
athammer, line 26: this doesn't get overwritten when the js bundle loads and apollo executes? mostly curious why we need this here instead of in the react code

apps/spruce/src/pages/waterfall/WaterfallFilters.tsx
```diff
@@ -78,6 +80,12 @@ export const WaterfallFilters: React.FC<WaterfallFiltersProps> = ({
         <ProjectSelect
           getProjectRoute={projectSelectRoute}
           onSubmit={(project: string) => {
+            cache.evict({
+              id: "ROOT_QUERY",
+              fieldName: "waterfall",
+            });
+            cache.gc();
```
athammer: does this eviction cause the query to refetch/ui to refresh before we move to the new project and load that query?

apps/spruce/src/pages/waterfall/caching/index.ts
```diff
@@ -1,111 +1,208 @@
...
+  FieldReadFunction,
+} from "@apollo/client";
 import { WaterfallQuery } from "gql/generated/types";
 import { VERSION_LIMIT } from "../constants";
 
-export const readVersions = ((existing, { args, readField }) => {
-  if (!existing) {
-    return undefined;
-  }
+const DEFAULT_CACHED_PAGE_LIMIT = 6;
```
athammer: this is bounding the limit right? as in it was unlimited before causing the cache to be invalidated?

apps/spruce/src/pages/waterfall/caching/caching.test.ts
athammer, line 1: good find on `Since server projects hit 60,000 upon first load, we don't want to have to begin cache eviction on initial load.` - would be very easy to miss.

apps/spruce/src/pages/waterfall/WaterfallGrid.tsx
```diff
@@ -187,6 +190,9 @@ export const WaterfallGrid: React.FC<WaterfallGridProps> = ({
   });
   // TODO DEVPROD-26717: This can be removed if the invalid arguments are fixed in useSuspenseQuery.
   const dataIsComplete = dataState === "complete";
+  useWaterfallNavigationTrace({
+    data: dataState === "complete" ? data : undefined,
```
minnakt: nit. did you want to reuse the `dataIsComplete` variable?

apps/spruce/src/pages/waterfall/PaginationButtons/index.tsx
```diff
@@ -8,10 +8,12 @@ import { usePaginationNavigation } from "../usePaginationNavigation";
 
 interface PaginationButtonsProps {
   pagination: Pagination | undefined;
+  projectIdentifier: string;
```
minnakt: nit. Should we just grab the project identifier from the URL to avoid the prop drilling?

apps/spruce/src/pages/waterfall/EmptyState/index.tsx
```diff
@@ -9,12 +9,16 @@ import { EmptyGraphic } from "./EmptyGraphic";
 
 interface EmptyStateProps {
   pagination: Pagination;
+  projectIdentifier: string;
```
minnakt: nit. similar comment

apps/spruce/src/pages/waterfall/caching/index.ts
minnakt, line 1: This file mostly looks like a refactor with the exception of `boundVersions`, but I would maybe say to split the file into one `read` file, one `merge` file and keep the functions scoped to the relevant function(s). Also if we could add comments on some of these helper functions that would be great!

apps/spruce/src/pages/waterfall/WaterfallFilters.tsx
```diff
@@ -78,6 +80,12 @@ export const WaterfallFilters: React.FC<WaterfallFiltersProps> = ({
         <ProjectSelect
           getProjectRoute={projectSelectRoute}
           onSubmit={(project: string) => {
+            cache.evict({
```
minnakt: Possibly makes sense to share in a hook like `useEvictWaterfallCache`

apps/spruce/index.html
minnakt: +1 I'm also wondering about the placement

sophstad: This value should be set [before Apollo is loaded](https://www.apollographql.com/docs/react/caching/memory-management#setting-cache-sizes-before-loading-the-apollo-client-library), and I think this location is the easiest place to ensure that behavior

apps/spruce/src/pages/waterfall/WaterfallFilters.tsx
sophstad: Set `broadcast: false` to prevent this!

apps/spruce/src/pages/waterfall/caching/index.ts
sophstad: Correct

apps/spruce/src/pages/waterfall/caching/index.ts
sophstad: Done!

apps/spruce/index.html
athammer: sounds good to me, if we don't already can we add a comment to where apollo is loaded just calling this out? just so we don't have to remember about this configuration

athammer (approved):
looks good to me, good klo/tech debt/whatever we're calling it work lol

minnakt (approved):
tweaked the title to make it a little more specific but you can change as you like!

PR #1831, opened 2026-08-03
DEVPROD-34590: Improve settings diff rendering

sophstad:
DEVPROD-34590

### Description

Simplify diff code as used in event logs and on the changes modal to more clearly show inserts/deletions. Previously, these would render as shifted arrays, which is really unclear.

The diff code is kinda funky in order to find the "new" entry but it's hard to work with untyped data like these event objects in a safe way

### Screenshots

Alias addition
image

Adding array element:
image

### Testing

- Added lots of tests
- Update storybook


apps/spruce/src/components/Settings/EventLog/EventDiffTable/index.tsx
```diff
@@ -76,7 +98,79 @@ const renderEventValue = (value: JSONValue): string => {
...
+  if (typeof customRenderedValue !== "string") {
+    return customRenderedValue;
+  }
+
+  if (Array.isArray(before) || Array.isArray(after)) {
+    return renderArrayValue(
+      Array.isArray(before) ? before : [],
+      Array.isArray(after) ? after : [],
+      side,
+    );
```
athammer: nit but do we want to turn `before` into an array? im guessing going from undefined, null, or "" into an array and would return the before into an empty array? 

if that's intentional that's fine and probably visibly easier to understand for users, but just wanted to call out in case there's more meaning there.

athammer (approved):
question on the `renderCellValue` function but looks good! thanks for adding the storybooks, made it easier to test locally. will review and approve the story book changes as well if not already done.

sophstad:
evergreen refresh

PR #1839, opened 2026-08-06
DEVPROD-37414: Remove :javascript task path vuln

athammer:
DEVPROD-37414

### DIFF
--- a/apps/spruce/src/pages/version/Tabs/TestAnalysis/GroupedTestMapList/FailedTestGroup/FailedTestGroupTable/index.tsx
+++ b/apps/spruce/src/pages/version/Tabs/TestAnalysis/GroupedTestMapList/FailedTestGroup/FailedTestGroupTable/index.tsx
@@ -8,6 +8,7 @@ import {
   useLeafyGreenTable,
 } from "@evg-ui/lib/components/Table";
 import { TaskStatus } from "@evg-ui/lib/types/task";
+import { isValidHttpUrl } from "@evg-ui/lib/utils/url";
 import { getTaskRoute } from "constants/routes";
 import { TaskBuildVariantField } from "pages/version/Tabs/TestAnalysis/types";
 import { TaskTab } from "types/task";
@@ -65,15 +66,18 @@ const getColumns = (): LGColumnDef<TaskBuildVariantField>[] => [
   {
     header: "Logs",
     meta: { width: "10%" },
-    cell: ({ row }) => (
-      <Button
-        data-cy="failed-test-group-parsley-btn"
-        href={row.original.logs.urlParsley}
-        size="xsmall"
-      >
-        Parsley
-      </Button>
-    ),
+    cell: ({ row }) => {
+      const { urlParsley } = row.original.logs;
+      return isValidHttpUrl(urlParsley) ? (
+        <Button
+          data-cy="failed-test-group-parsley-btn"
+          href={urlParsley}
+          size="xsmall"
+        >
+          Parsley
+        </Button>
+      ) : null;
+    },
   },
 ];
 
--- a/packages/lib/src/utils/url/index.ts
+++ b/packages/lib/src/utils/url/index.ts
@@ -1,0 +1,20 @@
+const HTTP_PROTOCOLS = new Set(["http:", "https:"]);
+
+/**
+ * Returns whether a URL is an absolute HTTP(S) URL.
+ * @param url - URL to validate.
+ * @returns Whether the URL uses the HTTP or HTTPS protocol.
+ */
+const isValidHttpUrl = (url?: string | null): url is string => {
+  if (!url) {
+    return false;
+  }
+
+  try {
+    return HTTP_PROTOCOLS.has(new URL(url).protocol);
+  } catch {
+    return false;
+  }
+};
+
+export { isValidHttpUrl };
--- a/apps/spruce/src/pages/jobLogs/JobLogsTable.tsx
+++ b/apps/spruce/src/pages/jobLogs/JobLogsTable.tsx
@@ -7,6 +7,7 @@ import {
   TablePlaceholder,
   useLeafyGreenTable,
 } from "@evg-ui/lib/components/Table";
+import { isValidHttpUrl } from "@evg-ui/lib/utils/url";
 import { useJobLogsAnalytics } from "analytics/joblogs/useJobLogsAnalytics";
 import { EvergreenTestResult } from "./types";
 
@@ -70,6 +71,9 @@ const ParsleyLink = ({
   testName: string;
 }) => {
   const { sendEvent } = useJobLogsAnalytics();
+  if (!isValidHttpUrl(parsleyUrl)) {
+    return testName;
+  }
   return (
     <Link
       hideExternalIcon
--- a/apps/spruce/src/pages/task/taskTabs/FileTable/GroupedFileTable/AssociatedLinks.test.tsx
+++ b/apps/spruce/src/pages/task/taskTabs/FileTable/GroupedFileTable/AssociatedLinks.test.tsx
@@ -1,0 +1,56 @@
+import { useTaskAnalytics } from "analytics";
+import { processFilesWithAssociatedLinks } from "./AssociatedLinks";
+
+const taskAnalytics = {
+  sendEvent: vi.fn(),
+} as unknown as ReturnType<typeof useTaskAnalytics>;
+
+describe("processFilesWithAssociatedLinks", () => {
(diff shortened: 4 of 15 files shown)


apps/parsley/src/constants/errors.tsx
```diff
@@ -12,7 +12,7 @@ export const LOG_FILE_DOWNLOAD_TOO_LARGE_WARNING = (
     Parsley was only able to partially download this log. Use the Evergreen CLI
     command in the Details panel to download the full log. See other download
     methods{" "}
-    <StyledLink href={`${docsURL}/Downloading-Logs`} target="__blank">
```
athammer, line 15: not strictly related to this pr, but there were a few `__blank` instead of `_blank` so where it made sense to link out to a new tab i removed the second `_` and if not removed it entirely

packages/lib/src/utils/url/index.ts
```diff
@@ -0,0 +1,20 @@
...
+ * @param url - URL to validate.
+ * @returns Whether the URL uses the HTTP or HTTPS protocol.
+ */
+const isValidHttpUrl = (url?: string | null): url is string => {
+  if (!url) {
+    return false;
+  }
+
+  try {
+    return HTTP_PROTOCOLS.has(new URL(url).protocol);
```
athammer, line 38: This is the core of the PR. `:javascript` is considered a url so we pretty much have to check by protocol sadly, we can't just check by url validity. This means it's a bit of a risky change as someone could pass a relative url and we'd fail this.

I went through the BE and the use cases where we call `isValidHttpUrl` and this _shouldn't_ happen and we should always be dealing with a properly formatting and correct protocol url as the BE usually checks for that or it's set in env vars as already properly formatted, but it's still worth calling out here.


here's the validation table
| input | parses as a URL? | http protocols? | 
| --- | --- | --- |
| `javascript:alert(1)` | **true** | false |
| `data:text/html,<script>` | **true** | false |
| `vbscript:msgbox(1)` | **true** | false |
| `file:///etc/passwd` | **true** | false |
| `ms-msdt:/id x` | **true** | false |
| `https://ok.com/x` | true | true |
| `http://ok.com/x` | true | true |
| `some_link` | false | false |
| `/relative` | false | false |
| `https://` | false | false |

packages/lib/src/utils/url/index.ts
```diff
@@ -0,0 +1,41 @@
...
+  } catch {
+    reportUnparseableUrl(url);
```
athammer, line 40: created a honeycomb dashboard to watch this for a few days after deploy https://ui.honeycomb.io/mongodb-4b/environments/production/board/pbJmotmcubo/Spruce-URL-Validation-DEVPROD-37414

packages/lib/src/utils/url/index.ts
```diff
@@ -0,0 +1,41 @@
...
+  reportedUrls.add(url);
+
+  const span = trace
+    .getTracer("analytics")
+    .startSpan("System Event rejected unparseable url");
+  span.setAttribute("url.value", url.slice(0, MAX_REPORTED_URL_LENGTH));
+  span.setAttribute("url.is_relative_path", url.startsWith("/"));
+  span.end();
+};
+
```
minnakt, line 22: I feel like we should also check these URLs in the backend code using the util, to prevent these from being added as artifact URLs in the first place. Were you able to investigate that approach?

packages/lib/src/utils/url/index.ts
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Returns whether a URL is an absolute HTTP(S) URL.
+ * @param url - URL to validate.
+ * @returns Whether the URL uses the HTTP or HTTPS protocol.
+ */
+const isValidHttpUrl = (url?: string | null): url is string => {
```
minnakt, line 32: Is it possible to centralize the usage so that the implementer doesn't have to remember to call this every time? Can `isValidHttpUrl` be called in `StyledLink`, can we make a wrapper for `Button` that checks links?

packages/lib/src/utils/url/index.ts
athammer: totally agree on this. the ticket was just talking about FE code so if we get another vuln scan this should close any of them for this issue but will create a follow up ticket for the BE work.

edit: actually wondering if we already have a jira ticket for this from the scan lemme see

edit edit: 
noticing at the very end of the ticket it says "Defense-in-depth: also reject non-http(s) artifact URL schemes server-side in the Go app when a task registers an artifact." so going to just keep this ticket and do the BE work on it

packages/lib/src/utils/url/index.ts
athammer: let me look into this a bit. `isValidHttpUrl` is kinda a dangerous function to use globally as it will break for things such as relative routes, `blob:`, `mailto:`, ext ext so was a bit worried about using it more than i have to to fix the vuln here.

packages/lib/src/utils/url/index.ts
minnakt: sounds good as long as we have something tracking, the ownership here is blurry so realistically either team can own the fix

packages/lib/src/utils/url/index.ts
athammer: going to keep this here but make sure it's clear this is unsafe for relative urls so it's a bit more clear why it's here. let me know if that's good, i do think we can add a isSafeUrl or something to all the buttons and links and such but might be a bit too much work for this pr

update: added a comment there

packages/lib/src/utils/url/index.ts
athammer: that makes sense to me. i can carry this across the line so we dont have to do any handoff. might close the SEC ticket as it mentions its a follow up (for slo reasons)

apps/spruce/src/pages/task/taskTabs/buildBaronAndAnnotations/AnnotationTicketsList/AnnotationTicketRow/AnnotationTicketRow.test.tsx
minnakt, line 1: thanks for adding tests to these components

apps/spruce/src/pages/task/taskTabs/buildBaronAndAnnotations/AnnotationTicketsList/AnnotationTicketRow/AnnotationTicketRow.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+    renderRow("https://jira.mongodb.org/browse/DEVPROD-123");
+
+    expect(screen.getByTestId(issueKey)).toHaveAttribute(
+      "href",
+      "https://jira.mongodb.org/browse/DEVPROD-123",
+    );
+  });
+
+  // The annotation URL is submitted through a form that only enforces a minimum
+  // length, so it can be any string.
```
minnakt: nit. let's remove the comment, the test description makes sense to me
```suggestion
```

apps/spruce/src/pages/version/Tabs/TestAnalysis/GroupedTestMapList/FailedTestGroup/FailedTestGroupTable/index.tsx
```diff
@@ -65,15 +66,18 @@ const getColumns = (): LGColumnDef<TaskBuildVariantField>[] => [
...
-        size="xsmall"
-      >
-        Parsley
-      </Button>
-    ),
+    cell: ({ row }) => {
+      const { urlParsley } = row.original.logs;
+      return isValidHttpUrl(urlParsley) ? (
+        <Button
+          data-testid="failed-test-group-parsley-btn"
```
minnakt: nit. while we're here can we remove this, it seems unused
```suggestion
```

apps/spruce/src/pages/version/Tabs/TestAnalysis/GroupedTestMapList/FailedTestGroup/FailedTestGroupTable/index.tsx
```diff
@@ -65,15 +66,18 @@ const getColumns = (): LGColumnDef<TaskBuildVariantField>[] => [
...
+          href={urlParsley}
+          size="xsmall"
```
minnakt: nit. can we replace with `ButtonSize.XSmall`?

minnakt (approved):

athammer:
updated the pr based on your comments. gonna merge once pr is finished. thank you for the review!

